PR #53, opened 2025-09-29
feat: allow overriding provider

SgtPooki:
Sometimes, a provider may get in a bad state, or we may need to customize the provider we use. This PR adds the ability to use a custom provider by setting the `PROVIDER_ADDRESS` or `PROVIDER_ID` environment variables.

### DIFF
--- a/src/synapse/service.ts
+++ b/src/synapse/service.ts
@@ -3,6 +3,7 @@ import {
   type ProviderInfo,
   RPC_URLS,
   type StorageContext,
+  type StorageServiceOptions,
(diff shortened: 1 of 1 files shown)


rvagg:
re "muliple providers" - we're going to have to work on a story for duplicate deals here but first we need to figure out what that looks like in Synapse, then we can inherit it here .. maybe, but the ultimate goal is that we do have multiple storage deals per piece, it's just not clear how the mechanics of that will work and what layer of the stack it's up to to sort that out.

PR #81, opened 2025-10-08
refactor: remove all uses of ReturnType with explicit typing

rvagg:


rvagg:
oh, whoops! sorry @SgtPooki wrong window, but I've merged this into your PR now so you can just merge that one which I already approved and we have it in master

PR #114, opened 2025-10-14
refactor: remove duplication in CAR blockstores

rvagg:
***Draft for now, sitting on top of #94 which is slightly blocked for now. This is not urgent, just something I had sitting in my local and wanted to get pushed.***

Also does the dedupe thing in the Node.js one as well. And cleans up a bunch of other guff accumulated along the way.

I think this is a fairly clean separation, where we have different "backends" but there's a shared logic layer in `CARBlockstoreBase`.

This is going to have to mutate again when we get streaming upload, but we'll also want to retain the Node.js read/write form to support the pinning service, but since both Node.js and the browser will need streaming for the `filecoin-pin add` style ingest then we'll still have a shared logic for that.

### DIFF
--- a/src/test/unit/car-blockstore-shared.test.ts
+++ b/src/test/unit/car-blockstore-shared.test.ts
@@ -4,13 +4,63 @@ import { CID } from 'multiformats/cid'
 import * as raw from 'multiformats/codecs/raw'
 import { sha256 } from 'multiformats/hashes/sha2'
 import { afterEach, beforeEach, describe, expect, it } from 'vitest'
-import { CARWritingBlockstore } from '../../core/car/index.js'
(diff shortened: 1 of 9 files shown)


rvagg:
Pulled a bit of this into #94 (it-to-buffer use) so it's slightly simpler now.

src/test/unit/car-blockstore-shared.test.ts
```diff
@@ -0,0 +1,383 @@
...
+      new BrowserCARBlockstore({
+        rootCID,
+      }),
+    cleanup: async () => {
+      // No cleanup needed for in-memory
+    },
+  },
+]
+
+describe.each(factories)('CARWritingBlockstore - $name', (factory) => {
```
SgtPooki: TIL about describe.each with vitest. nice.

PR #140, opened 2025-10-20
feat: add sp health monitor script (DRAFT)

rvagg:
End-to-end SP health testing script that exercises the Filecoin Pin flow with IPNI publishing. Let's us see who's broken and figure out where to dive in.

Not necessarily wanting to check this in here, just parking it for now. It makes a little more sense here because it's Filecoin Pin and we have `@ipld/car` here already so I don't need anything additional from Synapse. Although I could just manually craft a CAR ..

### DIFF
--- a/utils/test-provider-health.js
+++ b/utils/test-provider-health.js
@@ -4,20 +4,22 @@
  * Storage Provider Health Check Script
  *
  * This script tests the complete workflow of a storage provider by:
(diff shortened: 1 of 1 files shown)


BigLep (approved):
Nice.  seems like a good clean example too of the end to end flow.  Your call of course, but I can see this being useful to have checked in.

PR #55, opened 2025-09-30
feat: allow overriding withCDN via env-var

SgtPooki:
```
> npm run build && WITH_CDN=true npx filecoin-pin add ~/Downloads/images/IMG_1341.png


> filecoin-pin@0.6.0 build
> tsc

┌  Filecoin Pin Add
│
◇  ✓ File validated (17.8 MB)
│
◇  ✓ Connected to calibration
│
◇  ✓ Payment capacity verified
│
◇  ✓ File packed with root CID: bafybeierfvvhvb73ntvschtheysvpscpynjuxgsi7jtst4wg3n3jittuca
│
◇  ✓ IPFS content loaded (17.8 MB)
│
◇  ✓ Payment capacity verified
│
◇  ✓ Storage context ready
│
◇  ━━━ Add Complete ━━━
│
│  Network: calibration
│  
│  Add Details
│    File: /Users/sgtpooki/Downloads/js-colo-media/IMG_1341.png
│    Size: 17.8 MB
│    Root CID: bafybeierfvvhvb73ntvschtheysvpscpynjuxgsi7jtst4wg3n3jittuca
│  
│  Filecoin Storage
│    Piece CID: bafkzcibfqh67ubquflgzvzb76hayv6o4yqer2wpa7susa6xna3yfktvb5zxv37nbvqbq
│    Piece ID: 0
│    Data Set ID: 244
│  
│  Storage Provider
│    Provider ID: 3
│    Name: ezpdpz-calib
│    Direct Download URL: https://calib.ezpdpz.net/piece/bafkzcibfqh67ubquflgzvzb76hayv6o4yqer2wpa7susa6xna3yfktvb5zxv37nbvqbq
│  
│  Transaction
│    Hash: 0x88314b2d3f576fb0a248f82be0971cc77a4b68ad67acc3df4db013058ff6538e
│
└  Add completed successfully

> npx filecoin-pin data-set 244

┌  Filecoin Onchain Cloud Data Sets
│
◇  ━━━ Data Sets ━━━
│
│  Data Set #244 • live
│    Managed by Warm Storage: yes
│    CDN add-on: enabled
│    Pieces stored: 1
│    Leaf count: 588804
│    Total size: 18.0 MB
│    Client data set ID: 3
│    PDP rail ID: 485
│    CDN rail ID: 487
│    Cache-miss rail ID: 486
│    Payer: 0x44f08D1beFe61255b3C3A349C392C560FA333759
[...]

PR #144, opened 2025-10-21
fix: more ethers.js cleanup silencing

rvagg:
Ref: https://github.com/filecoin-project/filecoin-pin/pull/55#issuecomment-3425432916

### DIFF
--- a/src/core/synapse/index.ts
+++ b/src/core/synapse/index.ts
@@ -558,11 +558,27 @@ export function getDefaultStorageContextConfig(overrides: any = {}) {
  */
 export async function cleanupProvider(provider: any): Promise<void> {
   if (provider && typeof provider.destroy === 'function') {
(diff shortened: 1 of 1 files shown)


src/core/synapse/index.ts
```diff
@@ -558,11 +558,27 @@ export function getDefaultStorageContextConfig(overrides: any = {}) {
...
+
     try {
       await provider.destroy()
     } catch {
       // Ignore cleanup errors
     }
+
+    // Small delay to allow any pending async operations to complete
+    // This prevents errors from scheduled operations that trigger after destroy()
+    await new Promise((resolve) => setTimeout(resolve, 100))
```
SgtPooki: pretty wild that .destroy() contains side-effects instead of waiting for all the cleanup to actually complete.

PR #183, opened 2025-10-30
fix: check ipni advertisement during upload

### LINKED ISSUE #137: Action: check `filecoinpin.contact/cid/$cid` before showing IPFS Mainnet retrieval URLs
## Done Criteria
The example GitHub Action doesn't show the closing summary with IPFS Mainnet retrieval URLs until it confirms that `filecoinpin.contact/cid/$cid` is non-empty.

## Why Important
1. We don't want to set the expectation that IPFS Mainnet retrieval works by showing URLs if we can know ourselves that it won't if there is no record in filecoinpin.contact.
2. Prematurely accessing IPFS Mainnet URLs makes the situation worse because of the negative caching in cid.contact.

## Notes
1. This filecoinpin.contact logic already exists in filecoin-pin-website.  Ideally we pull it in as a helper function to filecoin-pin and then the example action and filecoin-pin-website can use it.
2. While the helper function today may be tied to filecoinpin.contact polling, ideally the helper is named more abractly like `validateIpniAdvertisment(String cid)`

SgtPooki:
This change adds first-class IPNI announcement validation to filecoin-pin core and wires it into the GitHub Action:

- Core: executeUpload now (by default) waits for the IPFS Root CID to be announced to IPNI after onUploadComplete, returning ipniValidated.
- Helper: validateIPNIAdvertisement emits retryUpdate and a final complete event for consumers (e.g. website, gh-action, and CLI after #169) to observe the progress of the validation.
- GitHub Action: uses executeUpload with onProgress to log attempts, and only shows IPFS gateway URLs if `ipniValidated` is true.
- CLI: disables IPNI validation by default (we don't show gateway links yet).

Follow-up: The CLI improvements in #169 will switch to consume this core method and its onProgress callback instead of rolling its own polling, and switch the CLI default of ipniValidation to `enabled: true`.

Fixes #137

References:
- https://github.com/filecoin-project/filecoin-pin/pull/169

### DIFF
--- a/upload-action/src/upload.js
+++ b/upload-action/src/upload.js
@@ -114,8 +114,12 @@ export async function runUpload(buildContext = {}) {
     throw new Error('walletPrivateKey is required for upload phase')
   }
 
-  /** @type {Partial<UploadResult>} */
-  let pieceCid, pieceId, dataSetId, provider, previewUrl, network, ipniValidated
+  /**
+   * @type {Partial<UploadResult>}
+   * This is a simple trick to get each of these variables defined easily without using a separate let statement and jsdoc for each one.
+   * We already have them in the type we want.
+   */
+  let { pieceCid, pieceId, dataSetId, provider, previewUrl, network, ipniValidated } = {}
   /** @type {PaymentStatus} */
   let paymentStatus
 


rvagg:
This has become annoying for me thanks to https://github.com/nodejs/node/pull/60334; it always fails the test because filecoinpin.contact is a dual stack DNS, I don't have ipv6, I'm on the other side of the world from filecoinpin.contact, and Node.js's fail-fast implementation means the failure of ipv6 leads to a total failure.

Not saying anything should be done here, but I'm not going to be the only one hitting this and until that PR lands and gets out into a v25 release and then eventually 🤞 gets back into an lts, this could be painful. There's not a whole lot to be done unfortunately, unless we turn off dual-stack for filecoinpin.contact.

This won't impact the browser though, just Node.js.

Issue #127, opened 2025-10-16
feat: version check on CLI cmd run and post when updated version is available


SgtPooki:
tracking issue for letting users know when there is an updated version available

- [ ] CLI
- [ ] example github action

BigLep:
Some other UX thoughts:
1. For CLI we notify once per day (I assume this means we need to getting into storing some state on disc)
2. For GitHub action we notify on every run (as I assume we don't have state that can be stored between runs)

PR #219, opened 2025-11-07
feat: perform update checks

SgtPooki:
## Summary
- add a shared `checkForUpdate` helper that reads the local package metadata via `core/utils/version.ts` and reports disabled/up‑to‑date/update/error states
- wire the CLI to start the update check in `preAction`, surface a concise stdout reminder with a link to the releases page in `postAction`, and honor `--no-update-check` without touching Clack output
- surface the same update reminder inside the GitHub Action via `core.notice`, including the releases URL
- centralize package name/version access (both CLI + telemetry) through `core/utils/version.ts` and fill it after the build with `scripts/write-version.mjs` to avoid importing `package.json` at runtime
- document the new behaviour in the main README and the upload-action README, and add unit coverage around the helper (update available, up-to-date, network failure, opt-out)

## Testing
- `npm run lint`
- `npm run test:unit -- version-check`

Addresses intitial items from https://github.com/filecoin-project/filecoin-pin/issues/127

### DIFF
--- a/upload-action/README.md
+++ b/upload-action/README.md
@@ -10,8 +10,6 @@ This GitHub Action is provided to illustrate how to use filecoin-pin, a new IPFS
 
 See the two-workflow approach in the [examples directory](./examples/) for complete workflow files and setup instructions.
 
-**Heads-up on updates:** the action checks npm for a newer `filecoin-pin` release at the start of each run and posts a GitHub Actions notice when one is available.
-
 ## Inputs & Outputs
 
 See [action.yml](./action.yml) for complete input documentation including:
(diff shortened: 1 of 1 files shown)


upload-action/src/run.mjs
```diff
@@ -1,11 +1,31 @@
...
+    }
+  } catch (error) {
+    const message = error instanceof Error ? error.message : 'unknown error'
+    core.debug(`Update check failed: ${message}`)
+  }
+}
+
 async function main() {
+  // Check for updates in the background.
+  void maybeNotifyAboutUpdates()
```
SgtPooki: should not need a catch here since it's handled in `maybeNotifyAboutUpdates`

SgtPooki:
@BigLep I guess the only question here now is do we want to do the once-per-day check with a config file?

BigLep (commented):
The concept and output seems good to me.  I made a slight update on the action readme: https://github.com/filecoin-project/filecoin-pin/pull/219/commits/6cbde9abe268fd6ec2506770f5756e013ae4417e

Issue #200, opened 2025-11-04
Metadata for `add` and `import`


rvagg:
Extend `ImportOptions` and `AddOptions` to include `metadata?: Record<string, string>` and extend the CLI to allow provision of this. I suggest `--metadata key=value` where you can provide as many `--metadata` arguments as you want (until the contract rejects it for putting too many). `value` can be an empty string.

This then needs to be plumbed through `UploadFlowOptions` passed to `performUpload()` which can then be passed in to `executeUpload()` which already accepts an optional `metadata` (we just don't currently use it there).

In addition, some CLI special case sugar we are considering for EIP 8004 to get consistent and well-formed metatadata (spit-balling here, feel free to propose better alternatives to getting this metadata in):

* `--8004-type=<registration|validationrequest|validationresponse|feedback>`
* `--8004-agent=id`

You would have to provide both, or neither.

They would map internally to `--metadata 8004registration=agentId` or `--metadata 8004validationrequest=agentId`, etc.

PR #226, opened 2025-11-10
feat: allow passing metadata via cli commands

SgtPooki:
Add metadata options to CLI commands and pipe through to the core module.

Validation is done in the core module and used by CLI.

Also added some 8004 agent option sugar, and some CLI/commander parsing utils

Fixes https://github.com/filecoin-project/filecoin-pin/issues/200


src/add/add.ts
```diff
@@ -69,6 +70,11 @@ export async function runAdd(options: AddOptions): Promise<AddResult> {
 
   const spinner = createSpinner()
 
+  const { metadata: uploadMetadata, dataSetMetadata } = normalizeMetadataConfig({
+    metadata: options.metadata,
```
BigLep: call this upload or piece metadata rather than the generic "metadata"?

src/test/unit/add.test.ts
```diff
@@ -194,6 +194,44 @@ describe('Add Command', () => {
...
+        expect.anything(),
+        expect.objectContaining({
+          dataset: {
+            metadata: { purpose: 'erc8004' },
+          },
+        })
+      )
+
+      expect(vi.mocked(initializeSynapse)).toHaveBeenCalledWith(
+        expect.objectContaining({
```
BigLep: JS newbie: is there a reason we check the synapse initialization at the end?  I would have thought to check it first since it happens first.

BigLep (commented):
Generally seems good. 

A couple of things:
1. Can we give a more specific name to pieceMetadata than metadata given there is also dataSetMetadata?
2. I think I grokked how everything would look on the command line, but can you please show various command line options showing the functionality.

src/test/unit/add.test.ts
SgtPooki: fixed and put them in order

PR #237, opened 2025-11-13
fix: add shared guard to prevent zero pricing

### LINKED ISSUE #38: bug: calculateStorageFromUSDFC allows 0 pricePerTib
We should throw an error if pricePerTib is 0

> yeah, we can throw on this - we should expect it to always be positive, in fact we shouldn't even need that call because it's stable (for now); we've been discussing pricing a lot this week and we expect it to change but unclear how yet but there's lots of pain in presenting this to users

_Originally posted by @rvagg in https://github.com/filecoin-project/filecoin-pin/pull/35#discussion_r2374542295_

entuziaz:
**Fixes #38**. Prevents zero pricePerTiBPerEpoch in capacity and storage calculations. 

- Introduced assertPriceNonZero() in core/utils/validate-pricing.ts
- Integrated into calculateActualCapacity() and calculateStorageFromUSDFC()
- Updated tests to expect throws on zero pricing
- Prevents Infinity / unsafe deposit calculations

#### Changes
- Added `core/utils/validate-pricing.ts` with `assertPriceNonZero()`
- Integrated guard into:
   - `calculateActualCapacity()`
   - `calculateStorageFromUSDFC()`
- Updated unit tests to expect errors instead of silent 0-capacity returns.

Check this out @SgtPooki.

### DIFF
--- a/src/core/payments/index.ts
+++ b/src/core/payments/index.ts
@@ -18,6 +18,7 @@
 import { SIZE_CONSTANTS, type Synapse, TIME_CONSTANTS, TOKENS } from '@filoz/synapse-sdk'
 import { ethers } from 'ethers'
 import { isSessionKeyMode } from '../synapse/index.js'
+import { assertPriceNonZero } from '../utils/validate-pricing.js'
 import {
   BUFFER_DENOMINATOR,
   BUFFER_NUMERATOR,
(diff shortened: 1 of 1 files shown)


entuziaz:
> Lgtm but needs a rebase?

Aye aye, sir!
Done, @SgtPooki.

entuziaz:
> @entuziaz looks like you need to run `npm run lint:fix` on your code, see https://github.com/filecoin-project/filecoin-pin/actions/runs/19337489097/job/55344993018?pr=237

Done, @SgtPooki

PR #251, opened 2025-11-18
fix: pad rawSize for payment calculations

### LINKED ISSUE #213: fix: use padded rawSize for payment calculations
> [...] pad rawSize `const paddedSize = Math.ceil(size / 32) * 32` and use that for pricing in the `calculate*` functions in filecoin-pin

 _Originally posted by @SgtPooki in [#339](https://github.com/FilOzone/synapse-sdk/issues/339#issuecomment-3434938779)_

kaf-lamed-beyt:
`calculateRequiredAllowances()`, supposedly, is the entry point for all the calculate* functions in /core/payments.

calling `padSizeToPDPLeaves` once in it, shares that converted value with the rest.

closes #213

### DIFF
--- a/src/core/payments/constants.ts
+++ b/src/core/payments/constants.ts
@@ -57,13 +57,3 @@ export const STORAGE_SCALE_MAX_BI = BigInt(STORAGE_SCALE_MAX)
  * @see - https://github.com/FilOzone/synapse-sdk/issues/339#issue-3539254596
  */
 export const PDP_LEAF_SIZE = 32
-
-/**
- * Pad raw size to the next multiple of 32 bytes
- *
- * @param rawSizeBytes - The actual size in bytes
- * @returns Padded size (next multiple of 32)
- */
-export function padSizeToPDPLeaves(rawSizeBytes: number): number {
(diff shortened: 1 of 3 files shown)


src/core/payments/constants.ts
```diff
@@ -51,3 +51,19 @@ export const BUFFER_DENOMINATOR = 10n
...
+
+/**
+ * Pad raw size to the next multiple of 32 bytes
+ *
+ * @param rawSizeBytes - The actual size in bytes
+ * @returns Padded size (next multiple of 32)
+ */
+export function padSizeToPDPLeaves(rawSizeBytes: number): number {
+  return Math.ceil(rawSizeBytes / PDP_LEAF_SIZE) * PDP_LEAF_SIZE
+}
```
SgtPooki: this shouldn't go in constants.ts it should probably go in a new file: `/core/payments/utils.ts`

SgtPooki (approved):
lgtm

SgtPooki:
> quick one, @SgtPooki, did you eventually find the remaining `pieceSize` values?

I took a quick peek and I believe all the code does go through here.. but we may have to address this later when we run into them

PR #262, opened 2025-11-22
fix: use StorageManager for upload

wjmelements:
Reviewer @SgtPooki
CC @nijoe1
This might fix the undefined CID issue caused by https://github.com/FilOzone/synapse-sdk/pull/446
It is a temporary fix to workaround synapse.
#### Changes
* use StorageManager for upload

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,8 @@
   "homepage": "https://github.com/filecoin-project/filecoin-pin#readme",
   "dependencies": {
     "@clack/prompts": "^0.11.0",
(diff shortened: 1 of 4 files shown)


wjmelements:
tests are failing because `MockSynapse` doesn't have a complete StorageManager.
```
  // Storage namespace matches SDK structure
  public readonly storage = {
    createContext: this.createStorageContext.bind(this),
  }
```

SgtPooki:
it also looks like we can't pass in metadata for piece's with synapse-sdk currently?

PR #267, opened 2025-11-24
fix: temporarily make logger optional and move into createStorageCont…

### LINKED ISSUE #252: logger is required by `createStorageContext` in filecoin-pin
we shouldn't require logger in our `createStorageContext` function.. it should be optional and a part of the `options` object.

this will be a breaking change.

entuziaz:
Fixes #252

This PR updates createStorageContext so logger is no longer a required positional argument and is now an optional field on CreateStorageContextOptions. This implementation still supports the previous structure internally to avoid breaking changes. 

Formerly:
```
createStorageContext(synapse, logger, options?)
```

Updated:
```
createStorageContext(synapse, {
  ...options,
  logger
})
```

**On Breaking Change:** Internal callers and tests have been updated accordingly.

### DIFF
--- a/src/core/synapse/index.ts
+++ b/src/core/synapse/index.ts
@@ -428,17 +428,19 @@ export async function initializeSynapse(config: Partial<SynapseSetupConfig>, log
  * @example
  * ```typescript
  * // Create a new dataset (multi-user scenario)
- * const { storage } = await createStorageContext(synapse, logger, {
+ * const { storage } = await createStorageContext(synapse, {
+ *   logger,
  *   dataset: { createNew: true }
  * })
  *
  * // Connect to existing dataset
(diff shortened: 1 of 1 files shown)


entuziaz:
@SgtPooki  Okay, I understand. Thank you.

entuziaz:
@SgtPooki I got failures in validate-ipni-advertisement tests that seem unrelated to the createStorageContext change (e.g. lastActualMultiaddrs.intersection is not a function).

Should I look into the failing tests and see if I could fix them, or should I keep this PR focused on the logger-options issue at hand?

SgtPooki:
@entuziaz it looks like the code in `/upload-action` also needs fixed to use the new function signature.

PR #270, opened 2025-11-25
docs: fix wrong service providers URL

### LINKED ISSUE #269: storage provider URL in gloassary.md  returns a 404
> https://filecoin.cloud/providers

see here: 

https://github.com/filecoin-project/filecoin-pin/blob/master/documentation/glossary.md#service-provider-registry

i think it should be this(?): https://www.filecoin.services/providers

kaf-lamed-beyt:
closes #269

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ test-car-output/
 test-e2e-cars/
 test-simple-car-output/
 coverage
-.zed


kaf-lamed-beyt:
ah, let me do that now.  my bad.

that must be prettier doing what it is not supposed to.

kaf-lamed-beyt:
> @kaf-lamed-beyt you can always do `git add -p` to only stage piece of changes you want to commit before committing, or even do a `hunk` based staging via many GUI editors.
> 
> usually, a .gitignore should only contain the relevant files for a particular project and the standard tools of the maintainers.. while it's probably not harmful, having everyone that commits to a repo add their preferred tools to the `.gitignore` could result in a very long and unmanageable file where core maintainers may not be aware what/why certain lines exist.
> 
> It's usually best practice to just not stage & commit changes unless you know they are relevant to a particular repo

oh, i get you now. thank you for the feedback, @SgtPooki 

apologies for not responding on time. i see that you've already removed the line.

kaf-lamed-beyt:
> @kaf-lamed-beyt also, you can do global git ignore files (and override that for individual repos/folders if you desire), or a local only ignore (no commit needed) for just this repo by doing:
> 
> ```shell
> > echo ".zed" >> .git/info/exclude
> ```

thank you! i'll do this too.

SgtPooki (approved):

PR #296, opened 2025-12-25
feat: add --data-set and --new-data-set flags to add command

beck-8:


SgtPooki (changes requested):
makes sense, but one callout: 

naming conventions are DataSet is two separate words, so we should stick to that, or add an alias at the very least. `dataSet` or `data-set` over `dataset`, but we have an alias in other places, so I'm good with `dataset` as long as we also have `data-set` in the CLI.

vars internally should use camelCase

SgtPooki (changes requested):
thanks for the quick updates @beck-8.. a few last things

SgtPooki:
Approved, but we need to get the CI passing.. looks like we just need lint fix

PR #320, opened 2026-02-04
feat(rm): add --all option to remove all pieces from DataSet

beck-8:
```
# Delete all pieces (confirmation required)
filecoin-pin rm --all --data-set 123
# Skip confirmation and delete directly
filecoin-pin rm --all --data-set 123 --force
```

other: 
- fix spinner hang issue in rm commands
- fix not being able to delete data with CDN turned on

### DIFF
--- a/src/commands/rm.ts
+++ b/src/commands/rm.ts
@@ -1,36 +1,34 @@
-import { Command } from 'commander'
+import { Command, Option } from 'commander'
 import pc from 'picocolors'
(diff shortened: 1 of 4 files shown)


beck-8:
Usage Example

https://github.com/user-attachments/assets/d6a3ed00-d472-43c9-82fe-5f1fe92fb1c1

SgtPooki (approved):
I know we've been waiting on this for a while, I think it's good enough to get it in now

Issue #321, opened 2026-02-04
`Fetching data set details` too slow


beck-8:
I just wanted to take a look at this data set and it took him almost 1 minute. Very impact on user experience
```
➜  filecoin-pin git:(feat/rm-all-pieces) ✗ time NODE_DEBUG=fetch node  dist/cli.js --no-update-check dataset show 11170 --rpc-url http://127.0.0.1:1234/rpc/v1
┌  Filecoin Onchain Cloud Data Set Details for #11170
│
◑  Fetching data set detailsFETCH 1201884: connecting to pdp.660688.xyz:8443:8443 using https:undefined
FETCH 1201884: connecting to pdp.660688.xyz:8443:8443 using https:undefined
◒  Fetching data set detailsFETCH 1201884: connected to pdp.660688.xyz:8443:8443 using https:h1
FETCH 1201884: connected to pdp.660688.xyz:8443:8443 using https:h1
FETCH 1201884: sending request to GET https://pdp.660688.xyz:8443/pdp/data-sets/11170
FETCH 1201884: sending request to GET https://pdp.660688.xyz:8443/pdp/data-sets/11170
FETCH 1201884: received response to GET https://pdp.660688.xyz:8443/pdp/data-sets/11170 - HTTP 200
FETCH 1201884: received response to GET https://pdp.660688.xyz:8443/pdp/data-sets/11170 - HTTP 200
FETCH 1201884: trailers received from GET https://pdp.660688.xyz:8443/pdp/data-sets/11170
FETCH 1201884: trailers received from GET https://pdp.660688.xyz:8443/pdp/data-sets/11170
◇  ━━━ Data Set ━━━
│
│  Network: calibration
│  Client address: 0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│
│  #11170
│    Status: live
│    CDN add-on: disabled
│
│    Provider
│      ID: 6
│      Address: 0x86d026029052c6582d277d9b28700Edc9670B150
│      Name: beck-calib
│      Description: beck-calib
│      Service URL: https://pdp.660688.xyz:8443
│      Active: yes
│      Location: C=CN;ST=Zhejiang;L=Hangzhou
│
│    Metadata
│      none
│
│    Payment
│      PDP rail ID: 12223
│      Payer: 0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│      Payee: 0x86d026029052c6582d277d9b28700Edc9670B150
│
[...]

PR #322, opened 2026-02-04
perf(storage): optimize dataset fetching with O(1) lookup

beck-8:
https://github.com/filecoin-project/filecoin-pin/issues/321

- Refactor getDetailedDataSet to use direct WarmStorageService.getDataSet() instead of fetching all datasets via listDataSets()
- Add fast path in createStorageContext when dataSetId is explicitly provided, using createStorageContextFromDataSetId helper
- Directly instantiate StorageContext to bypass SDK's expensive resolveByDataSetId which makes redundant RPC calls
- Fetch metadata and provider info in parallel for better performance

This significantly speeds up 'dataset show <id>' and 'add --data-set <id>' commands by avoiding O(N) dataset scanning and redundant RPC calls.

This was generated for me by claude, and I tested it locally without any problems. If there is anything that does not comply with the specifications, you can directly submit and modify my PR.

### DIFF
--- a/src/core/synapse/index.ts
+++ b/src/core/synapse/index.ts
@@ -506,47 +506,14 @@ export async function createStorageContext(
     // The storage context manages the data set and provider interactions
     logger?.info?.({ event: 'synapse.storage.create' }, 'Creating storage context')
 
-    // FAST PATH: If a specific dataSetId is provided, use the optimized helper
-    // This avoids SDK's expensive resolveByDataSetId which makes redundant RPC calls
-    if (options?.dataset?.useExisting != null) {
-      const { createStorageContextFromDataSetId } = await import('./storage-context-helper.js')
-
-      logger?.info?.(
-        {
-          event: 'synapse.storage.dataset.existing',
-          dataSetId: options.dataset.useExisting,
-        },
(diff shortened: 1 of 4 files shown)


src/core/synapse/index.ts
```diff
@@ -506,22 +506,50 @@ export async function createStorageContext(
...
-    const sdkOptions: StorageServiceOptions = {
-      ...DEFAULT_STORAGE_CONTEXT_CONFIG,
-    }
-
-    // Apply dataset options
+    // FAST PATH: If a specific dataSetId is provided, use the optimized helper
+    // This avoids SDK's expensive resolveByDataSetId which makes redundant RPC calls
     if (options?.dataset?.useExisting != null) {
-      sdkOptions.dataSetId = options.dataset.useExisting
+      const { createStorageContextFromDataSetId } = await import('./storage-context-helper.js')
```
SgtPooki: we should just import at the top

SgtPooki (commented):
i'm not sure this is needed.. and we will be upgrading synapse-sdk soon. I think if there are performance issues, we should be updating synapse-sdk to do the right thing. That helper I added months ago was merged into synapse-sdk a while ago and we probably need to remove it.

Issue #328, opened 2026-02-11
executeUpload and uploadToSynapse should accept and handle abortSignals


SgtPooki:
see description. This is needed for https://github.com/FilOzone/dealbot/issues/258 to fully propogate

PR #332, opened 2026-02-12
fix: executeUpload and uploadToSynapse to use abortSignals

entuziaz:
Fixes issue #328 

- Add optional `signal` to `UploadExecutionOptions` and `SynapseUploadOptions`
- Perform early abort checks to fail fast

### DIFF
--- a/src/core/upload/index.ts
+++ b/src/core/upload/index.ts
@@ -268,6 +268,7 @@ export async function executeUpload(
 
           // Start validation (runs in parallel with other operations)
           ipniValidationPromise = waitForIpniProviderResults(rootCid, validationOptions).catch((error) => {
+            validationOptions.signal?.throwIfAborted()
             logger.warn({ error }, 'IPNI provider results check was rejected')
             return false
           })


SgtPooki (changes requested):
There are some fixes that need addressed. Also, basic best practices for abort are to check if signal is aborted:

1. prior to starting heavy work
2. after any heavy/long async work where `signal` is not respected
3. inside any loops within async work

src/core/upload/synapse.ts
```diff
@@ -138,8 +146,16 @@ export async function uploadToSynapse(
...
 
+  //  Only include `signal` when defined to satisfy `exactOptionalPropertyTypes`
+  if (options.signal != null) {
+    uploadOptions.signal = options.signal
+  }
+
   const synapseResult = await synapseService.synapse.storage.upload(carData, uploadOptions)
 
+  // Check if aborted after upload completes
+  options.signal?.throwIfAborted()
```
SgtPooki: we shouldn't need the abort here since the work is done and there is no further async work to do.

PR #327, opened 2026-02-09
Update to synapse 0.38.0. Closes #314

juliangruber:
Closes #314

PR #343, opened 2026-02-27
feat: upgrade to synapse-sdk 0.38 and synapse-core 0.2

rvagg:
Replace ethers.js with viem, strip wrapper layer down to CLI config mapping, update all commands and tests for new SDK API surface.

Similar to https://github.com/filecoin-project/filecoin-pin/pull/327, goes a bit further in ripping some things out though.

The aim here is to _make it work_, and it doesn't go the whole way that we need. The remaining work could be punted to follow-up PRs and we could just push and release a 0.38 version and then a multi-copy version.

**We are missing:**

* Multi-copy upload path: `uploadToSynapse()` still returns single-provider result, still passes explicit storage context to upload. Needs redesign for `copies[]`/`failures[]`, remove context passing (let SDK manage), update progress events
* CLI display: `displayUploadResults()`, spinner stages, add/import commands all still show single-provider output. Blocked  
* Devnet support (not critical but helpful)

I also haven't done extensive testing of this, it's a pretty rough pass to be honest so I'm marking it as draft for now. Maybe @SgtPooki has time to take a go at it, otherwise I'll get back to it next week.

### DIFF
--- a/src/add/add.ts
+++ b/src/add/add.ts
@@ -171,6 +171,7 @@ export async function runAdd(options: AddOptions): Promise<AddResult> {
       spinner,
       ...(pieceMetadata && { pieceMetadata }),
       ...(dataSetMetadata && { metadata: dataSetMetadata }),
+      ...(options.count != null && { count: options.count }),
     }
     if (contextSelection.providerIds) {
       uploadOptions.providerIds = contextSelection.providerIds
--- a/src/commands/add.ts
+++ b/src/commands/add.ts
@@ -10,6 +10,7 @@ export const addCommand = new Command('add')
   .argument('<path>', 'Path to the file or directory to add')
   .option('--bare', 'Add file without directory wrapper (files only, not supported for directories)')
   .option('--auto-fund', `Automatically ensure minimum ${MIN_RUNWAY_DAYS} days of runway before upload`)
+  .option('--count <n>', 'Number of storage copies to create (default: 2)', Number.parseInt)
 
 addCommand.action(async (path: string, options: any) => {
   try {
--- a/src/common/upload-flow.ts
+++ b/src/common/upload-flow.ts
@@ -463,7 +463,7 @@ export function displayUploadResults(
   if (result.size != null) {
     log.indent(`Piece Size: ${formatFileSize(result.size)}`)
   }
-  log.indent(`Explorer: ${pc.gray(`https://pdp.vxb.ai/${network}/piece/${result.pieceCid}`)}`)
+  log.indent(`Explorer: ${pc.gray(`https://pdp.vxb.ai/${encodeURIComponent(network)}/piece/${result.pieceCid}`)}`)
   log.line('')
 
   if (result.copies.length > 0) {
--- a/src/add/types.ts
+++ b/src/add/types.ts
@@ -6,6 +6,8 @@ export interface AddOptions extends CLIAuthOptions {
   bare?: boolean
   /** Auto-fund: automatically ensure minimum 30 days of runway */
   autoFund?: boolean
+  /** Number of storage copies to create */
+  count?: number
   /** Piece metadata attached to each upload */
   pieceMetadata?: Record<string, string>
   /** Data set metadata applied when creating or updating the storage context */
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -9,6 +9,7 @@ export const importCommand = new Command('import')
   .description('Import an existing CAR file to Filecoin via Synapse')
   .argument('<file>', 'Path to the CAR file to import')
   .option('--auto-fund', `Automatically ensure minimum ${MIN_RUNWAY_DAYS} days of runway before upload`)
+  .option('--count <n>', 'Number of storage copies to create (default: 2)', Number.parseInt)
   .action(async (file: string, options) => {
     try {
       const {
--- a/src/import/import.ts
+++ b/src/import/import.ts
@@ -217,6 +217,7 @@ export async function runCarImport(options: ImportOptions): Promise<ImportResult
       spinner,
       ...(pieceMetadata && { pieceMetadata }),
       ...(dataSetMetadata && { metadata: dataSetMetadata }),
+      ...(options.count != null && { count: options.count }),
     }
     if (contextSelection.providerIds) {
       uploadOptions.providerIds = contextSelection.providerIds
--- a/src/import/types.ts
+++ b/src/import/types.ts
@@ -5,6 +5,8 @@ export interface ImportOptions extends CLIAuthOptions {
   filePath: string
   /** Auto-fund: automatically ensure minimum 30 days of runway */
   autoFund?: boolean
+  /** Number of storage copies to create */
+  count?: number
   /** Piece metadata attached to the imported CAR */
   pieceMetadata?: Record<string, string>
   /** Data set metadata applied when creating or updating the storage context */
--- a/src/rm/remove-all-pieces.ts
+++ b/src/rm/remove-all-pieces.ts
@@ -176,12 +176,17 @@ export async function runRmAllPieces(options: RmAllPiecesOptions): Promise<RmAll
     )
 
     // Display results
-    log.spinnerSection('Results', [
+    const resultLines = [
       pc.gray(`Total Pieces: ${result.totalPieces}`),
       pc.gray(`Removed: ${result.removedCount}`),
       pc.gray(`Failed: ${result.failedCount}`),
       pc.gray(`Network: ${network}`),
-    ])
+    ]
+    const failures = result.transactions.filter((t) => !t.success)
+    for (const f of failures) {
+      resultLines.push(pc.red(`  ${f.pieceCid}: ${f.error ?? 'unknown error'}`))
+    }
+    log.spinnerSection('Results', resultLines)
 
     if (result.failedCount > 0) {
       outro(`Remove completed with ${result.failedCount} failure(s)`)


package.json
```diff
@@ -112,35 +112,35 @@
...
     "picocolors": "^1.1.1",
-    "pino": "^10.0.0",
-    "semver": "^7.6.3"
+    "pino": "^10.3.1",
+    "semver": "^7.7.4",
+    "viem": "^2.46.3"
   },
   "devDependencies": {
-    "@biomejs/biome": "2.4.4",
+    "@biomejs/biome": "^2.4.4",
```
rvagg:
```suggestion
    "@biomejs/biome": "2.4.4",
```

src/add/add.ts
```diff
@@ -227,27 +199,33 @@ export async function runAdd(options: AddOptions): Promise<AddResult> {
...
+    // Exit non-zero if any copies failed
+    if (uploadResult.failures.length > 0) {
+      log.line('')
+      log.line(
+        pc.yellow(
+          `${pc.yellow('⚠')} ${uploadResult.failures.length} copy failure(s). Data is stored but with reduced redundancy.`
+        )
+      )
+      log.flush()
+      outro('Add completed with warnings')
```
juliangruber:
```suggestion
      outro('Add completed with errors')
```

I think either say warnings and exit with 0, or errors and exit with 1

juliangruber (changes requested):

src/core/data-set/list-data-sets.ts
```diff
@@ -34,35 +33,29 @@ import type { DataSetSummary, ListDataSetsOptions } from './types.js'
...
-      providerMap = new Map(providers.map((provider) => [provider.id, provider]))
-    } catch (error) {
-      logger?.warn({ error }, 'Failed to fetch provider info from sp-registry for provider enrichment')
+  // Fetch provider info for unique provider IDs
+  let providerMap: Map<bigint, PDPProvider> = new Map()
+  if (withProviderDetails) {
+    const uniqueProviderIds = Array.from(new Set(filteredDataSets.map((ds) => ds.providerId)))
+    if (uniqueProviderIds.length > 0) {
+      try {
+        const providers = await synapse.providers.getProviders({ providerIds: uniqueProviderIds })
```
juliangruber: nit:

```suggestion
    const uniqueProviderIds = new Set(filteredDataSets.map((ds) => ds.providerId))
    if (uniqueProviderIds.size > 0) {
      try {
        const providers = await synapse.providers.getProviders({ providerIds: Array.from(uniqueProviderIds) })
```

src/commands/add.ts
```diff
@@ -11,18 +11,6 @@ export const addCommand = new Command('add')
...
-addCommand.addOption(
-  new Option('--data-set <id>', 'ID of the existing data set to use').conflicts(['newDataSet', 'new-data-set'])
-)
-
-addCommand.addOption(
-  new Option('--new-data-set', 'Create a new data set instead of using an existing one').conflicts([
-    'dataSet',
-    'data-set',
-  ])
-)
```
SgtPooki: is there still a way to add a piece to a new dataset?

src/common/upload-flow.ts
```diff
@@ -285,56 +302,79 @@ export async function performUpload(
...
+            `${roleLabel('secondary')} Failed: provider ${event.data.providerId} - ${event.data.error.message}`,
+            { type: 'warning' }
+          )
+          break
+        }
+        case 'onPiecesAdded': {
+          const role = getRole(event.data.providerId)
+
+          const commitId = `commit-${event.data.providerId}`
+          flow.addOperation(commitId, `${roleLabel(role)} Adding piece to DataSet...`)
```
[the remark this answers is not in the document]
SgtPooki: We have kind of had it all over the place, but "two word" data-set variations are all good. I think DataSet makes sense and is used elsewhere, but also:

* data-set
* Data-set
* dataSet
* Data Set

but not "dataset" nor "DATASET" (ive used DATASET in env vars in dealbot tho)

src/common/upload-flow.ts
```diff
@@ -285,56 +302,79 @@ export async function performUpload(
...
+
+          // Show per-SP transaction URL as indented line under the "added" message
+          const afterLines: string[] = []
```
SgtPooki:
```suggestion
          const spSuffixLines: string[] = []
```

src/core/data-set/get-data-set-pieces.ts
```diff
@@ -55,7 +40,6 @@ export async function getDataSetPieces(
 ): Promise<DataSetPiecesResult> {
   const logger = options?.logger
   const includeMetadata = options?.includeMetadata ?? false
-  const signal = options?.signal
```
SgtPooki: it would be nice if we could keep abort handling intact, but maybe we can add a follow-up issue. There's a lot of synapse still not supporting abortable operations (instead accepting a timeout value) which makes it really hard to cap a nest of loop & wait operations under the hood

SgtPooki (commented):
some comments.. 22/88 files reviewed.. will hit this again another time.. 

this PR and your demos make me really want a full "e2e" test suite that spins up new wallet(s) automatically, and runs through all the possible commands. We should be able to do that with foc-devnet right?

src/common/upload-flow.ts
```diff
@@ -246,31 +250,44 @@ function displayPaymentIssues(capacityCheck: PaymentCapacityCheck, fileSize: num
...
   const flow = createSpinnerFlow(spinner)
 
   // Start with upload operation
   flow.addOperation('upload', 'Uploading to Filecoin...')
 
-  let transactionHash: string | undefined
+  // Track primary provider ID from onStored to label subsequent events
+  let primaryProviderId: bigint | undefined
+
+  function getRole(providerId: bigint): 'primary' | 'secondary' {
```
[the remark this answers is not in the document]
rvagg: Not comitted to this terminology but the primary is going to be the endorsed SP, so there's some importance to it. We could say "endorsed" and .. what? "other", but the secondary SP could also be endorsed, it was just chosen at random from the approved. Happy to use alternative names here but there is a _primary_ because (1) it's going to be an endorsed SP, blessed by us as super-special, and (2) it's the one that we send our data to, and the only one that we send our data to, all other SPs in our list (there could be more than 2 SPs) have to pull from that first one.

So do we want to leak "endorsed" here? and what are we saying about the secondaries and their endorsement status? I don't want to suggest they're not endorsed because they may be (we could check, but for what?).

src/common/upload-flow.ts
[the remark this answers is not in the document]
rvagg: yeah, sorry about this one slipping through, I've been the "data set" (two word) pedant around here!

src/add/add.ts
```diff
@@ -116,28 +110,36 @@ export async function runAdd(options: AddOptions): Promise<AddResult> {
...
     spinner.stop(`${pc.green('✓')} ${pathType} validated${sizeDisplay}`)
 
-    // Initialize Synapse SDK (without storage context)
+    // Validate context selection options early (before expensive operations)
+    let contextSelection: ReturnType<typeof parseContextSelectionOptions>
+    try {
+      contextSelection = parseContextSelectionOptions(options)
+    } catch (error) {
+      cancel(error instanceof Error ? error.message : 'Invalid options')
+      process.exit(1)
```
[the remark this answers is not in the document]
rvagg: IIRC there was a reason for this, I think it was the inability to clean up viem's websocket connection properly. If I remove this, the process hangs in multiple places. I tried hard to make it work like we had for the old "cleanup" that tore down the ethers websocket thing, but I couldn't make the same thing work for viem. Might need more investigation although I did spend a bit of time on this.

src/common/get-rpc-url.ts
```diff
@@ -13,27 +18,27 @@ import type { CLIAuthOptions } from '../utils/cli-auth.js'
...
-    return rpcUrl
+    return wsUrl
   }
 
-  return RPC_URLS.calibration.websocket
+  const defaultUrl = calibration.rpcUrls.default.webSocket?.[0] ?? calibration.rpcUrls.default.http[0]
+  if (!defaultUrl) {
+    throw new Error('No RPC URL available for calibration network')
+  }
+  return defaultUrl
```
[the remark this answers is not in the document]
rvagg: no, it's fine, either way should work for us here

src/common/upload-flow.ts
```diff
@@ -48,14 +41,28 @@ export interface UploadFlowOptions {
...
+  providerIds?: bigint[]
+
+  /** Specific data set IDs to use. */
+  dataSetIds?: bigint[]
+
+  /** Provider IDs to exclude from selection. */
+  excludeProviderIds?: bigint[]
+
+  /** Data set metadata applied when creating or matching contexts. */
+  metadata?: Record<string, string>
```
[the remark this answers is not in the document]
rvagg: ^ that :+1:
must match, won't get updated, that would be unexpected behaviour

src/common/upload-flow.ts
```diff
@@ -246,31 +250,44 @@ function displayPaymentIssues(capacityCheck: PaymentCapacityCheck, fileSize: num
...
 
   // Start with upload operation
   flow.addOperation('upload', 'Uploading to Filecoin...')
 
-  let transactionHash: string | undefined
+  // Track primary provider ID from onStored to label subsequent events
+  let primaryProviderId: bigint | undefined
+
+  function getRole(providerId: bigint): 'primary' | 'secondary' {
+    if (primaryProviderId == null || providerId === primaryProviderId) {
```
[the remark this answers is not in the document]
rvagg: The `== null` guard is correct -- before `onStored` fires, `primaryProviderId` is `undefined`; only when we get a successful `onStored` can we set it, the primary SP is the one we "store" on, i.e. send our data to, secondaries all do a "pull".

src/core/payments/index.ts
```diff
@@ -260,33 +277,33 @@ export async function depositUSDFC(
...
       amount,
-      synapse.getWarmStorageAddress(),
-      MAX_RATE_ALLOWANCE,
-      MAX_LOCKUP_ALLOWANCE,
-      BigInt(DEFAULT_LOCKUP_DAYS) * TIME_CONSTANTS.EPOCHS_PER_DAY
-    )
+      operator: synapse.chain.contracts.fwss.address,
+      rateAllowance: MAX_RATE_ALLOWANCE,
+      lockupAllowance: MAX_LOCKUP_ALLOWANCE,
+      maxLockupPeriod: BigInt(DEFAULT_LOCKUP_DAYS) * TIME_CONSTANTS.EPOCHS_PER_DAY,
```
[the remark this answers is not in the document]
rvagg: yeah, I'm just going to hand-wave this away as out of scope; we could re-litigate separately but the aim in here is to be as human-friendly as we can
also, the contract deals in 30-days, not months, so that gets awkward pretty quick

src/core/data-set/get-data-set-pieces.ts
rvagg: hah .. did we remove `signal` from `getPieces()` or was it never there? I glossed over this and didn't think about it tbh..z

src/common/upload-flow.ts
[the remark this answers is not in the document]
rvagg: https://github.com/FilOzone/synapse-sdk/issues/653

introducing it here but it'd be good if Synapse exported this.

src/core/data-set/types.ts
```diff
@@ -93,7 +94,7 @@ export interface DataSetSummary extends EnhancedDataSetInfo {
  */
 export interface ListDataSetsOptions {
   /** Address to list datasets for (defaults to synapse client address) */
-  address?: string
+  address?: `0x${string}`
```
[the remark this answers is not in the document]
rvagg: good catch

src/add/add.ts
rvagg: fwiw I've now refactored this code because I realised that I'd broken the recommendation that _I wrote_ about handling this (which points to the complexity here). I reworked the failure check to compare `copies.length` against `requestedCopies`. There's three outcomes:
* Copies < requested: exitCode = 1, message says "completed with errors" and reports X/Y copies
* Copies met but failures exist: exit 0, grey informational note about non-critical failures (i.e. legit to report, but we got your copies stored)
* No failures: exit 0, "completed successfully"

src/core/upload/index.ts
```diff
@@ -200,42 +218,61 @@ export interface UploadExecutionOptions {
      */
     enabled?: boolean
   } & Omit<WaitForIpniProviderResultsOptions, 'onProgress'>
+
+  /** Number of storage copies to create (default determined by SDK). */
+  count?: number
```
juliangruber: similar to above, use `copies` instead?

src/core/upload/synapse.ts
```diff
@@ -28,149 +35,248 @@ export interface SynapseUploadOptions {
...
+      onCopyFailed: (providerId, pieceCid, error) => {
+        logger.warn(
+          {
+            event: 'synapse.upload.copy_failed',
+            contextId,
+            providerId: String(providerId),
+            pieceCid: pieceCid.toString(),
+            error: error.message,
+          },
+          'Secondary copy failed'
```
juliangruber: Maybe we can also call the replicas "upload" & "copy", so don't call the endorsed-SP-stored thing a copy. For user facing communication, that might work.

Gemini had this to say:

Image

src/cli.ts
```diff
@@ -71,6 +71,13 @@ program.hook('postAction', async () => {
     console.log(header)
     console.log(instruction)
   }
+
+  // Viem transports (especially WebSocket) hold persistent connections that
+  // prevent the process from exiting. The server command manages its own
+  // lifecycle via SIGINT/SIGTERM handlers, so only exit for CLI commands.
+  if (actionCommand.name() !== 'server') {
+    process.exit(0)
```
juliangruber:
```suggestion
    process.exit(process.exitCode)
```

Might lead to less accidental swallowing of non-zero exit codes?

juliangruber (changes requested):

juliangruber (approved):
Great work!

PR #353, opened 2026-03-06
fix: support mainnet CLI network and payment guidance

### LINKED ISSUE #352: Mainnet CLI network selection and USDFC guidance regressions
Two regressions need follow-up:

- `--mainnet` / mainnet network selection is not propagated consistently through CLI auth and RPC resolution
- mainnet users without USDFC only see the minting guide instead of the bridge and swap paths that are now recommended

davidgasquez:
Fixes #352

- propagate `--mainnet` and network selection through CLI auth and RPC resolution
- reuse a single helper for mainnet USDFC acquisition guidance in validation and `payments status`
- make config defaults tests independent from ambient `NETWORK` and `RPC_URL`

Validation:
```sh
npm test
# Test Files  35 passed (35)
# Tests  336 passed | 11 skipped (347)
```

### DIFF
--- a/src/utils/cli-auth.ts
+++ b/src/utils/cli-auth.ts
@@ -6,7 +6,7 @@
  */
 
 import type { Chain, Synapse } from '@filoz/synapse-sdk'
-import { getConfiguredChain, getRpcUrl } from '../common/get-rpc-url.js'
+import { getRpcUrl, NETWORK_CHAINS } from '../common/get-rpc-url.js'
 import type { SynapseSetupConfig } from '../core/synapse/index.js'
 import { initializeSynapse } from '../core/synapse/index.js'
 import { createLogger } from '../logger.js'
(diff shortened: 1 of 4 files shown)


src/utils/cli-auth.ts
```diff
@@ -26,6 +26,8 @@ export interface CLIAuthOptions {
   viewAddress?: string | undefined
   /** Filecoin network: mainnet or calibration */
   network?: string | undefined
+  /** Commander shorthand for --network mainnet */
+  mainnet?: boolean | undefined
```
rvagg:
```suggestion
```

rvagg:
I pushed another commit that removes a few things, commander's `implies()` does most of the validation work for us with `--mainnet` and `--network` so we shouldn't be double-validating. Passing on the `chain` is the main fix here, the USDFC additions are very nice, and the env cleanup is good too.

Issue #348, opened 2026-03-05
Give option to skip content routing checks


BigLep:
## Done Criteria
There is a way to add a piece and not block on verifying to see that the ipfsRootCid is indexed by filecoinpin.contact.

## Why Important
This is particularly for foc-devnet testing where want to do E2E testing of FOC (https://github.com/FilOzone/foc-devnet/issues/8).  We don't have a mocked IPNI, so just want to skip the step.
 
## User/Customer
Developers creating  foc-devnet tests

## Notes
- Curio will be updated to skip IPNI advertisment announcing in a devnet with https://github.com/filecoin-project/curio/issues/1073

PR #300, opened 2026-01-09
docs: created DEVELOPMENT.md with some tips

BigLep:
This includes some info I would have wanted when working on https://github.com/FilOzone/filecoin-services/issues/380 (or it includes the headers I wish were there).

PR #354, opened 2026-03-09
feat: add devnet network support and --skip-ipni flag

rvagg:
* Support `--network devnet` using foc-devnet's devnet-info.json via synapse-core's `toChain()`. Private key auto-resolves from devnet user.
* Add `--skip-ipni` flag for add/import commands, auto-enabled on devnet.

Closes: https://github.com/filecoin-project/filecoin-pin/issues/348

image

^ URLs aren't going to work there obviously .. but I'm not sure it's worth chasing those down and turning them off, hopefully any devnet user will know that they are meaningless.


rvagg:
> could be useful to add some notes to DEVELOPMENT.md

stalled in #300, I've added to the README for now

PR #359, opened 2026-03-13
chore: 7day dep update cooldown, use `npm ci` for stable install in action

rvagg:
Dialling up the risk aversion slightly here.


rvagg:
~GitHub~ dependabot shows up unasked when there's a CVE or general GitHub security notice out

PR #369, opened 2026-03-19
feat!: update to synapse-sdk@0.40.0

SgtPooki:
This PR aligns filecoin-pin with the breaking upload API changes introduced in `synapse-sdk@0.40.x` and the merged Synapse follow-up work.

On the filecoin-pin side, the public upload interface now matches Synapse directly: upload options use copies instead of count, upload results use failedAttempts instead of failures, and the old FailedCopy compatibility surface is removed in favor of FailedAttempt. The CLI has been updated accordingly, so add and import now use --copies. We also re-export the relevant Synapse types needed by downstream consumers.

Internally, this PR updates the upload pipeline and tests to the new SDK shape, passes the now-required source when creating Synapse clients, and aligns @filoz/synapse-core to the version expected by synapse-sdk@0.40.x to avoid duplicate type conflicts. Mocks, unit tests, and type definitions were updated to reflect the new names and behavior.

replaces #357
related https://github.com/FilOzone/synapse-sdk/pull/664

### DIFF
--- a/src/common/upload-flow.ts
+++ b/src/common/upload-flow.ts
@@ -268,7 +268,7 @@ function roleLabel(role: CopyRole): string {
  * @param carData - CAR file data as Uint8Array
  * @param rootCid - Root CID of the content
  * @param options - Upload flow options
- * @returns Upload result with copies and completion status
+ * @returns Upload result with copies and network information
  */
 export async function performUpload(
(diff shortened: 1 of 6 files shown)


rvagg:
also thanks for doing this, I should have got here first but I've been distracted

PR #370, opened 2026-03-19
fix: use pnpm

SgtPooki:
migrate filecoin-pin to use pnpm. This addresses an issue with npm not supporting the secure npm package age configuration seen in https://github.com/filecoin-project/filecoin-pin/actions/runs/23318010434/job/67822423276?pr=369

We were trying to update to the latest synapse-sdk versions, but they were not older than 7 days, and we couldn't exclude our own packages from the update, we got an error like this:

```shell
Run npm install --no-progress
npm error code ETARGET
npm error notarget No matching version found for @filoz/synapse-core@^0.3.1 with a date before 3/12/2026, 9:34:43 PM.
npm error notarget In most cases you or one of your dependencies are requesting
npm error notarget a package version that doesn't exist.
npm error A complete log of this run can be found in: /Users/runner/.npm/_logs/2026-03-19T21_34_43_586Z-debug-0.log
```

so now we use pnpm-workspace.yaml to configure safe release age, and exclude our `@filoz/*` pacakges. Still an attack vector, but still safer than removing entirely and locking us out of hotfixes.

related #357
related #369 
related https://github.com/FilOzone/synapse-sdk/pull/664

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,8 @@ jobs:
           cache-dependency-path: pnpm-lock.yaml
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
+      - name: Install Playwright browsers
+        run: pnpm exec playwright install chromium
       - name: Run Lint
         run: pnpm run lint
       - name: Run Tests


juliangruber:
I thought I had broken the PR, but just reran the original commit actions, which are also now failing

PR #383, opened 2026-03-25
refactor: optimize actual storage calculation to use PDP leaf counts

### LINKED ISSUE #281: fix: calculate-actual-storage.ts can be improved.
From @rvagg on slack:

> FYI re upload size accounting (https://github.com/filecoin-project/filecoin-pin/pull/276) you do have the option of taking the super-quick route if you want, and I think you might be able to do it in a way that doesn’t lose functionality:
> 1. Get clientDataSetIds from FWSS
> 2. For each data set, in parallel, get dataSetLeafCount from PDPVerifier - then you have the total byte size for a data set
>   * OR if you still want per-piece information, PDPVerifier has a pieceLeafCounts that you can get per data set — or maybe even better, there is a pieceCids per data set that you can get with a single call
> 
> We could put calls for these on PDPVerifier in Synpase if we don’t have them already.

ameeetgaikwad:
Closes #281

calculateActualStorage() was computing total stored bytes by creating a storage context for each dataset,
fetching every piece, and summing piece sizes. That path was doing much more work than needed for an aggregate storage total.

This change switches the calculation to use PDPVerifier.getDataSetLeafCount(dataSetId) and converts the result to bytes with PDP_LEAF_SIZE. This keeps the same public result shape and progress/abort behavior, but avoids the per-piece retrieval path.

### DIFF
--- a/src/core/data-set/calculate-actual-storage.ts
+++ b/src/core/data-set/calculate-actual-storage.ts
@@ -1,9 +1,9 @@
-import type { Synapse } from '@filoz/synapse-sdk'
+import { PDPVerifier, type Synapse, WarmStorageService } from '@filoz/synapse-sdk'
 import PQueue from 'p-queue'
 import type { Logger } from 'pino'
+import { PDP_LEAF_SIZE } from '../payments/constants.js'
 import { getClientAddress } from '../synapse/index.js'
 import type { ProgressEvent, ProgressEventHandler, Warning } from '../utils/types.js'
-import { getDataSetPieces } from './get-data-set-pieces.js'
 import type { DataSetSummary } from './types.js'
 
 export interface ActualStorageResult {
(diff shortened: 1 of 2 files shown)


rvagg:
@ameeetgaikwad this is close but not quite - see https://github.com/FilOzone/filecoin-services/issues/451 where I discuss the FR32 expansion, so leaf count x 32 inflates the actual size by a little bit. The fix to that would be `const dataSetBytes = (BigInt(leafCount) * BigInt(PDP_LEAF_SIZE) * 127n) / 128n`

But, I think now the right approach might be to use the new `getActivePieceCids` iterator from Synapse and then getting the actual size out of the PieceCID itself, Synapse has utilities to decode a CID and give you the byte size of it.

So this is roughly correct, but looking back at #281 and my comments there, I think the third option I mentioned might be the right one. There'll be a newer Synapse release soon that has a better iterating `getActivePieceCids` type interface, I think the current one might work well enough but newer will be better, we just haven't triggered a release there yet.

Are you still able to take this on?

BigLep:
I'm putting this back to todo until someone raises their hand to drive it to completion.

shrirajpawar4:
@BigLep @rvagg 
Opened a follow-up PR with the PieceCID-based correction here #488 

 It keeps @ameeetgaikwad’s original commit as I could not add on to his branch and adds the change to sum raw sizes from active PieceCIDs via `getSizeFromPieceCID()` instead of using FR32-expanded leaf counts.

  `pnpm test` passes locally.

ameeetgaikwad:
thanks @shrirajpawar4 for picking this up

BigLep:
Thanks. Closing since superceded by https://github.com/filecoin-project/filecoin-pin/pull/488

PR #386, opened 2026-03-27
fix(identify): handle many origins

wjmelements:
Reviewer @SgtPooki
On linux I saw this issue with the server and kubo pinmfs related to there being too many origins.
There were 54 origins.
The repeated dial was causing problems and deduping fixed it.
There was also a problem related to the size of the identify response, which was too large.
#### Changes
* larger identify message size limit
* ~dedupe origins~
* pass all origins to dial

### DIFF
--- a/src/create-pinning-helia.ts
+++ b/src/create-pinning-helia.ts
@@ -32,7 +32,17 @@ export async function createPinningHeliaNode(options: PinningHeliaOptions): Prom
 }> {
   const { logger, rootCID, outputPath, origins = [] } = options
 
-  const dialAddrs = origins.map(multiaddr)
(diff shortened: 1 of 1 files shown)


SgtPooki (approved):
quick look before vacation. it sounds like we need this fix, but we might need a little more validation that the multiaddr we're selecting will work.. maybe check with @juliangruber and @rvagg ?

this pinning server hasn't received a ton of attention since @rvagg created it.. so he might be the best one to look, but i'm good with these changes, as long as the risk i pointed out in comments is acceptable (or i'm wrong)

rvagg (approved):
cool

Issue #395, opened 2026-04-01
Upload action: add CI to verify functionality


BigLep:
## Done Criteria
We have CI on PRs to validate that the upload action still functions.

## Why Important
The upload action is a demonstration of how to use Filecoin Pin and we do dogfood it in filecoin-pin-website. As a result, we want to make sure it functions, even if we have disclaimers that it is not supposed to be used for production usecases.

BigLep:
@juliangruber : did you have a proposal for how we would do this? We're thinking to punt this without some clear steps on what we'd do and assert.

juliangruber:
I already have a draft PR that should close this once merged. It tests a dry run upload and a real upload, and succeeds if the upload action passed. I think this is sufficient for now

juliangruber:
https://github.com/filecoin-project/filecoin-pin/pull/399

PR #400, opened 2026-04-02
fix: include sybil fees in shared upload funding

rjan90:
## Summary
This PR is the short-term mitigation for the recent `InsufficientLockupFunds` failures.

It keeps filecoin-pin's existing funding planner, but patches the shared upload funding path so both the GitHub Action and CLI auto-fund logic include the new-data-set 0.1 USDFC sybil fee.

## What changed
- add the WarmStorage new-data-set sybil fee constant to filecoin-pin payment constants
- extend the shared funding planner to accept `newDataSetCount` and include those fees in the planned deposit
- thread that input through `planFilecoinPayFunding()`
- update CLI auto-fund to resolve upload contexts and count how many new data sets will be created
- update the upload action to do the same and feed the result into the shared planner
- add targeted unit coverage for the planner and the add/import auto-fund call sites

## Why
The failure we hit was caused by funding only the lockup requirement for an upload while omitting the 0.1 USDFC fee charged when a new WarmStorage data set is created.

Because upload-time funding exists in both the upload action and CLI auto-fund flow, the short-term patch should live in the shared planner path rather than only in the action wrapper.

## Scope and tradeoff
This is intentionally the tactical fix.

It does **not** replace filecoin-pin's custom cost planner with Synapse's full upload-cost calculation. Instead, it patches the known missing fee in the existing planner so the immediate failure is addressed with a smaller change set.

The longer-term cleanup direction is in #401.

## Validation
- `pnpm run build`
- `pnpm --filter filecoin-pin-upload-action-helper typecheck`
- `pnpm exec vitest run --project unit src/test/unit/payments-funding.test.ts src/test/unit/add.test.ts src/test/unit/import.test.ts`
- `pnpm exec biome check src/core/payments/constants.ts src/core/payments/types.ts src/core/payments/funding.ts src/payments/types.ts src/payments/fund.ts src/common/upload-flow.ts src/add/add.ts src/import/import.ts src/test/unit/payments-funding.test.ts src/test/unit/add.test.ts src/test/unit/import.test.ts upload-action/src/filecoin.js`

### DIFF
--- a/src/add/add.ts
+++ b/src/add/add.ts
@@ -160,8 +160,21 @@ export async function runAdd(options: AddOptions): Promise<AddResult> {
     const carSize = carData.length
     spinner.stop(`${pc.green('✓')} IPFS content loaded (${formatFileSize(carSize)})`)
 
+    const autoFundOptions: Parameters<typeof performAutoFunding>[3] = {
+      ...(dataSetMetadata && { metadata: dataSetMetadata }),
+      ...(options.copies != null && { copies: options.copies }),
+    }
+    if (contextSelection.providerIds) {
+      autoFundOptions.providerIds = contextSelection.providerIds
+      autoFundOptions.copies = contextSelection.providerIds.length
+    }
+    if (contextSelection.dataSetIds) {
+      autoFundOptions.dataSetIds = contextSelection.dataSetIds
+      autoFundOptions.copies = contextSelection.dataSetIds.length
+    }
+
     if (options.autoFund) {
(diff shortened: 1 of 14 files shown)


rjan90:
@juliangruber feel free to take this over, or discard if its totally out of sync with what is needed to unblock https://github.com/filecoin-project/filecoin-pin/issues/395

rjan90:
Closing this draft, since some of the commits has been cherry-picked into https://github.com/filecoin-project/filecoin-pin/issues/395

PR #413, opened 2026-04-27
chore(ci): pin third-party actions to SHA, bump release-please-action v4 → v5

SgtPooki:
## Summary

- Bumps `googleapis/release-please-action` v4 → v5.0.0 (SHA-pinned). Fixes Node.js 20 deprecation warning seen in [recent runs](https://github.com/filecoin-project/filecoin-pin/actions/runs/25012872720). v5 only changes runtime to node24; no API changes.
- Pins `pnpm/action-setup@v5` to commit SHA (`fc06bc1...`) across `test.yml`, `docs.yml`, `release-please.yml` per security best practice for third-party actions.

Official `actions/*` actions are left as floating major version tags per GitHub convention.

### Out of scope (follow-ups)
- `pnpm/action-setup` v5 → v6 (v6 defaults to pnpm v11 — needs review)
- `actions/github-script` v8 → v9
- `actions/upload-pages-artifact` v4 → v5

## Test plan
- [x] CI passes on this PR (test.yml + docs.yml exercise pnpm/action-setup pin)
- [ ] Next push to master shows release-please job running on node24 without deprecation warning


rvagg:
@SgtPooki are you expressing a strong opinion here for hash based versioning? I really would prefer the simplicity of  tags for this. https://github.com/filecoin-project/filecoin-pin/pull/420 gives us proper version bumps plus the (minor) safety guard of a week's cooldown to help avoid obvious attacks which tend to be picked up very quickly.

juliangruber:
> @SgtPooki are you expressing a strong opinion here for hash based versioning? I really would prefer the simplicity of tags for this. #420 gives us proper version bumps plus the (minor) safety guard of a week's cooldown to help avoid obvious attacks which tend to be picked up very quickly.

Hash based versioning for GitHub actions is the safest path, together with updates from Dependabot

juliangruber (approved):

rvagg:
> Hash based versioning for GitHub actions is the safest path

That's not an answer tho, that's just an opinion. Hash based versioning is _very_ opaque, and that opacity introduces new risks, hence my push-back; just cause you're avoiding tags doesn't mean you're avoiding risk. Trade-offs everywhere. I vote for tags with a cooldown for a bit of extra safety. But I'm not going to block on this since this is just for the third-party one. Whatever.

juliangruber:
@rvagg how would the cooldown work when we target branches? We'd still get every new commit from that branch, right?

I should have been more clear, if you don't use commit references you don't have a lockfile and the code will change underneath. To me this is like using npm before package lock files. And I agree with the opacity, but dependabot updating hashes is good enough for me to trust the hash to be correct.

rvagg:
@juliangruber yes, I know, but when was the last time you upgraded your local pnpm by using an npm install from github using a git ref? If someone's got access to push tags, then they've likely got access to do much more sneaky things too. It's all a game of trade-offs here - by using a hash here you're playing a game of indirection from opaque to user-intent rather than the other way around, which is what tags give us. The fact that you need to `# v5` shows  that, "trust me bro, it's v5".
Not blocking concern, I just hate the DX of planting hashes. It means I have to go hunting to figure out what we're on or trust a comment is correct and in sync.

rvagg (approved):

rvagg:
ok, this bikeshedding is unproductive and I'm the cause of it

rvagg:
Coming back to this an acknowledging a mea culpa, I hadn't realised quite how easy it was to have automatic tooling (dependabot) keep sha's updated along with the comment and I'm coming around to the idea of just switching my stuff in bulk to sha's given that (and the fact that it's becoming trivial to verify the sha=version using AI and other tooling). So I take back my bickering above.

PR #418, opened 2026-04-28
fix(upload-action): repair pnpm caching broken by step ordering

### LINKED ISSUE #411: filecoin-pin-upload action failing due to pnpm
## Description

The `upload-action` on `master` (and `v0.20.0`) is failing when used in a GitHub action pipeline.

1. **Failing in a simple pipeline due to pnpm not being present:**
```
Error: Unable to locate executable file: pnpm. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable.
```

2. **If you try to pre-install pnpm in an earlier step (e.g. `npm install -g pnpm`):**
```
Error: Some specified paths were not resolved, unable to cache dependencies.
```

## The example filecoin-pin website deploy pipeline

I checked the example pipeline to see if it is executing without issue. But it has not triggered since Nov 2025. Meaning the latest changes have not been executed there.
- Action history https://github.com/filecoin-project/filecoin-pin-website/actions/workflows/filecoin-pin-upload.yml
- Action Definition https://github.com/filecoin-project/filecoin-pin-website/blob/main/.github/workflows/filecoin-pin-upload.yml

## Potential Root Cause

In `upload-action/action.yml`, the `Set up Node.js` step runs `actions/setup-node@v6` with `cache: pnpm` **before** the `Set up pnpm` step that actually installs pnpm via `pnpm/action-setup@v4`.

`setup-node` with `cache: pnpm` needs pnpm on PATH to run `pnpm store path --silent` for cache resolution. Since pnpm isn't installed yet at that point, it fails.

Even if you work around that by pre-installing pnpm globally, the second error occurs because `cache-dependency-path` points to the action's own `pnpm-lock.yaml` outside `GITHUB_WORKSPACE`, which `hashFiles()` cannot resolve.

Relevant lines in [`upload-action/action.yml#L132-L151`](https://github.com/filecoin-project/filecoin-pin/blob/0d104c83fd17a648256e7ded441e8d40f8102d72/upload-action/action.yml#L132-L151):
```yaml
- name: Set up Node.js
  uses: actions/setup-node@v6
  with:
    node-version: '24.x'
    cache: pnpm                          # <-- pnpm not yet installed
    cache-dependency-path: ${{ steps.repo-root.outputs.path }}/pnpm-lock.yaml  # <-- outside GITHUB_WORKSPACE

- name: Get pnpm version
  # ...

- name: Set up pnpm                     # <-- pnpm installed here, too late
  uses: pnpm/action-setup@v4
```

## Reproduction

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    permissions:            # copied from example repo https://github.com/filecoin-project/filecoin-pin-website/blob/main/.github/workflows/filecoin-pin-upload.yml
        contents: read
        actions: read        # Required for artifact reuse
        checks: write        # Required to create PR check status
    steps:
      - name: checkout code
        uses: actions/checkout@v6

      - name: Upload to Filecoin
        uses: filecoin-project/filecoin-pin/upload-action@master
        with:
          walletPrivateKey: ${{ secrets.WALLET_KEY }}
          path: _site
          network: calibration
          minStorageDays: "30"
          filecoinPayBalanceLimit: "5.00"
```

Tested with both `@master` and `@v0.20.0` on `ubuntu-latest`.

## Workaround

Pinning to `v0.18.0` executes, since that version uses npm before the switch to pnpm.

## Environment

- `ubuntu-latest` runner (Ubuntu 24.04)
- Node v24.14.1
- Introduced by PR #370 (switch to pnpm), partially addressed but not fixed by PR #393

SgtPooki:
Fixes #411.

## Problem

`upload-action` is broken on `master` and `v0.20.0`:

1. **`pnpm not found`** — `actions/setup-node@v6` was called with `cache: pnpm` *before* `pnpm/action-setup@v4`. setup-node invokes `pnpm store path --silent` to resolve its cache directory; pnpm wasn't on PATH yet.
2. **`Some specified paths were not resolved`** — `cache-dependency-path` pointed at `${repo-root}/pnpm-lock.yaml`, which lives at the action's checkout (`${{ github.action_path }}/..`) — outside `GITHUB_WORKSPACE`. `hashFiles()` only globs inside the workspace, so the cache key could never be built.

Introduced by #370 (switch to pnpm), partially addressed by #393.

## Fix

- Drop `cache: pnpm` and `cache-dependency-path` from `setup-node`. Those two inputs are the entire source of both failures.
- Replace with an explicit `actions/cache@v4` step run *after* `pnpm/action-setup@v4`, so pnpm is on PATH when we resolve the store path.
- Compute the lockfile hash with `node` (`crypto.createHash('sha256')`) instead of relying on `hashFiles()` or `sha256sum`. `node` reads any absolute path and is portable across linux/macOS/Windows runners.
- Harden the `packageManager` parser to fail loudly if the field is missing or not `pnpm@<version>`.

## Why this approach

- **Caching preserved.** `setup-node`'s built-in `cache: pnpm` is itself a thin wrapper over `actions/cache` that caches only the pnpm store (never `node_modules`). Doing it explicitly gives equivalent behavior — including save-on-job-success semantics — without setup-node's workspace assumptions.
- **Cache key namespaced and scoped tightly.** Key includes runner os + arch + Node version + pnpm version + lockfile sha256, prefixed with `filecoin-pin-upload-action-` to avoid collisions with the consumer repo's own pnpm caches. Node major can affect native deps and optional resolution, so it's part of the key.
- **No mutation of the consumer workspace.** Symlinking the lockfile into `GITHUB_WORKSPACE` to satisfy `hashFiles()` would be a leaky implementation detail for a public composite action; explicit cache avoids that.
- **Cross-platform.** Hashing via `node` works on every runner that already has Node — which we install one step earlier.

## Step order (final)

1. Get action path / repo root.
2. `actions/setup-node@v6` (Node only — no pnpm cache inputs).
3. Read + validate `packageManager` from `package.json`.
4. `pnpm/action-setup@v4`.
5. Get pnpm store path.
6. Compute lockfile sha256 via `node`.
7. `actions/cache@v4` for the pnpm store.
8. `pnpm install --frozen-lockfile`, `pnpm run build`.

## Test plan

- [x] Trigger consumer workflow at filecoin-project/filecoin-pin-website with `@fix/upload-action-pnpm-cache` and verify upload succeeds.
- [x] Verify cache hit on second run (lockfile unchanged).
- [x] Verify cache miss + new save on lockfile change.
- [x] Confirm the action still works when `packageManager` is set to a pnpm version with hash suffix (e.g. `pnpm@9.0.0+sha512.xxx`).

## Follow-up (out of scope, suggested for v0.21+)

Ship a pre-built `dist/` so consumers don't run `pnpm install && pnpm build` at action invocation time. Eliminates the entire class of pnpm-on-PATH issues for consumers and removes the cache concern altogether.

### DIFF
--- a/upload-action/action.yml
+++ b/upload-action/action.yml
@@ -159,7 +159,7 @@ runs:
         echo "version=$VERSION" >> $GITHUB_OUTPUT
 
     - name: Set up pnpm
-      uses: pnpm/action-setup@v4
+      uses: pnpm/action-setup@903f9c1a6ebcba6cf41d87230be49611ac97822e # v6.0.3
       with:
         run_install: false
         version: ${{ steps.pnpm-version.outputs.version }}
--- a/upload-action/action.yml
+++ b/upload-action/action.yml
@@ -139,6 +139,7 @@ runs:
       uses: actions/setup-node@v6
       with:
         node-version: '24.x'
+        package-manager-cache: false
 
     - name: Get pnpm version
       id: pnpm-version
@@ -147,7 +148,7 @@ runs:
       run: |
         VERSION=$(node -e "
           const pm = require('./package.json').packageManager;
-          if (typeof pm !== 'string') { console.error('package.json: packageManager field missing'); process.exit(1); }
+          if (typeof pm !== 'string') { console.error('package.json: packageManager field missing or not a string'); process.exit(1); }
           const match = pm.match(/^pnpm@(.+)$/);
           if (!match) { console.error('package.json: packageManager must be pnpm@<version>, got: ' + pm); process.exit(1); }
           process.stdout.write(match[1]);
@@ -163,9 +164,10 @@ runs:
     - name: Get pnpm store path
       id: pnpm-store
       shell: bash
+      working-directory: ${{ steps.repo-root.outputs.path }}
       run: |
         STORE_PATH=$(pnpm store path --silent)
-        echo "path=$STORE_PATH" >> $GITHUB_OUTPUT
+        printf 'path=%s\n' "$STORE_PATH" >> "$GITHUB_OUTPUT"
 
     - name: Compute pnpm-lock.yaml hash
       id: lockfile
--- a/upload-action/action.yml
+++ b/upload-action/action.yml
@@ -135,18 +135,23 @@ runs:
         echo "path=$REPO_ROOT" >> $GITHUB_OUTPUT
 
     - name: Set up Node.js
+      id: setup-node
       uses: actions/setup-node@v6
       with:
         node-version: '24.x'
-        cache: pnpm
-        cache-dependency-path: ${{ steps.repo-root.outputs.path }}/pnpm-lock.yaml
 
     - name: Get pnpm version
       id: pnpm-version
       shell: bash
       working-directory: ${{ steps.repo-root.outputs.path }}
       run: |
-        VERSION=$(node -pe "require('./package.json').packageManager.split('@')[1]")
+        VERSION=$(node -e "
+          const pm = require('./package.json').packageManager;
+          if (typeof pm !== 'string') { console.error('package.json: packageManager field missing'); process.exit(1); }
+          const match = pm.match(/^pnpm@(.+)$/);
+          if (!match) { console.error('package.json: packageManager must be pnpm@<version>, got: ' + pm); process.exit(1); }
+          process.stdout.write(match[1]);
+        ")
         echo "version=$VERSION" >> $GITHUB_OUTPUT
 
     - name: Set up pnpm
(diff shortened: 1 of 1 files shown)


SgtPooki:
@DavidBurela can you try filecoin-pin/upload-action against this branch and see if this addresses your issue? also check out some related PRs: 

* https://github.com/ipshipyard/ipfs-deploy-action/pull/50
* https://github.com/ipshipyard/ipfs-deploy-action/pull/49

upload-action/action.yml
rvagg: you could update this while  you're in here, we use v5 in the root

rvagg (approved):

juliangruber (approved):
Great that you fixed this 👏 

But I do really wonder what a solution without hacks (custom cache management) would look like

upload-action/action.yml
rvagg:
```suggestion
      uses: pnpm/action-setup@903f9c1a6ebcba6cf41d87230be49611ac97822e # v6.0.3
```

Issue #326, opened 2026-02-09
Expose real byte-level progress in uploadToSynapse callbacks


HavenCTO:
## Context / Problem

When calling `uploadToSynapse`, the progress reporting is limited to high-level milestone events (e.g., `onUploadComplete`, `onPieceAdded`). While the underlying `synapse-core` SDK provides real-time, byte-level upload progress via an `onProgress` callback, this information is not exposed or passed through by `filecoin-pin`. This prevents consumers from implementing accurate progress bars during the data transfer phase, as `bytesUploaded` remains unknown until the upload finishes.

## Current Implementation
https://github.com/filecoin-project/filecoin-pin/blob/f15bba4c4f616ac7e49cc7714ce1895e9dfe36e4/src/core/upload/synapse.ts#L93
`filecoin-pin/src/core/upload/synapse.ts` currently only maps lifecycle callbacks:

```typescript
const uploadCallbacks: UploadCallbacks = {
  onUploadComplete: (pieceCid) => { ... },
  onPieceAdded: (txHash) => { ... },
  onPieceConfirmed: (pieceIds) => { ... },
  // onProgress: ??? // ✗ NOT PASSED!
};
```

The SDK's `onProgress(bytesUploaded)` callback is completely ignored, leaving a gap in telemetry during the actual network transfer.

## The SDK Provides Real Progress
https://github.com/FilOzone/synapse-sdk/blob/4be67a838c5b602608b3ac33f9702c1343f87e09/packages/synapse-core/src/sp.ts#L523
**File:** `synapse-sdk/packages/synapse-core/src/sp.ts` (line ~520)
The lowest layer already tracks real progress:
```typescript
let bytesUploaded = 0
const trackingStream = new TransformStream<Uint8Array, Uint8Array>({
  transform(chunk, controller) {
    bytesUploaded += chunk.length
    if (options.onProgress) {
      options.onProgress(bytesUploaded) // Real bytes uploaded!
    }
    controller.enqueue(chunk)
  },
})
```

**File:** `synapse-sdk/packages/synapse-sdk/src/types.ts`
The `UploadCallbacks` interface supports this:
```typescript
export interface UploadCallbacks {
  /** Called periodically during upload with bytes uploaded so far */
  onProgress?: (bytesUploaded: number) => void
  onUploadComplete?: (pieceCid: PieceCID) => void
  // ...
}
```

## Requested Changes

[...]

Issue #325, opened 2026-02-09
Support ReadableStream in uploadToSynapse for memory-efficient uploads


HavenCTO:
## Context / Problem

We are integrating `filecoin-pin` (via the Synapse SDK) into a project where large CAR files (>1GiB) are pre-generated on disk. To utilize the Synapse upload capabilities, we must currently load the entire CAR file into memory before calling `uploadToSynapse`, as the function signature strictly accepts `Uint8Array`.

This results in peak memory usage of ~2.3 GiB for a 1 GiB file (original data + CAR format + memory buffers), which causes instability in resource-constrained environments. While the underlying `Synapse SDK` already supports streaming uploads via `ReadableStream<Uint8Array>` to minimize memory usage, the current `uploadToSynapse` implementation in `filecoin-pin` does not expose this capability, forcing a buffering strategy.

This limitation contradicts the goal of supporting files larger than 1GiB. As noted in [synapse-sdk#110](https://github.com/FilOzone/synapse-sdk/issues/110#issuecomment-3606422953):

IMO : Loading 1GB file into memory is not ideal and we need this change to fully support what @rjan90  said back in December.

To truly support larger files without crashing or requiring excessive RAM, `uploadToSynapse` needs to accept streams so that the file can be uploaded directly from disk rather than being loaded into memory first.

## Comparison with Issue #288

This request is distinct from **#288 (True Streaming CAR Uploads)**.

*   **Issue #288** addresses *creating* a CAR and uploading it simultaneously from source data, avoiding the need to buffer the CAR during creation. It is currently blocked by upstream Synapse API changes requiring a split `store`/`commit` flow to handle metadata timing.
*   **This Issue** addresses *uploading* an **already existing** CAR file from disk. The CAR is fully built and the root CID is known prior to the upload call. The goal is to pass this CAR to `uploadToSynapse` as a stream rather than a `Uint8Array` to prevent the memory overhead of loading the file into RAM.

Since the CAR is pre-generated, the metadata timing blockers in #288 do not apply here. This request can be fulfilled by updating the `uploadToSynapse` signature to accept streams.

## Requested Changes

Upstream code that needs to be aligned with - https://github.com/FilOzone/synapse-sdk/blob/4be67a838c5b602608b3ac33f9702c1343f87e09/packages/synapse-sdk/src/storage/context.ts#L907

[...]

Issue #288, opened 2025-12-10
True Streaming CAR Uploads


rvagg:
* Follow-up to #287 
* Related to #190

## Problem

Current CAR building buffers entire content before upload:

- **Node.js**: Writes CAR to temp file, then streams file to Synapse
- **Browser**: Accumulates CAR bytes in memory (up to 1GiB), then uploads

Two reasons for this:

1. **CAR header patching**: We write a placeholder root CID in the CAR header, build the DAG, then patch/rewrite the header with the actual root. This requires the CAR to be fully built before streaming. We can eliminate this by using `ZERO_CID` in the header instead (valid CAR, signals "no meaningful root"; technically an empty roots array is fine but there are some historical compatibility issues that make having one the safe option).

2. **Synapse metadata timing**: The `upload()` API requires IPFS root CID metadata at call time, but root CID is only known after DAG building completes. This requires API changes in Synapse.

## Blocker

Synapse SDK's `upload()` API requires metadata at call time. The IPFS root CID (`ipfsRootCID` metadata key) must be provided when calling `upload()`, but we don't know it until DAG building completes.

**Blocked by**: FilOzone/synapse-sdk#494 - Split `upload()` into `store()` + `commit()`

## Solution

Once Synapse provides separate `store()` and `commit()` operations:

1. **CAR builder changes**:
   - Use `CarWriter.create([ZERO_CID])` (where `ZERO_CID = bafkqaaa`)
   - Return `{ carStream: ReadableStream, rootCidPromise: Promise<CID> }` instead of buffered bytes/file
   - Remove `updateRootCidInCar()` post-processing

2. **Upload flow changes**:
   - Call `store(carStream)` - streams bytes directly as DAG builds
   - Await `rootCidPromise` - resolves when UnixFS completes
   - Call `commit(pieceCid, { ics: rootCid })` - registers with metadata

3. **Files to modify**:
   - `src/core/car/car-*-backend.ts` - expose `out` stream, use ZERO_CID
   - `src/core/unixfs/car-builder.ts` - return streaming result
   - `src/core/unixfs/browser-car-builder.ts` - return streaming result
   - `src/core/upload/synapse.ts` - use new store/commit API
   - `src/add/add.ts`, `src/import/import.ts`, `src/filecoin-pin-store.ts` - consume streaming API

rvagg:
[...]

PR #428, opened 2026-04-30
feat: support stream uploads to Synapse

shrirajpawar4:
## Summary

  addresses part of #325.

  Adds `ReadableStream<Uint8Array>` support to the shared Synapse upload helpers so callers with pre-generated CAR files can stream upload data without first loading the whole CAR into memory.

  ## Changes

  - Add `SynapseUploadData = Uint8Array | ReadableStream<Uint8Array>`
  - Update `uploadToSynapse()` and `executeUpload()` to accept `SynapseUploadData`
  - Pass upload data directly through to `synapse.storage.upload()` without buffering
  - Export the new upload data type from the public upload/type surface
  - Add unit coverage proving stream objects are forwarded unchanged
  - Update the CAR upload example to show streaming from disk with `Readable.toWeb(createReadStream(...))`

  ## Notes

  This intentionally does not implement true CAR creation streaming from #288, and it does not add byte-level progress callbacks from #326. This PR only covers uploading an already-created CAR as a stream. Will create separet PR for #326

  ## Testing

  ```bash
  pnpm run typecheck
  pnpm exec vitest run --project unit src/test/unit/synapse-service.test.ts src/test/unit/upload-validation.test.ts
  pnpm test

  Full pnpm test passed locally:

  - 37 test files passed
  - 372 tests passed
  - 11 skipped

### DIFF
--- a/src/core/payments/README.md
+++ b/src/core/payments/README.md
@@ -69,10 +69,12 @@ const synapseService = await setupSynapse(config, logger, {
 ### Upload CAR File
 
 ```typescript
+import { createReadStream } from 'node:fs'
+import { Readable } from 'node:stream'
 import { uploadToSynapse } from 'filecoin-pin/core/upload'
 import { CID } from 'multiformats/cid'
 
-const carData = await fs.readFile('path/to/file.car')
+const carData = Readable.toWeb(createReadStream('path/to/file.car')) as ReadableStream<Uint8Array>
 const rootCid = CID.parse('bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi')
 
 const result = await uploadToSynapse(
@@ -81,16 +83,15 @@ const result = await uploadToSynapse(
   rootCid,
   logger,
   {
-    callbacks: {
-      onUploadComplete: (pieceCid) => {
-        console.log(`Upload complete: ${pieceCid}`)
+    onProgress: (event) => {
+      if (event.type === 'onStored') {
+        console.log(`Stored on provider ${event.data.providerId}: ${event.data.pieceCid}`)
       }
     }
   }
 )
 
 console.log(`Piece CID: ${result.pieceCid}`)
-console.log(`Download URL: ${result.providerInfo?.downloadURL}`)
 ```
 
 ### Setup Payments
--- a/src/core/payments/README.md
+++ b/src/core/payments/README.md
@@ -92,6 +92,7 @@ const result = await uploadToSynapse(
 )
 
 console.log(`Piece CID: ${result.pieceCid}`)
+console.log(`Retrieval URL: ${result.copies[0]?.retrievalUrl}`)
 ```
 
 ### Setup Payments
--- a/src/core/payments/README.md
+++ b/src/core/payments/README.md
@@ -74,7 +74,7 @@ import { Readable } from 'node:stream'
 import { uploadToSynapse } from 'filecoin-pin/core/upload'
 import { CID } from 'multiformats/cid'
 
-const carData = Readable.toWeb(createReadStream('path/to/file.car')) as ReadableStream<Uint8Array>
+const carData = Readable.toWeb(createReadStream('path/to/file.car'))
 const rootCid = CID.parse('bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi')
 
 const result = await uploadToSynapse(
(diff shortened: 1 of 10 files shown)


src/core/payments/README.md
```diff
@@ -69,10 +69,12 @@ const synapseService = await setupSynapse(config, logger, {
 ### Upload CAR File
 
 ```typescript
+import { createReadStream } from 'node:fs'
+import { Readable } from 'node:stream'
 import { uploadToSynapse } from 'filecoin-pin/core/upload'
 import { CID } from 'multiformats/cid'
 
-const carData = await fs.readFile('path/to/file.car')
+const carData = Readable.toWeb(createReadStream('path/to/file.car')) as ReadableStream<Uint8Array>
```
rvagg:
```suggestion
const carData = Readable.toWeb(createReadStream('path/to/file.car'))
```

I know this block says typescript, but it it's actually runnable JS as is and since it's just an example we may as well leave this alone (and let the typescript user figure it out); although I think we might be able to inline type it to make it work for both (`/** @type {ReadableStream<Uint8Array>} */` I _think_ might work?).

rvagg:
This looks pretty good, tightly scoped and solves the specific issue.

There's one additional thing you could do with this @shrirajpawar4, you look in `src/import/import.ts` you'll see we `await readFile(options.filePath)` for `filecoin-pin import`; we should be streaming that out. And I think the change is pretty simple there, we `stat()` separately and we even read the root CID from the CAR separately (using a stream).

rvagg:
@shrirajpawar4 have a look at https://github.com/filecoin-project/filecoin-pin/pull/287 which was a draft of a previous attempt and it does a few more things. I think the code might have moved on from that PR but it had a bit more, including streaming the temp car for `add` operations which you can do as well, and there's some minor changes to the docs in there but as per my comment we can't over-promise because we still cache things on disk and `add` is still bad in the browser because it makes a temp CAR in memory before streaming it out. So don't overpromise like the original text in there.
Also watch mocks when `readFile` is replaced with `createReadStream`, I think we need to add `stat` in there.

But, if you'd prefer to land this one first and do a follow-up with the second batch we can do that; this is a decent start but it shouldn't be hard to go a bit further.

shrirajpawar4:
Thanks @rvagg . I applied the README suggestion and updated `filecoin-pin import` to stream the CAR file from disk instead of reading it fully into memory before upload.

I kept the broader `add` / server / action streaming work out of this PR so this can stay focused on #325; I can handle that as a follow up after this lands.

src/core/payments/README.md
```diff
@@ -81,16 +83,15 @@ const result = await uploadToSynapse(
...
+    onProgress: (event) => {
+      if (event.type === 'onStored') {
+        console.log(`Stored on provider ${event.data.providerId}: ${event.data.pieceCid}`)
       }
     }
   }
 )
 
 console.log(`Piece CID: ${result.pieceCid}`)
-console.log(`Download URL: ${result.providerInfo?.downloadURL}`)
```
SgtPooki, line 93: recommend to not remove this

src/import/import.ts
```diff
@@ -208,10 +209,10 @@ export async function runCarImport(options: ImportOptions): Promise<ImportResult
       await validatePaymentSetup(synapse, fileStat.size, spinner)
     }
 
-    // Read CAR file and upload to Synapse
+    // Stream CAR file and upload to Synapse
```
SgtPooki:
```suggestion
    // Stream CAR file to Synapse
```

src/test/mocks/synapse-mocks.ts
```diff
@@ -47,7 +47,7 @@ export class MockStorageContext extends EventEmitter {
   public readonly provider = mockPDPProvider
   public readonly serviceProvider = mockPDPProvider.serviceProvider
 
-  async upload(_data: ArrayBuffer | Uint8Array, options?: any): Promise<any> {
+  async upload(_data: Uint8Array | ReadableStream<Uint8Array>, options?: any): Promise<any> {
```
SgtPooki: for consistency
```suggestion
  async upload(_data: SynapseUploadData, options?: any): Promise<any> {
```

SgtPooki (changes requested):
short and sweet, nice PR. Just a few things to touch up and i'm good with it.

shrirajpawar4:
@SgtPooki made changes ser.

HavenCTO:
**Approve**   this unblocks the streaming contract we need, thanks for the implementation!

rvagg (approved):
lgtm but we'll just have to make sure t he linked issue stays open so we can track the other part of t his

shrirajpawar4:
@rvagg yes works, will close out the issue in next PR.

SgtPooki (approved):

PR #433, opened 2026-05-01
feat(cli): add --min-runway-days and --max-balance to --auto-fund

### LINKED ISSUE #427: Add --min-runway-days and --max-balance to --auto-fund for upload-action parity
Coming out of https://github.com/ipshipyard/ipfs-deploy-action/pull/49 and noting that in the current [action](https://github.com/filecoin-project/filecoin-pin/blob/master/upload-action/action.yml) has two options that don't map directly on to the CLI and which need to be run as a manual funding plan: `minStorageDays` and `filecoinPayBalanceLimit`.

`--auto-fund` currently defaults to `MIN_RUNWAY_DAYS`=`30` and doesn't have a built-in limiter.

Proposal:

1. Add optional `--min-runway-days <n>` which defines the minimum runway to maintain (not gross, actual runway, see https://github.com/FilOzone/synapse-sdk/pull/760); maps to `minStorageDays` in the action.
2. Add optional `--max-balance <usdfc>` caps the Filecoin Pay balance after deposit, not the deposit amount itself, `executeTopUp` already implements clamp-or-skip behaviour against this limit; we just don't expose it to `--auto-fund`; maps to `filecoinPayBalanceLimit` in the action.

Both would _require_ `--auto-fund` to be given on the CLI (let's not implicitly invoke auto funding unless they ask for it).

They're fairly simple additions but would allow the same automation the GitHub action gives us with a CLI:

```sh
PRIVATE_KEY=0x... \
  npx filecoin-pin import build.car \
    --mainnet \
    --auto-fund \
    --min-runway-days 365 \
    --max-balance 5.0
````

There might be some interactions with https://github.com/filecoin-project/filecoin-pin/pull/412 to clean up later when we inherit https://github.com/filecoin-project/filecoin-pin/pull/412 .. we'll see.

### PR DESCRIPTION
**Pull request by rvagg**
Brings add/import auto-fund to parity with upload-action's minStorageDays and filecoinPayBalanceLimit. Both modifiers strictly require --auto-fund.
Closes: #427

I've tried this out with the various possible permutations on a calibnet wallet and it seems to work really well.

Unblocks https://github.com/ipshipyard/ipfs-deploy-action/pull/49

### DIFF
--- a/src/core/payments/top-up.ts
+++ b/src/core/payments/top-up.ts
@@ -5,6 +5,48 @@ import { depositUSDFC, getPaymentStatus } from './index.js'
 import type { TopUpResult } from './types.js'
 
 /**
+ * Result of clamping a requested deposit against a balance limit.
+ *
+ * - `passthrough`: limit is undefined or unreached; deposit equals requested
+ * - `already-at-limit`: current balance already meets/exceeds limit; deposit is 0n
+ * - `clamped`: deposit was reduced to the largest amount that doesn't exceed limit
+ */
+export interface ClampDepositResult {
+  deposit: bigint
+  reason: 'passthrough' | 'already-at-limit' | 'clamped'
+  message?: string
+}
+
+/**
+ * Pure helper: clamp a requested deposit so the resulting balance does not exceed `limit`.
+ *
+ * @param currentBalance - Current Filecoin Pay balance
+ * @param requested - Requested deposit amount (must be > 0n for clamping to apply)
+ * @param limit - Maximum allowed post-deposit balance; undefined means no limit
+ */
+export function clampDepositToLimit(currentBalance: bigint, requested: bigint, limit?: bigint): ClampDepositResult {
+  if (limit == null || limit < 0n || requested <= 0n) {
+    return { deposit: requested, reason: 'passthrough' }
+  }
+  if (currentBalance >= limit) {
+    return {
+      deposit: 0n,
+      reason: 'already-at-limit',
+      message: `Current balance (${formatUSDFC(currentBalance)}) already equals or exceeds the configured balance limit (${formatUSDFC(limit)}). No additional deposits will be made.`,
+    }
+  }
+  if (currentBalance + requested > limit) {
+    const maxAllowed = limit - currentBalance
+    return {
+      deposit: maxAllowed,
+      reason: 'clamped',
+      message: `Required top-up (${formatUSDFC(requested)}) would exceed the configured balance limit (${formatUSDFC(limit)}). Reducing to ${formatUSDFC(maxAllowed)}.`,
+    }
+  }
+  return { deposit: requested, reason: 'passthrough' }
+}
+
+/**
  * Execute a top-up operation with balance limit checking
  *
  * This function handles the complete top-up process including:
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: src/core/payments/top-up.ts [hunk 1]
```diff
@@ -4,6 +4,48 @@ import { formatUSDFC } from '../utils/index.js'
...
+
+/**
+ * Pure helper: clamp a requested deposit so the resulting balance does not exceed `limit`.
+ *
+ * @param currentBalance - Current Filecoin Pay balance
+ * @param requested - Requested deposit amount (must be > 0n for clamping to apply)
+ * @param limit - Maximum allowed post-deposit balance; undefined means no limit
+ */
+export function clampDepositToLimit(currentBalance: bigint, requested: bigint, limit?: bigint): ClampDepositResult {
+  if (limit == null || limit < 0n || requested <= 0n) {
```
SgtPooki, line 28: is there a chance folks could result in limit < 0 here and that be intentional? if limit is negative, i'm thinking they may want to withdraw funds, and we shouldn't just passthrough, but i might be overthinking without following all the flows through.

src/import/types.ts
```diff
@@ -3,8 +3,12 @@ import type { CLIAuthOptions } from '../utils/cli-auth.js'
 
 export interface ImportOptions extends CLIAuthOptions {
   filePath: string
-  /** Auto-fund: automatically ensure minimum 30 days of runway */
+  /** Auto-fund: automatically ensure minimum runway (default 30 days) before upload */
   autoFund?: boolean
+  /** Override the minimum runway (in days) targeted by auto-fund */
+  minRunwayDays?: number
+  /** Cap on Filecoin Pay balance after deposit, in USDFC base units */
+  maxBalance?: bigint
```
SgtPooki: it would be nice to extend a `CLIAutoFundOptions` as well so we don't have to port every change in every place

src/utils/cli-options.ts
```diff
@@ -108,3 +111,86 @@ export function addUploadOptions(command: Command): Command {
...
+ * Add auto-fund options to a command.
+ * Used by `add` and `import` commands. Modifiers require `--auto-fund`; validate
+ * post-parse with {@link validateAndNormalizeAutoFundOptions}.
+ */
+export function addAutoFundOptions(command: Command): Command {
+  return command
+    .option(
+      '--auto-fund',
+      `Automatically deposit USDFC before upload to maintain runway (default: ${MIN_RUNWAY_DAYS} days)`
+    )
```
SgtPooki: you can use commanders `.conflicts` here with `view-address` to get the validation for free. It would be nice if they had the inverse too: `requiresOtherOption(` or similar for the options below.. but sadly it does not.

src/utils/cli-options.ts
```diff
@@ -108,3 +111,86 @@ export function addUploadOptions(command: Command): Command {
...
+
+  if (raw.minRunwayDays !== undefined && !autoFund) {
+    throw new Error('--min-runway-days requires --auto-fund')
+  }
+  if (raw.maxBalance !== undefined && !autoFund) {
+    throw new Error('--max-balance requires --auto-fund')
+  }
+  if (autoFund && raw.viewAddress !== undefined) {
+    throw new Error('--auto-fund cannot be used with --view-address (read-only mode cannot sign deposits)')
+  }
```
SgtPooki: use `.conflicts`

SgtPooki (approved):
overall lgtm, a few comments

src/core/payments/top-up.ts
rvagg: This is just a refactor of what's already there, the existing gates on `balanceLimit != null && balanceLimit >= 0n`. We're validating the limit in `validateAndNormalizeAutoFundOptions` so this is not expected to be hit by the common path it's just defensive; you'd only hit this if you were to call this yourself. I think leave for now but if someone wants to use this as a way to withdraw then we could adjust accordingly

PR #439, opened 2026-05-01
feat(ipni): surface per-CID outcomes for diagnostic visibility

### LINKED ISSUE #417: waitForIpniProviderResults: surface per-CID outcomes for diagnostic visibility
## Problem

`waitForIpniProviderResults` returns `Promise<boolean>` and on failure throws a single `Error` whose message references one CID (the last one checked). Consumers cannot tell, after the fact, which CIDs in `[rootCid, ...childBlocks]` were verified vs. timed out vs. missing-expected-provider vs. fetch-errored.

This matters most on the abort/timeout path: when an outer `AbortSignal.timeout(...)` fires mid-walk, the in-flight `fetch` throws `AbortError` and we lose which CID was current and which CIDs had already been verified successfully.

Concrete example from the dealbot consumer (FilOzone/dealbot):

- Deals are uploaded with rootCID + ~12 childBlocks. Outer timeout 60s, delayMs 2500ms, so each CID gets ~25 attempts.
- 7 of 8 SPs verify cleanly at p50 ~2.4s. SP6 (Hangzhou CN) verifies at p50 ~31s with 39% slow_verification rate; ~22 deals/24h time out at the 60s budget.
- All timeouts surface as a single generic message. We currently cannot distinguish \"rootCID never indexed\" from \"rootCID indexed but childBlock N never indexed.\"
- Investigation: https://github.com/FilOzone/dealbot/issues/473#issuecomment-4336343553

## Why this is currently low priority

1. In observed dealbot data, all timeouts are wall-clock SLO breaches, not selective \"missing expected provider\" failures. Per-CID granularity would not have shifted root-cause attribution (path quality between filecoinpin.contact / Reston VA and SP6 / Hangzhou CN, AS4134 ChinaNet, ICMP packet loss 24% from a peer test vantage).
2. At dealbot's current SLO budget (60s with 2.5s polls and 12 CIDs), only the first 4-5 CIDs get any attempts before the outer abort fires anyway — so child-CID granularity is not measurable end-to-end without also rebudgeting.
3. PR https://github.com/FilOzone/dealbot/pull/490 + https://github.com/FilOzone/dealbot/pull/491 close the immediate visibility gap that prompted dealbot#473 by carrying the existing single failureReason via `Error.cause`.

## Proposal (when prioritized)

Change return type from `Promise<boolean>` to something like:

\`\`\`ts
type IpniValidationOutcome = {
  verified: Array<{ cid: CID; matchedAt: number; attempts: number }>;
  failed: Array<{
    cid: CID;
    reason:
      | { type: 'timeout'; attempts: number }
      | { type: 'missingProviders'; missingServiceUrls: string[]; actualMultiaddrs: string[] }
      | { type: 'fetch'; message: string }
      | { type: 'parse'; message: string }
      | { type: 'http'; status: number; statusText?: string };
  }>;
  ipniIndexerUrl: string;
};
\`\`\`

Existing single-error throw can be preserved as a thin wrapper for callers that prefer the boolean shape, or replaced with always returning the outcome and letting callers decide whether to throw.

## Acceptance criteria

- [ ] `waitForIpniProviderResults` returns or exposes per-CID verified/failed lists.
- [ ] Abort during `setTimeout(delayMs)` (currently signal-unaware at https://github.com/filecoin-project/filecoin-pin/blob/main/src/core/utils/validate-ipni-advertisement.ts#L387) is preserved or fixed in the same change. Optional related fix: make the inter-attempt sleep signal-aware so abort overshoot is bounded by the in-flight fetch rather than by `delayMs`.
- [ ] Existing consumers that only need pass/fail can keep that behavior via a wrapper or a boolean accessor on the new return type.
- [ ] Tests cover: timeout-mid-walk, partial verification (some CIDs verified, some failed), all-CIDs-missing, single-CID success.

## Consumer signal

Open this when:
- A real dealbot incident appears where timeout reason \"some childBlock never indexed\" needs to be distinguished from \"rootCID never indexed.\"
- Or filecoin-pin gains another consumer that needs per-CID outcomes for its own product semantics.

SgtPooki:
Closes #417.

### What changed

`waitForIpniProviderResults` now exposes a structured `IpniValidationOutcome` (`verified[]` / `failed[]` with a `reason` discriminator: `timeout` / `missingProviders` / `fetch` / `parse` / `http` / `aborted` / `notAttempted`). Existing callers keep the `Promise<boolean>` shape — on failure the outcome is attached to `Error.cause` so the legacy thrown-error path stays intact and tests/messages are unchanged.

New surface:
- `waitForIpniProviderResultsDetailed(cid, opts)` — non-throwing variant that always returns the outcome.
- `ipniProviderResults.outcome` progress event — emitted on every run with the full per-CID detail.
- Inter-attempt sleep is now signal-aware (`abortableDelay`), so an outer `AbortSignal.timeout(...)` interrupts the wait instead of having to ride out `delayMs` until the next fetch.

### How to verify

```
pnpm test
```

New tests cover: single-CID success outcome, partial verification (some CIDs verified before failure), all-CIDs-missing, timeout-mid-walk via abort during sleep, signal-aware sleep (no extra fetch after abort), and `Error.cause` carrying the outcome.

### Notes

- Public type exports added in `src/index-types.ts`: `IpniValidationOutcome`, `IpniFailureReason`, `IpniVerifiedEntry`, `IpniFailedEntry`.
- Walk still stops at the first per-CID failure to preserve existing throw semantics; later CIDs land in `failed` with `reason.type === 'notAttempted'` so consumers can distinguish "never checked" from "checked and failed".

### DIFF
--- a/src/core/utils/validate-ipni-advertisement.ts
+++ b/src/core/utils/validate-ipni-advertisement.ts
@@ -45,8 +45,14 @@ interface ProviderResult {
  * for a given CID. Earlier intermediate failures (e.g. transient fetch errors)
  * may have been retried away and are not reported here.
  */
+/**
+ * Per-CID terminal failure classification.
+ *
+ * Whichever observation was current on the *last* attempt becomes the
+ * terminal reason — there is no separate "timeout" type, since attempts
+ * exhaustion always carries one of the per-attempt observations.
+ */
 export type IpniFailureReason =
-  | { type: 'timeout'; attempts: number; lastObservation?: string }
   | {
       type: 'missingProviders'
       attempts: number
@@ -488,9 +494,14 @@ async function validateOneCid(cid: CID, config: ValidateOneCidConfig): Promise<C
       continue
     }
 
+    // Every loop iteration that does not return success sets `lastReason`,
+    // so this fallback should be unreachable. Surface a structured fallback
+    // rather than throwing to keep the function total.
     const finalReason: IpniFailureReason = lastReason ?? {
-      type: 'timeout',
+      type: 'missingProviders',
       attempts: retryCount,
+      missingServiceUrls: [],
+      actualMultiaddrs: [],
     }
     return { verified: false, reason: finalReason, attempts: retryCount }
   }
@@ -543,10 +554,7 @@ function buildOutcomeError(
   }
 
   if (firstFailure.reason.type === 'aborted') {
-    const error = new Error('Check IPNI announce aborted', {
-      cause: { signal: options?.signal, outcome },
-    })
-    return error
+    return new Error('Check IPNI announce aborted', { cause: outcome })
   }
 
   const expectedProviders = options?.expectedProviders?.filter((p) => p != null) ?? []
@@ -587,8 +595,6 @@ function formatLastObservation(reason: IpniFailureReason): string | undefined {
       return `Failed to parse IPNI response body: ${reason.message}`
     case 'http':
       return `IPNI indexer request failed with status ${reason.status}`
-    case 'timeout':
-      return reason.lastObservation
     case 'aborted':
     case 'notAttempted':
       return undefined
--- a/src/core/utils/validate-ipni-advertisement.ts
+++ b/src/core/utils/validate-ipni-advertisement.ts
@@ -38,6 +38,51 @@ interface ProviderResult {
   }
 }
 
+/**
+ * Per-CID failure classification produced by {@link waitForIpniProviderResults}.
+ *
+ * The `type` discriminator tells consumers what happened on the *last* attempt
+ * for a given CID. Earlier intermediate failures (e.g. transient fetch errors)
+ * may have been retried away and are not reported here.
+ */
+export type IpniFailureReason =
+  | { type: 'timeout'; attempts: number; lastObservation?: string }
+  | {
+      type: 'missingProviders'
+      attempts: number
+      missingServiceUrls: string[]
+      actualMultiaddrs: string[]
+    }
+  | { type: 'fetch'; attempts: number; message: string }
+  | { type: 'parse'; attempts: number; message: string }
+  | { type: 'http'; attempts: number; status: number; statusText?: string }
+  | { type: 'aborted'; attempts: number }
+  | { type: 'notAttempted' }
+
+export interface IpniVerifiedEntry {
+  cid: CID
+  attempts: number
+}
+
+export interface IpniFailedEntry {
+  cid: CID
+  reason: IpniFailureReason
+}
+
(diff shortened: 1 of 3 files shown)


src/core/utils/validate-ipni-advertisement.ts
```diff
@@ -231,181 +333,281 @@ async function waitForIpniProviderResultsForCid(
...
-          event: 'check-ipni-announce',
-          ipfsRootCid: cid.toString(),
+  let retryCount = 0
+  let lastReason: IpniFailureReason | null = null
+  let lastActualMultiaddrs: Set<string> = new Set()
+  let lastActualUris: Set<string> = new Set()
+
+  while (true) {
+    if (options?.signal?.aborted) {
+      return { verified: false, reason: { type: 'aborted', attempts: retryCount }, attempts: retryCount }
```
rvagg: there is a `signal.reason` here you could use too .. probably not so helpful for CLI but .. loss of information; up to you

src/core/utils/validate-ipni-advertisement.ts
```diff
@@ -115,25 +166,69 @@ export interface WaitForIpniProviderResultsOptions {
...
- * @param ipfsRootCid - The IPFS root CID to check
- * @param options - Options for the check
- * @returns True if the IPNI announce succeeded, false otherwise
+ * Should not be called until you receive confirmation from the SP that the
+ * piece has been parked (e.g. `onPieceAdded` in `synapse.storage.upload`).
+ *
+ * @returns `true` when every CID verified.
+ * @throws Error with `cause: IpniValidationOutcome` on any failure.
  */
 export async function waitForIpniProviderResults(
```
rvagg, line 197: The event emitting pattern split is weird here between `waitForIpniProviderResults` and `waitForIpniProviderResultsDetailed`. The latter advertises itself as the detailed variant but it doesn't emit `outcome`, `complete` or `failed` events, but it takes the same `options` with `onProgress` and does some `onProgress` work itself for retries.
Do we really need this split? Or could we divide it more cleanly? `waitForIpniProviderResults` is left looking like a backward-compat wrapper that does some basic summing, maybe fine, but also .. maybe just break the API and don't worry too much about it? Or introduce some nice utilities to help reframe this?

SgtPooki:
~~after seeing #689 i'm going to go ahead and close this for now.. it seems like we need to pivot how we handle this anyway.~~

we still need per CID visibility.

SgtPooki:
Closing. #689 changed the indexing flow enough that per-CID visibility would need a fresh, smaller PR. Leaving #417 open until someone else needs it.

PR #442, opened 2026-05-05
feat(cli)!: default network selection to mainnet

rjan90:
BREAKING CHANGE: CLI commands now use Filecoin Mainnet when --network/NETWORK is omitted. Use --network calibration or NETWORK=calibration for Calibration testnet.

## Summary

- Change the shared CLI network default from Calibration to Mainnet.
- Centralize default network normalization with `DEFAULT_NETWORK = 'mainnet'`.
- Ensure CLI auth, server config, and interactive payment setup pass the matching Synapse chain along with the resolved RPC URL.
- Update README and AGENTS docs to describe Mainnet as the default.
- Update unit tests to cover the new default and custom RPC behavior.

## Testing

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run test:unit`
- `pnpm exec tsx src/cli.ts --no-update-check add --help`
- `pnpm exec tsx src/cli.ts --no-update-check server --help`

## Notes

Users who want Calibration testnet now need to pass `--network calibration` or set `NETWORK=calibration`.

PR #445, opened 2026-05-05
feat(cli)!: default network to mainnet

SgtPooki:
### What changed

The CLI default network flips from Calibration to Mainnet. Commander option default in `addNetworkOptions()` is now `'mainnet'`, the `--mainnet` shorthand is dropped (redundant), and `getRpcUrl()` falls back to mainnet when no `--rpc-url`, `--network`, `RPC_URL`, or `NETWORK` is provided. Docs and unit tests updated to match. README Status section reframed to signal mainnet readiness.

Minimal alternative to #442. Only the default flip, no refactor of `get-rpc-url.ts` and no `chain` plumbing through `Config`. Server-side chain plumbing is split into #446.

### How to verify

- `pnpm run typecheck`
- `pnpm run lint:fix`
- `pnpm run test:unit`
- `pnpm exec tsx src/cli.ts --no-update-check add --help` shows `(default: "mainnet")`

### Notes / risks

Breaking change for anyone relying on the implicit Calibration default. Migration: pass `--network calibration` or set `NETWORK=calibration`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,7 @@
 
 [![NPM](https://nodei.co/npm/filecoin-pin.svg?style=flat&data=n,v)](https://nodei.co/npm/filecoin-pin/)
 
-**Store IPFS content on Filecoin's decentralized storage network with verifiable persistence.**
-
-## Status
-
-**Ready for persistent, verifiable data on Filecoin Mainnet.**
-
-Register for updates and a later 2025 Q4 GA announcement at [filecoin.cloud](https://filecoin.cloud/).
+**Store your files on IPFS with Filecoin Pin, backed by [Filecoin’s decentralized storage network](https://filecoin.cloud/), with storage managed by your wallet through onchain payments.**
 
 ## What is Filecoin Pin?
 
@@ -111,7 +105,7 @@ Filecoin Pin's CLI collects telemetry.  A few things:
 * Telemetry always [has a way to be disabled](#how-to-disable-telemetry).
 * We don't collect Personal identifiable information (PII).
 * With our [end user affordance](#affordances) we expect to make telemetry on by default, requiring a consumer/user to opt out.  We are defaulting as "enabled" to help make sure we have a good pulse on the user experience and can address issues correctly.
-* In this [pre-v1 season](https://github.com/filecoin-project/filecoin-pin/issues/187), we are particularly focused on helping maintainers validate functionality and iron out problems throughout the whole Filecoin Onchain Cloud stack that `filecoin-pin` relies on. 
+* In this [pre-v1 season](https://github.com/filecoin-project/filecoin-pin/issues/187), we are particularly focused on helping maintainers validate functionality and iron out problems throughout the whole Filecoin Onchain Cloud stack that `filecoin-pin` relies on.
 
 ### How to disable CLI telemetry
 
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ filecoin-pin add myfile.txt
 # 3. Verify storage with cryptographic proofs
 filecoin-pin data-set <dataset-id>
 
-# To use Calibration testnet instead:
+# To use Calibration testnet (not persistent) instead:
 filecoin-pin add myfile.txt --network calibration
 ```
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 ## Status
 
-**⚠️ Not ready yet for production** - Filecoin Pin supports both Filecoin Mainnet and Calibration testnet. The CLI defaults to Calibration testnet for safety, but you can use `--network mainnet` to connect to Mainnet, but filecoin-pin is not ready for production use yet.
+**Ready for persistent, verifiable data on Filecoin Mainnet.**
 
 Register for updates and a later 2025 Q4 GA announcement at [filecoin.cloud](https://filecoin.cloud/).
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -6,7 +6,7 @@
 
 ## Status
 
-**⚠️ Not ready yet for production** - Filecoin Pin supports both Filecoin Mainnet and Calibration testnet. The CLI defaults to Calibration testnet for safety, but you can use `--network mainnet` to connect to Mainnet, but filecoin-pin is not ready for production use yet.
+**Ready for persistent, verifiable data on Filecoin Mainnet.**
```
SgtPooki: @BigLep @jennijuju @rjan90 open to different wording here.

SgtPooki (commented):
self review

src/common/get-rpc-url.ts
```diff
@@ -107,9 +107,9 @@ export function getRpcUrl(options: CLIAuthOptions): string {
     return wsUrl
   }
 
-  const defaultUrl = calibration.rpcUrls.default.webSocket?.[0] ?? calibration.rpcUrls.default.http[0]
+  const defaultUrl = mainnet.rpcUrls.default.webSocket?.[0] ?? mainnet.rpcUrls.default.http[0]
   if (!defaultUrl) {
-    throw new Error('No RPC URL available for calibration network')
+    throw new Error('No RPC URL available for mainnet network')
   }
   return defaultUrl
```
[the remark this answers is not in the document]
SgtPooki: Good catch. The CLI path is fixed in this PR. Commander default is now `mainnet`, so `parseCLIAuth()` threads the chain. The server path (`createConfig()` then `buildSynapseConfig()`) still skips chain. That is a pre-existing bug independent of the default flip, fixed separately in #446.

README.md
jennijuju:
```suggestion
**Store your files on IPFS with Filecoin Pin, backed by Filecoin’s decentralized storage network, with storage managed by your wallet through onchain payments.**
```

README.md
```diff
@@ -148,14 +148,14 @@ npm install -g filecoin-pin
...
-# 2. Upload a file to Filecoin (defaults to Calibration testnet)
+# 2. Upload a file to Filecoin (defaults to Mainnet)
 filecoin-pin add myfile.txt
 
 # 3. Verify storage with cryptographic proofs
 filecoin-pin data-set <dataset-id>
 
-# To use Mainnet instead:
-filecoin-pin add myfile.txt --network mainnet
+# To use Calibration testnet instead:
```
jennijuju: ### 
```suggestion
# To use Calibration testnet (not persistent) instead:
```

rjan90:
I think maybe there are is a places where this minimal change can leave RPC URL and chain metadata out of sync if I read the code correctly:

1. Interactive `payments setup`:

https://github.com/filecoin-project/filecoin-pin/blob/c9575973c305fa9a869edb281fa39c7db736f574/src/payments/interactive.ts#L80-L92

`initializeSynapse()` then falls back to Calibration when `chain` is omitted:

https://github.com/filecoin-project/filecoin-pin/blob/c9575973c305fa9a869edb281fa39c7db736f574/src/core/synapse/index.ts#L177-L179

SgtPooki:
Good catch. Fixed in f10d847. `runInteractiveSetup()` now goes through `parseCLIAuth()`, which threads the chain alongside the RPC URL, so `initializeSynapse()` no longer falls back to calibration. Same approach as the rest of the CLI path (`add`, `import`, etc.).

The pinning server has the same class of bug and is fixed separately in #446.

README.md
SgtPooki: @jennijuju I feel like this wording is good, but slightly out of place under a "status" heading. should we just remove the status heading and use this text? or maybe move this up higher than the status heading?

SgtPooki:
@rjan90 added a new test that asserts all commands have a --network option and default to mainnet.. will be easy to prevent new commands from getting off-track with #454

rjan90:
Potentially one more where it could drop the selected/default network (but would need your review on it as well @SgtPooki). Whenever `--rpc-url` or `RPC_URL` is present here: https://github.com/filecoin-project/filecoin-pin/blob/6ddd27c78d7f083015c47a1f6b2204e75dec1b77/src/commands/server.ts#L35-L39

`createConfig()` sees `RPC_URL` but no `NETWORK`, so the chain can remain undefined here: https://github.com/filecoin-project/filecoin-pin/blob/7557466944a4fd811d7d19757bd141181c26e42c/src/config.ts#L54-L59

And `initializeSynapse()` falls back to Calibration:

https://github.com/filecoin-project/filecoin-pin/blob/7557466944a4fd811d7d19757bd141181c26e42c/src/core/synapse/index.ts#L177-L179

rjan90 (approved):

README.md
rvagg: yeah @SgtPooki I agree, move that text up to replace the line above status; you have a line just below here that talks about GA and 2025 Q4 which needs updating. I think we could probably just ditch the status section entirely so we don't have to come back to it. Put the filecoin.cloud link into new line above here.

Could just replace lines 5-11 with:

```suggestion
**Store your files on IPFS with Filecoin Pin, backed by [Filecoin’s decentralized storage network](https://filecoin.cloud/), with storage managed by your wallet through onchain payments.**
```

rvagg (approved):

README.md
BigLep, line 1: Latest propose mainnet changes look good to me (including Rod's suggestions).

README.md
SgtPooki: done in [`df6404f` (this PR)](https://github.com/filecoin-project/filecoin-pin/pull/445/commits/df6404f5dd4840c91a1888225b563adeb27ae22d)

PR #448, opened 2026-05-05
feat(unixfs)!: adopt IPIP-499 unixfs-v1-2025 profile

### LINKED ISSUE #306: Adopt IPIP-499 default "modern" UnixFS CID profile
Ref: https://github.com/ipfs/specs/pull/499

We currently differ from the proposed `unixfs-v1-2025` profile in the following ways:

* Directory wrapping: we wrap in a directory so we can put a name on the file, this has come up before as something we could potentially not do
* Hidden files: we include them
* HAMT threshold: Helia only supports what they're calling `links-bytes`

The benefit of conforming to a standard profile will be that when we pack a file or directory for a user, they end up with the same root CID as they get when they use any other conforming implementation. This may be of interest to the kind of people who care about NFTs or other items where reproducibility matters. It's also a general UX thing because people just don't tend to understand how all this works and complain or question when they end up with different CIDs, so consistency makes a class set of user complaints go away. How much do we care though?

SgtPooki:
## What changed

Configure `@helia/unixfs` with the IPIP-499 `unixfs-v1-2025` profile so root CIDs reproduce across conforming implementations (Kubo, Boxo, Helia) for the same input. Drop the directory wrapper that `add` previously inserted around single files; the source basename is routed into piece metadata as `filename` (or `dirname` for directories) so the human-readable name survives. Hidden directory entries are excluded by default; opt back in with `--include-hidden`.

This is the breaking-change follow-up agreed in #306 (decision: 2026-04-09).

## How to verify

- `pnpm install && pnpm run build && pnpm test`
- Run `filecoin-pin add <file>` and confirm:
  - Root CID matches `ipfs add --cid-version=1 --raw-leaves` (or any other conforming implementation) for the same input.
  - Returned piece metadata includes `filename` (or `dirname` for a directory).
  - Output no longer includes a wrapping directory block.
- `filecoin-pin add <dir>` excludes dotfiles by default; rerun with `--include-hidden` to include them.

## Notes / risks

- BREAKING: root CIDs change for any single-file input (no parent directory) and for directories that crossed the 256 KiB HAMT threshold under the old `links-bytes` estimation.
- The `--bare` flag is removed. The new profile makes it the default and only behavior.
- User-supplied `--piece-metadata filename=...` continues to win over the auto-derived basename.
- Synapse SDK has no canonical `filename` / `dirname` constant yet; using local string keys. If the SDK adds them, we will switch over.

Closes #306


src/core/unixfs/car-builder.ts
```diff
@@ -178,25 +175,35 @@ async function createCarFromSingleFile(filePath: string, options: CreateCarOptio
...
+    // glob's hidden filter would discard every match because the root path
+    // segment starts with `.`. Force hidden traversal in that case; the user
+    // selected this directory, so its contents go in regardless of the
+    // dotfile-exclusion default.
+    const rootIsHidden = dirName.startsWith('.')
+    const hidden = includeHidden || rootIsHidden
+    logger?.info(
+      { absolutePath, parentDir, dirName, pattern, includeHidden, rootIsHidden, hidden },
+      'Directory structure for UnixFS'
+    )
```
SgtPooki: do we want to force --include-hidden in this case?

SgtPooki (commented):
self review.. lgtm, but one open question about implicitly adding a dir with dot-prefix.

src/core/unixfs/car-builder.ts
```diff
@@ -27,30 +28,38 @@ export type Spinner = {
 // Placeholder CID used during CAR creation (will be replaced with actual root)
 const PLACEHOLDER_CID = CID.parse('bafyaaiaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
 
-// Whether to include hidden files (starting with .) when adding directories
-const INCLUDE_HIDDEN_FILES = true
+// Whether to include hidden files (starting with .) when adding directories.
+// Default false to match Kubo's `ipfs add` default; callers can opt in via `includeHidden`.
```
rvagg:
```suggestion
// Default false according to the unixfs-v1-2025 CID profile.
```
?

src/core/unixfs/car-builder.ts
rvagg: yerp, I think so, the surprise would be worse; I'm not even sure a log is necessary to note the special case either - if they asked for it, they get it

src/core/metadata/index.ts
```diff
@@ -34,6 +47,33 @@ export function normalizeMetadataConfig(input: MetadataConfigInput): MetadataCon
...
+  pieceMetadata: Record<string, string> | undefined,
+  derived: { kind: 'file' | 'directory'; name: string }
+): Record<string, string> | undefined {
+  if (!derived.name) {
+    return pieceMetadata
+  }
+
+  const key = derived.kind === 'directory' ? PIECE_METADATA_DIRNAME_KEY : PIECE_METADATA_FILENAME_KEY
+  const existing = pieceMetadata?.[key]
+  if (existing != null) {
```
rvagg, line 73: so .. empty string hits this, which is fine, but should we document it somewhere? at least in the jsdoc for this function? otherwise it'll look liable for a cleanup; tests would also lock the behviour in

rvagg:
Misc notes from me:

* I'm not convinced about `dirname`, the duality is making me wonder whether `entname`, `name`, `basename` might be better; I'm imagining the downstream use of this which now has to have a double check. Although squashing them into one loses information. Mixed feelings here, I'll come back if I form a firmer opinion but for now I'm just noting it.
* Since we're trying to do conformance and consistency, we need tests to prove that and should import at least a few standard cases that kubo and/or helia would produce. Here's a starting point: https://github.com/ipfs/kubo/blob/master/test/cli/cid_profiles_test.go, https://github.com/ipfs/js-ipfs-unixfs/blob/main/packages/ipfs-unixfs-importer/test/ipip-499-profiles.spec.ts; "hello world" and empty directory are easy. https://github.com/ipfs/js-ipfs-unixfs/blob/main/packages/ipfs-unixfs-importer/test/helpers/deterministic.ts has `deterministicRandomBytes()` so you can copypasta that and do 1 MiB seed "chunk-v1-seed", 1 MiB + 1 byte seed "chunk-v1-seed", 1024 × 1 MiB seed "v1-2025-seed", 1024 × 1 MiB + 1 seed "v1-2025-seed". The ones that hit HAMT transition are a bit annoying and can probably be skipped, but we could use `deterministicFilenames` and cover those threshold cases. AI should be great at this if you point it to those resources.

Otherwise looks pretty good.

SgtPooki:
@rvagg 
> dirname [Qs]

instead of dirname/filename/basename/etc.. I just went with `name` and folks can be pointed to IPFS-y way of determining the kind? see [`d797f49` (this PR)](https://github.com/filecoin-project/filecoin-pin/pull/448/commits/d797f4999c40b24aeeec491dfb44e64461bee2c1)

> conformance and consistency

added in [`aaa451b` (this PR)](https://github.com/filecoin-project/filecoin-pin/pull/448/commits/aaa451bad898cfa7ea19d9f033dedfad5d6d5be0)

src/core/metadata/index.ts
SgtPooki: done

SgtPooki (commented):
self review

src/core/unixfs/browser-car-builder.ts
```diff
@@ -65,45 +69,25 @@ export async function createCarFromFile(file: File, options: CreateCarOptions =
...
-  }
-
-  // Single file with bare mode
-  if (options.bare && files.length === 1 && files[0] != null) {
+  if (files.length === 1 && files[0] != null) {
     return createCarFromFile(files[0], options)
   }
 
-  return createCar(async (fs) => {
+  return createCar({ name: '' }, async (fs) => {
```
juliangruber: can we use `null` instead of empty string?

src/core/unixfs/browser-car-builder.ts
```diff
@@ -167,8 +151,12 @@ export async function createCarFromFileList(files: File[], options: CreateCarOpt
     return createCarFromFiles(files, options)
   }
 
+  // Top-level directory name is the first segment of any webkitRelativePath.
+  const sample = (files.find((f) => (f as any).webkitRelativePath) as any)?.webkitRelativePath as string | undefined
+  const dirName = sample?.split('/')[0] ?? ''
```
juliangruber: similar to above, can we use `null` if no `name`?

juliangruber (approved):

juliangruber:
FYI

```
$ pnpm i
Scope: all 2 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +109 -85
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------------------------------------------------------------------------------
Progress: resolved 109, reused 53, downloaded 56, added 62, done
node_modules/.pnpm/node-datachannel@0.32.3/node_modules/node-datachannel: Running install script, done in 1.4s

dependencies:
- @filoz/synapse-core 0.5.1
+ @filoz/synapse-core 0.5.2
- datastore-core 11.0.2
+ datastore-core 11.0.4
- helia 6.0.22
+ helia 6.1.4
- interface-blockstore 6.0.1
+ interface-blockstore 6.0.2
- interface-store 7.0.1
+ interface-store 7.0.2
- it-to-buffer 4.0.10
+ it-to-buffer 4.0.12

devDependencies:
- @ipld/dag-cbor 9.2.5
+ @ipld/dag-cbor 9.2.6

Done in 3.6s using pnpm v10.32.1
$ LOG_LEVEL=debug ./node_modules/.bin/tsx src/cli.ts add ~/Desktop/photo/photos/000025590037.png
/Users/julian/dev/filecoin-project/filecoin-pin/node_modules/.pnpm/@libp2p+tls@3.1.0/node_modules/@libp2p/tls/src/utils.ts:11
import { Crypto } from '@peculiar/webcrypto'
         ^
SyntaxError: The requested module '@peculiar/webcrypto' does not provide an export named 'Crypto'
    at #asyncInstantiate (node:internal/modules/esm/module_job:302:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:405:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:660:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

```

Same error after `pnpm i --force`

juliangruber:
`pnpm test` passes for me locally, does the above commented CLI command work for you?

src/core/unixfs/browser-car-builder.ts
SgtPooki: Done. Switched the derived name to `null` when there's no basename. `CreateCarResult.name` is now `string | null`, and `withDerivedNameMetadata` treats `null` as "no name to attach". Kept the empty-string check too, since a user-set `''` still means an explicit opt-out from the auto-derived name.

src/core/unixfs/browser-car-builder.ts
SgtPooki: Done. Switched the derived name to `null` when there's no basename. `CreateCarResult.name` is now `string | null`, and `withDerivedNameMetadata` treats `null` as "no name to attach". Kept the empty-string check too, since a user-set `''` still means an explicit opt-out from the auto-derived name.

SgtPooki (commented):
self review

PR #455, opened 2026-05-06
feat(synapse): derive chain from RPC probe

SgtPooki:
## What changed

`--network` and `--rpc-url` (and their `NETWORK` / `RPC_URL` env equivalents) are now mutually exclusive. Commander enforces this at parse time so any CLI invocation that supplies both fails fast with a clear error. `createConfig` mirrors the check for library consumers that bypass Commander.

When `--rpc-url` is set, `initializeSynapse` probes the endpoint's `eth_chainId` and selects the matching chain (mainnet, calibration, or a configured devnet). The Synapse SDK does not verify chainId against the configured chain, so a `--rpc-url` pointing at calibration would otherwise be paired with mainnet contract addresses and fail at the first contract call.

The default network is still mainnet (PR #445); this PR adds the probe + exclusivity on top.

Stacked on PR #445.

## How to verify

- `pnpm run typecheck`
- `pnpm run lint:fix`
- `pnpm run test:unit`
- `pnpm run test:browser`
- `filecoin-pin payments status --rpc-url <calibration-ws>` resolves to calibration without needing `--network`.
- `filecoin-pin payments status --network mainnet --rpc-url wss://x` exits with a Commander conflicts error.
- `NETWORK=mainnet RPC_URL=wss://x filecoin-pin server` errors the same way (env source counted by Commander).

## Notes / risks

Breaking change: any existing caller that sets both `--network` / `NETWORK` and `--rpc-url` / `RPC_URL` will now get a hard error instead of silently using one of them.


src/core/synapse/resolve-chain-from-rpc.ts
```diff
@@ -0,0 +1,34 @@
...
+  const client = createPublicClient({ transport })
+  const id = await client.getChainId()
+
+  if (id === mainnet.id) return mainnet
+  if (id === calibration.id) return calibration
+
+  // Devnet support pulls in node:fs via resolveDevnetConfig, so import it lazily to keep this
+  // module browser-safe. Mainnet/calibration probes never reach this branch.
+  try {
+    const { resolveDevnetConfig } = await import('../../common/get-rpc-url.js')
```
rvagg, line 29: let's avoid await imports, smelly, move to the top

rvagg: doh, sorry, I see your comment .. my bad

src/core/synapse/resolve-chain-from-rpc.ts
```diff
@@ -0,0 +1,34 @@
...
+ *   devnet   → chain from devnet-info.json (if available)
+ *
+ * Throws when the chainId does not match any known chain.
+ */
+import { type Chain, calibration, mainnet } from '@filoz/synapse-sdk'
+import { createPublicClient, type Transport } from 'viem'
+
+export async function resolveChainFromRpc(transport: Transport): Promise<Chain> {
+  const client = createPublicClient({ transport })
+  const id = await client.getChainId()
```
rvagg, line 19: this could be the first chain call and it could be problematic or have significant lang, I think a debug log here might be helpful, and the logger is available in the caller of `resolveChainFromRpc` so we could just pass it in here and then if something goes wrong or it takes a while at least we can see what it might be blocking on

rvagg (approved):
👍 "--network and --rpc-url are mutually exclusive", this is good

main suggestion is a bit of logging

SgtPooki:
CI is failing due to provider/chain failure, not related to these changes.

PR #456, opened 2026-05-06
refactor!: rename onProgress event types

### LINKED ISSUE #91: feat: migrate callback API to onProgress with typed responses
https://github.com/filecoin-project/filecoin-pin/blob/300ecd58f4132410c401a2dae45073975d98e9a2/src/core/synapse/index.ts#L185-L189

Instead of surfacing the granularity of callbacks that exist in synapse-sdk, we should allow consumers to pass an `onProgress` handler that receives an event for any of the callbacks. Similar to what we have in `checkUploadReadiness`:

```ts

/**
 * Options for evaluating whether an upload can proceed.
 */
export type UploadReadinessProgressEvent =
  | { type: 'checking-balances' }
  | { type: 'checking-allowances' }
  | { type: 'configuring-allowances' }
  | { type: 'allowances-configured'; transactionHash?: string }
  | { type: 'validating-capacity' }

export interface UploadReadinessOptions {
  /** Initialized Synapse instance. */
  synapse: Synapse
  /** Size of the CAR file (bytes). */
  fileSize: number
  /**
   * Automatically configure allowances when they are missing.
   * Defaults to `true` to match current CLI/action behaviour.
   */
  autoConfigureAllowances?: boolean
  /** Optional callback for progress updates. */
  onProgress?: (event: UploadReadinessProgressEvent) => void
}
```

what do you think @rvagg ?

SgtPooki:
Closes #91.

Unifies the `type` strings across every `onProgress` event union: camelCase, no `on` prefix, colon for namespaced flows. Covers upload, readiness, IPNI, and remove-piece/remove-all events. Updated `upload-action/src/filecoin.js` and a few stale doc/comment refs while I was in there.

New doc at `documentation/progress-events.md` describing the convention so we don't drift again.

Breaking for anyone matching `event.type`. Pre-major, so release-please will treat `feat!` as a minor bump.

## How to verify
`pnpm run build && pnpm test && pnpm run lint:fix`

### DIFF
--- a/documentation/progress-events.md
+++ b/documentation/progress-events.md
@@ -1,68 +1,47 @@
 # Progress Events
 
-filecoin-pin exposes progress through a single `onProgress` handler per public function rather than a bag of granular `onX` callbacks. This document describes the convention, the rationale, and the rules for naming new event types.
-
-## Why one `onProgress` instead of many callbacks
-
-The synapse-sdk uses an object-of-callbacks shape (`{ callbacks: { onProviderSelected, onDataSetResolved, ... } }`). filecoin-pin wraps that into a single typed-event-union handler.
-
-Reasons:
-- One subscription point per call. Callers do not have to remember to wire a new callback every time we add an event.
-- Discriminated unions force exhaustive handling at the type level. New events show up as compile errors in `switch` statements without a `default`.
-- Stable surface area as the SDK evolves. We can map new SDK callbacks into existing event types or add new ones without changing the public function signature.
-- Easy fan-out and logging. Consumers can log every event with one line: `onProgress: (e) => log.info(e)`.
+Public functions in filecoin-pin report progress through a single `onProgress` handler that receives a discriminated union of typed events. This document describes the event shape, the naming convention, and the rules for adding new events.
 
 ## Shape
 
-Defined in `src/core/utils/types.ts`:
+Event and handler types are defined in [`src/core/utils/types.ts`](../src/core/utils/types.ts):
 
-```ts
-export type ProgressEvent<T extends string = string, D = undefined> = D extends undefined
-  ? { type: T }
-  : { type: T; data: D }
+- `ProgressEvent<Type, Data>` — discriminated event with a `type` string and optional `data` payload.
+- `ProgressEventHandler<Events>` — handler that receives a union of events.
+- `AnyProgressEvent` — base type used as the default constraint, shaped `{ type: string; data?: unknown }`.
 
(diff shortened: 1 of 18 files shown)


SgtPooki (commented):
self review..

documentation/progress-events.md
```diff
@@ -0,0 +1,106 @@
+# Progress Events
```
rvagg, line 1: this reads like a critique of Synapse choices; "synapse did it wrong, we're going to do it right for you, and here's why our opinions are correct"

documentation/progress-events.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Each function exports a discriminated union of its events and accepts `onProgress?: ProgressEventHandler<MyEvents>`.
+
+## Naming convention
+
+The `type` discriminator string follows these rules:
+
+1. **camelCase** for words. No kebab-case, no PascalCase, no `snake_case`.
+2. **No `on` prefix** on the type string. The handler is already called `onProgress`; the prefix is redundant. Use `stored`, not `onStored`.
```
rvagg: this one sounds more like a description of the change being made rather than being future oriented documentation, could be rephrased "unlike the Synapse callbacks ..." I suppose

documentation/progress-events.md
```diff
@@ -0,0 +1,106 @@
...
+3. **Colon (`:`) namespace separator** when an event belongs to a logical sub-flow whose events would otherwise collide with other unions or be ambiguous in logs. Use `removePiece:submitting`, not `removePieceSubmitting`.
+4. **No dot (`.`) separator.** Dots imply a property path; we use them only inside `data`, never in `type`.
```
rvagg: again, read this as your future self, why are you being told this one? I reckon remove it

documentation/progress-events.md
```diff
@@ -0,0 +1,106 @@
...
+5. **Both sides of the colon are camelCase.** `removePiece:confirmationFailed`, not `removePiece:confirmation-failed`.
```
rvagg: covered in 1

documentation/progress-events.md
SgtPooki: 😬 lmao.. did not mean it to come off that way!

rvagg (approved):
good change, just needs a doc polish: read it as a future user who has no history of how it was, also consider the framing wrt Synapse, it sounds critical/snobbish

juliangruber (approved):

PR #228, opened 2025-11-11
[WIP] chore: saving load-test scripts in draft PR

SgtPooki:
related to https://github.com/filecoin-project/filecoin-pin-website/issues/103

PR #458, opened 2026-05-06
fix(auth): explicit auth flags no longer lose to env vars

### LINKED ISSUE #457: Epic: wallet auth without raw private keys
### Description

`filecoin-pin` wallet auth is a raw private key handed over as an env var: `PRIVATE_KEY` everywhere, plus session-key secrets in CI and server configs. No encrypted-at-rest option, no OS keychain, and the recommended GitHub Actions setup stores the wallet key as a repo secret.

This is the tracking epic for the signer backends that replace raw-key auth. OpenWallet Standard (OWS) was evaluated as a backend (spike: #458) and rejected: as of 2026-08-17 its repo had no maintainer commit, review, or merge since 2026-06-19, the sole human maintainer had been removed from the npm packages on 2026-08-03, the SECURITY.md contact routed to that same removed identity, and a policy-bypass report ([open-wallet-standard/core#228](https://github.com/open-wallet-standard/core/issues/228)) had been open without response since May. Evidence: [OWS proof-of-life memo](https://gist.github.com/SgtPooki/c44b6a8a667bb2c80a48d5e0767b8cce). The OWS backend was removed from #458; the auth seam it was built on ships without it.

The seam this epic hangs on is in two PRs: #458 adds source-aware auth-mode resolution in `src/utils/cli-auth.ts`, and #687 added the injectable viem `Account` accepted by `initializeSynapse()`. That seam covers ordinary CLI commands only. The pinning server builds private-key/session-key configs directly, the upload action requires `walletPrivateKey` and calls `initializeSynapse({privateKey})`, and session owner commands call `privateKeyToAccount()`. Unifying those surfaces is its own child issue below.

We deliberately do not recommend cloud KMS (AWS/GCP) backends. filecoin-pin is a reference implementation for Filecoin Onchain Cloud; pointing users at centralized cloud HSMs is off-mission. The injectable `Account` seam means anyone who wants a KMS-backed signer can wire one up themselves; we document the seam, not a vendor.

### Impact

- CI users store the wallet's raw private key as a repo secret when a scoped, revocable session key would bound the blast radius.
- Server operators leave keys readable in process env and shell history, and the daemon holds the decrypted key in memory for its lifetime.
- Blocks adoption by teams with key-handling policies.

### Child issues

Each child is tagged with its primary beneficiary: `[agent]` agent runtimes, `[human]` operators and CLI users, `[risk]` containment for everyone.

- [ ] #681 `[human]` keystore v3: encrypted keystore files + passphrase prompt, the recommended CLI default (port from synapse-sdk [#554](https://github.com/FilOzone/synapse-sdk/pull/554)); prerequisite for keychain
- [ ] #682 `[agent]` session keys as the recommended CI path: upload-only permission profile (CreateDataSet+AddPieces only; the default permission set also grants SchedulePieceRemovals and TerminateService), upload-action `sessionKey` input, expiry/runway warnings, rotation story, secret redaction
- [ ] #679 `[agent]` unify auth surfaces: pinning server, upload action, and session owner commands consume the same signer resolution as the CLI; prerequisite for any hosted-MCP or daemon auth story
- [ ] #680 `[human]` OS keychain storage via `@napi-rs/keyring`; headless unlock for assistant-spawned local MCP tools is scoped to session-key material, never the owner keystore
- [ ] #589 `[human]` password-manager vaults (1Password/Bitwarden/LastPass); outside the agent mapping

### Out of scope

- OpenWallet Standard (OWS). The #458 spike integrated it; we evaluated upstream health and removed the backend from the PR rather than shipping it as experimental. Rationale with evidence: [OWS proof-of-life memo](https://gist.github.com/SgtPooki/c44b6a8a667bb2c80a48d5e0767b8cce). The source-aware precedence the spike built ships in #458 without it; the injectable viem `Account` landed in #687.
- Hardware wallets (Ledger/Trezor); separate issue if requested.
- First-party cloud KMS backends (AWS/GCP); the seam supports bring-your-own `Account`, documented in the session-key child.
- Devnet keeps auto-resolving from `devnet-info.json`.

Recommended posture regardless of backend: a dedicated hot wallet holding only working funds, capped via `payments setup` allowances (`setMaxAllowances`), funded from a cold wallet. Document this alongside whichever backend lands first.

SgtPooki:
### What changed

`parseCLIAuth` resolves the mutually exclusive auth flags by where each value came from instead of a blind if/else chain. An explicit flag beats an env var. Two explicit modes are a hard error. When every supplied mode came from the environment, the canonical order picks the winner (view address, then session key, then private key, the same order master used) and a warning names the ignored variables. A lone `--wallet-address` or `--session-key` never outranks a complete mode.

Provenance comes from Commander's `getOptionValueSource()` through a `preAction` hook in `addAuthOptions`, stashed on the parsed options as `optionSources`.

Interactive `payments setup` no longer prompts for a private key when a complete session key is supplied. It refuses the session key through `assertOwnerAuth`, matching `payments setup --auto`.

Part of #457. This PR began as an OpenWallet Standard spike. The OWS backend was removed in 7928f6f after an upstream health check ([OWS proof-of-life memo](https://gist.github.com/SgtPooki/c44b6a8a667bb2c80a48d5e0767b8cce)). The injectable viem `Account` on `initializeSynapse()` that the spike also built landed on master in #687.

### How to verify

```
pnpm install && pnpm run build && pnpm test
```

`src/test/unit/cli-auth.test.ts` covers explicit-beats-env, env-only tie-break with warning, explicit conflicts, lone session-key halves, and the devnet fallback. `src/test/unit/run-interactive-setup.test.ts` covers the prompt suppression and the session-key refusal.

### Notes

No flags added or removed. Two behaviour changes from master: a warning when more than one auth mode is exported in the environment, and a hard error when two modes are passed as explicit flags (master silently picked one).


SgtPooki:
@HavenCTO @apha205, updated the PR.. feel free to leave feedback

src/utils/cli-auth.ts
```diff
@@ -72,9 +82,19 @@ export function parseCLIAuth(options: CLIAuthOptions): SynapseSetupConfig {
...
+  if (owsWalletId) {
+    const owsOptions: Parameters<typeof getOwsAccount>[0] = {
+      walletId: owsWalletId,
+      chain: chain ?? calibration,
+    }
+    if (owsPassphrase != null) owsOptions.passphrase = owsPassphrase
+    config.account = await getOwsAccount(owsOptions)
+  } else if (privateKey) {
+    config.privateKey = privateKey
+  }
```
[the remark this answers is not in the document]
SgtPooki: Fixed in b8570f3. parseCLIAuth resolves the auth mode in precedence order (read-only, session key, then owner signer), so the OWS adapter only loads when an owner signer is actually needed and the config no longer mixes modes.

src/utils/cli-auth.ts
```diff
@@ -78,24 +91,44 @@ export function parseCLIAuth(options: CLIAuthOptions): SynapseSetupConfig {
...
     config.walletAddress = viewAddress
     config.readOnly = true
+  } else if (walletAddress && sessionKey) {
+    config.walletAddress = walletAddress
+    config.sessionKey = sessionKey
+  } else if (owsWalletId) {
+    const owsOptions: Parameters<typeof getOwsAccount>[0] = {
+      walletId: owsWalletId,
+      chain: chain ?? calibration,
+    }
```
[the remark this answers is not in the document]
SgtPooki: Fixed in 1dbb700. When --rpc-url is set we now probe the endpoint's chainId (same resolveChainFromRpc helper initializeSynapse uses) before requesting the OWS account, so the eip155 hint matches the target RPC. No --network and no --rpc-url falls back to mainnet, matching the CLI default.

src/utils/cli-auth.ts
```diff
@@ -78,24 +91,44 @@ export function parseCLIAuth(options: CLIAuthOptions): SynapseSetupConfig {
...
+    if (owsPassphrase != null) owsOptions.passphrase = owsPassphrase
+    config.account = await getOwsAccount(owsOptions)
```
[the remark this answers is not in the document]
SgtPooki: Fixed in 1dbb700. Empty OWS_WALLET_PASSPHRASE is treated as not provided.

src/utils/cli-options.ts
```diff
@@ -77,11 +77,24 @@ export function addSigningAuthOptions(command: Command): Command {
...
-  addSigningAuthOptions(command).addOption(
-    new Option('--view-address <address>', 'View-only mode (no signing) for the specified wallet address').env(
-      'VIEW_ADDRESS'
+  addSigningAuthOptions(command)
+    .addOption(
+      new Option(
+        '--wallet <id>',
+        'OpenWallet Standard wallet name or ID (signs via OWS, key never leaves vault). Can also use OWS_WALLET_ID env.'
+      ).env('OWS_WALLET_ID')
+    )
```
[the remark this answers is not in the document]
SgtPooki: Right.. session commands use addOwnerAuthOptions and the runners take a raw private key into core/session, so OWS there means changing createSessionKey to accept an Account. More than this spike should carry. Updated the PR description to scope OWS to commands using addAuthOptions; session support is follow-up work.

src/utils/cli-options.ts
```diff
@@ -77,11 +77,24 @@ export function addSigningAuthOptions(command: Command): Command {
...
+    .addOption(
+      new Option(
+        '--wallet-passphrase <passphrase>',
+        'Passphrase for the OWS wallet (also OWS_WALLET_PASSPHRASE env)'
+      ).env('OWS_WALLET_PASSPHRASE')
+    )
```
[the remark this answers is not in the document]
SgtPooki: Same story for server: it builds SynapseSetupConfig from env in filecoin-pinning-server.ts, not through parseCLIAuth. Scoped out of this PR, description updated.

SgtPooki:
ping @hugomrdias @apha205 @BravoNatalie

package.json
```diff
@@ -139,6 +139,8 @@
     "@libp2p/tcp": "^11.0.17",
     "@multiformats/multiaddr": "^13.0.1",
     "@multiformats/multiaddr-to-uri": "^12.0.0",
+    "@open-wallet-standard/adapters": "^1.3.2",
+    "@open-wallet-standard/core": "^1.3.2",
```
rvagg: npm download numbers aren't exactly a vote of confidence, but at least they mention Filecoin in their readme and have native support, neato https://github.com/open-wallet-standard/core/pull/51

src/utils/cli-auth.ts
```diff
@@ -78,24 +92,51 @@ export function parseCLIAuth(options: CLIAuthOptions): SynapseSetupConfig {
...
   } = {}
 
-  if (privateKey) config.privateKey = privateKey
   if (viewAddress) {
     config.walletAddress = viewAddress
     config.readOnly = true
+  } else if (walletAddress && sessionKey) {
+    config.walletAddress = walletAddress
+    config.sessionKey = sessionKey
+  } else if (owsWalletId) {
```
rvagg: This sitting above `if (privateKey)` means that if you happen to have a `OWS_WALLET_ID` in your environment but supply an explicit `--private-key` then it'll get ignored, silently. Since we are dealing with mutually exclusive options here we should be careful about obeying the explicit args over env vars. You should be able to use commander (`getOptionValueSource`) to figure out if this was an env var or not and work our precedence. Another option might just be to put `privateKey` above this so it takes precedence if set in any way. `sessionKey` is a bit similar too .. but it gets even more complicated then when `walletAddress` coupling.

I think you should lay out precedence rules here for all the mutually exclusive options and then be explicit in working out what comes from where and not just have a blind fall-through else/if with silent failures.

src/utils/cli-auth.ts
```diff
@@ -54,13 +60,21 @@ export interface CLIAuthOptions {
  * @param options - CLI authentication options
  * @returns Synapse setup config (validation happens in initializeSynapse)
  */
-export function parseCLIAuth(options: CLIAuthOptions): SynapseSetupConfig {
+export async function parseCLIAuth(options: CLIAuthOptions): Promise<SynapseSetupConfig> {
```
rvagg, line 146: continuing on my precedence suggestion below, the complexity in here now deserves a bunch of tests to show it does what you expect, and document what you expect too

rvagg:
@SgtPooki did your testing extend beyond the `payments` command into actual data onboarding? This `JSON.stringify` looks like a problem for our bigints in 712 payloads, does it get past this? https://github.com/open-wallet-standard/core/blob/76c8c67d6c73bf57805432fc15a2e48d12168c6c/bindings/node-adapters/src/viem.js#L36

package.json
rvagg: you should bump versions too, they are up to [1.4.2](https://www.npmjs.com/package/@open-wallet-standard/core?activeTab=versions) and that version introduces trusted publishing so it'd be the preferred one to start with

src/utils/cli-auth.ts
SgtPooki: precedence rules are in now

SgtPooki:
> @SgtPooki did your testing extend beyond the `payments` command into actual data onboarding? This `JSON.stringify` looks like a problem for our bigints in 712 payloads, does it get past this? [open-wallet-standard/core@`76c8c67`/bindings/node-adapters/src/viem.js#L36](https://github.com/open-wallet-standard/core/blob/76c8c67d6c73bf57805432fc15a2e48d12168c6c/bindings/node-adapters/src/viem.js#L36)

I didn't do super thorough testing, but I have now. I fixed this in our implementation here, added unit tests, and also opened an upstream fix: github.com/open-wallet-standard/core#242

src/utils/cli-auth.ts
SgtPooki: preferences should be much more clear now

src/core/ows/index.ts
```diff
@@ -36,23 +37,68 @@ interface OwsViemAdapter {
...
+ *   digits"), so `0x0` and `0x1ab` are not valid.
+ *
+ * Even-length hex satisfies both and produces signatures identical to viem
+ * across 0, 2^128 - 1, 2^128, and 2^256 - 1.
+ *
+ * Negative values fall back to decimal. synapse-sdk uses only uint EIP-712
+ * fields; the hex form here is unsigned.
+ */
+function bigintToOwsHex(value: bigint): string {
+  if (value < 0n) return value.toString()
```
rvagg, line 97: what's going on here?

rvagg (approved):
tbh this was a very light re-review; the only thing that stands out to my eye is the hex <0n thing, otherwise as long as this works in practice then it seems good to me to ship as an experimental feature
@jennijuju might like to try and integrate this into the skill(s) she's working on

Apha205:
@SgtPooki @rvagg , tested against Haven CLI

The CLI can authenticate and sign transactions using an OWS-backed viem `Account` (`OWS_WALLET_ID`) without requiring access to a raw private key.
- `OWS_WALLET_ID` successfully resolved to EVM account address
- Synapse initialized using the OWS-backed viem `Account` path and selected two Calibration storage providers
- `pieces_added` transactions were submitted and signed by the OWS-managed wallet
- Upload completed successfully

No `PRIVATE_KEY` was set or used during the test.

Screenshots from the debug logs attached below.
Screenshot 2026-08-04 045802

Screenshot 2026-08-04 051138

Screenshot 2026-08-04 045614

src/core/ows/index.ts
SgtPooki: just to prevent `0x-1` when trying to convert from any negative bigint passed through OWS. shouldn't happen based on looking through code in syanpse-sdk, but it could happen in the future..

SgtPooki:
> it seems good to me to ship as an experimental feature

@rvagg do you have any preference on how we communicate this as an "experimental feature" ?

rvagg:
@SgtPooki nope; I'll leave that to you, I'm not demanding you even take action on making it "experimental", I'm just noting that "OpenWallet" is pretty young and not very well used and obviously has rough edges; it suggests some caution in enshrine it in a way that might be easy to yank it out in the future.

SgtPooki:
Following up on the "pretty young and not very well used" concern: I dug into OWS upstream health and you were more right than either of us knew. No maintainer commit, review, or merge since 2026-06-19; the sole human maintainer was removed from the npm packages on 2026-08-03 and his GitHub account is gone; SECURITY.md routes to that same removed identity; and a policy-bypass report ([#228](https://github.com/open-wallet-standard/core/issues/228), `wallet delete --confirm` skips the policy engine and all safety gates) has sat unanswered since May. Full evidence with dates: https://gist.github.com/SgtPooki/c44b6a8a667bb2c80a48d5e0767b8cce

So rather than enshrine it, or even flag it experimental, I removed the OWS backend from this PR entirely (7928f6f). What remains is the part worth merging: source-aware auth precedence resolution and the injectable viem `Account` seam in `initializeSynapse`, which the keystore and session-key work in #457 builds on. I'll update the PR title/description to match.

src/utils/cli-auth.ts
rvagg: no more need for `async` in here, you can drop that and all of the `await` additions to make the diff shrink significantly

rvagg (approved):
approval pending removal of `async` and the `await`s **and renaming this before it lands as a misnamed commit**

PR #459, opened 2026-05-07
feat: detect CAR input in add and refuse to wrap

### LINKED ISSUE #199: Provide a warning when `add`ing a CAR
When the user does a `filecoin-pin add foo.car` then it's likely not what they want to be doing (it may be, but I can't think of why). We should instead present a clear warning to them and suggest `import` is what they want.

* We should probably proceed anyway cause it's what they asked for and we don't do interactive-mode for `add`
* We _could_ fail as long as we provide a `--force` or similar override for this case, cause it could be legitimate

Using [`decoder.readHeader(reader)`](https://github.com/ipld/js-car?tab=readme-ov-file#async-decoderreadheaderreader) (from a `@ipld/car/decoder` import) should be enough to do this ~gracefully. If it doesn't error then it's almost certainly a CAR.

rvagg:
Sniff the CAR header via @ipld/car/decoder in createCarFrom* and throw with code: INPUT_IS_CAR. The add CLI catches early and points users at 'filecoin-pin import'.

Closes: #199

---

I saw https://github.com/filecoin-project/filecoin-pin/pull/275 was closed and figured this was a very easy addition to fire off. Low priority but it does resolve a footgun.

### DIFF
--- a/src/add/add.ts
+++ b/src/add/add.ts
@@ -5,12 +5,14 @@
  * It encodes content as UnixFS, creates CAR files, and uploads to Filecoin.
  */
 
+import { createReadStream } from 'node:fs'
 import { readFile, stat } from 'node:fs/promises'
 import pc from 'picocolors'
 import pino from 'pino'
 import { warnAboutCDNPricingLimitations } from '../common/cdn-warning.js'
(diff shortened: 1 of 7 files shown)


SgtPooki (approved):
lgtm

PR #462, opened 2026-05-07
fix(ci): stabilize Windows test runner

SgtPooki:
### What changed

Hardened `src/test/unit/upload-action.test.ts` so `process.exit` spy and dynamic `vi.doMock` always restore via `try/finally` + `vi.doUnmock` + `vi.resetModules`.

The failing job on #445 reported `Worker exited unexpectedly` after all 423 tests passed. Root cause: the PR-comment test mutated `process.exit` and registered a dynamic module mock without guaranteed cleanup, leaving worker state inconsistent under parallel execution on Windows.

### How to verify

`pnpm test` on all platforms; Windows CI run on this PR should not reproduce the worker crash.

Refs: #445

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,19 @@
 name: Test
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+concurrency:
+  group: test-${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


rvagg (approved):
sure, why not

SgtPooki:
I can't believe these jobs are all still running.. going to cancel and restart

PR #467, opened 2026-05-08
feat(cli): add --cdn flag to add and import commands

### LINKED ISSUE #464: Add --egress-provider flag with beam as default
### Done criteria

- [ ] `--egress-provider beam` is the default behavior for `filecoin-pin add` and `filecoin-pin import` (no flag required to enable it)
- [ ] `--egress-provider none` is available to opt out
- [ ] The CLI help text and documentation explain what `--egress-provider beam` means (see below)

### Background / How things work today

Currently, opting into Filecoin Beam (FilBeam) CDN support requires passing a raw metadata flag:

```
filecoin-pin add --data-set-metadata withCDN='true'
```

This is not user-friendly and exposing the term "cdn" (either as metadata flag above or as a `--cdn` as originally proposed) is currently confusing per the the detailed explanation in https://github.com/filecoin-project/filecoin-pin/pull/467#issuecomment-4418757726. The short version: FilBeam today supports **piece/whole-CAR retrieval** only (i.e. `/piece/baga...`), **not** IPFS-style block retrieval (`/ipfs/bafy...`). Advertising this as "CDN support" or "faster IPFS retrievals" would be inaccurate for **`filecion-pin`** users, who think in IPFS terms.

### Why now

We are approaching mainnet GA and expect more paying customers to onboard, including Storacha migrations. We want those datasets positioned to benefit from improved retrieval in the future without requiring users to take action or the dev team to do migration work. This is a deliberate bet that FilBeam's IPFS retrieval capabilities will mature (tracked at https://github.com/filbeam/roadmap/issues/85), and we're setting datasets up accordingly today.

### Why make `beam` the **default** rather than opt-in

If it stays opt-in, the explanation burden is too high. Users would need to understand what FilBeam is, what piece retrieval means, why they might incur charges, and what the future benefit is - all before seeing any present-day upside. Most users won't opt in under those conditions. Default-on avoids that problem; users who don't want it can opt out with `--egress-provider none`.

### Implementation specifics

**CLI change:** add `--egress-provider <provider>` to `filecoin-pin add` and `filecoin-pin import`, with `beam` as the **default** (no flag required to opt in).

```
# default behavior — beam is active, no flag needed:
filecoin-pin add ...

# explicitly opt out:
filecoin-pin add --egress-provider none

# explicitly opt in (same as default):
filecoin-pin add --egress-provider beam
```

Under the hood, `--egress-provider beam` maps to passing `withCDN=true` in the dataset metadata to the synapse SDK.

### Documentation required for `--egress-provider beam`

The docs (help text, glossary, or relevant guide) should cover the following:

**What it does today:** When `--egress-provider beam` is active, your uploaded CAR will be retrievable from the FilBeam CDN using the piece CID via piece retrieval. This is a Filecoin-specific concept. Traditional IPFS users can safely ignore it, but can read more in https://github.com/filecoin-project/filecoin-pin/blob/master/documentation/glossary.md#piece-retrieval.

**What's coming:** FilBeam retrieval for IPFS blocks (`/ipfs/...` paths) is in development. See https://github.com/filbeam/roadmap/issues/85 for updates. When that ships, datasets with FilBeam already enabled will benefit automatically.

**Cost implications:** This option may incur egress costs for the dataset owner. Any `/piece` retrieval made through FilBeam's endpoint will be charged to your account. Users on standard IPFS tooling (see https://github.com/filecoin-project/filecoin-pin/blob/master/documentation/glossary.md#standard-ipfs-tooling) will not trigger FilBeam retrievals and will not be charged.

**Opting out:** If you don't want this behavior, pass `--egress-provider none`.

### Open things not currently answered by this issue description:
1. Where someone learns more about FilBeam and its retrieval endpoints.
2. Where are all the places we want the documentation? Without further direction, I would assume just in the cli help text.

beck-8:
Adds `--cdn` as a shorthand alias for `--data-set-metadata withCDN=true` on both the add and import commands, making Filecoin Beam opt-in more discoverable. The existing `WITH_CDN` env var is preserved for backwards compatibility.

Closes #464

### DIFF
--- a/src/add/add.ts
+++ b/src/add/add.ts
@@ -87,7 +87,7 @@ export async function runAdd(options: AddOptions): Promise<AddResult> {
   })
 
   // Check CDN status and warn if enabled
-  const withCDN = process.env.WITH_CDN === 'true'
+  const withCDN = process.env.WITH_CDN === 'true' || options.cdn === true
   if (withCDN) {
     const proceed = await warnAboutCDNPricingLimitations()
     if (!proceed) {
--- a/src/add/add.ts
+++ b/src/add/add.ts
@@ -87,7 +87,7 @@ export async function runAdd(options: AddOptions): Promise<AddResult> {
   })
 
   // Check CDN status and warn if enabled
-  const withCDN = process.env.WITH_CDN === 'true' || options.cdn === true
+  const withCDN = options.cdn === true
   if (withCDN) {
     const proceed = await warnAboutCDNPricingLimitations()
     if (!proceed) {
--- a/src/add/types.ts
+++ b/src/add/types.ts
@@ -13,6 +13,8 @@ export interface AddOptions extends CLIAuthOptions, CLIAutoFundOptions {
   dataSetMetadata?: Record<string, string>
   /** Skip IPNI advertisement verification after upload */
   skipIpniVerification?: boolean
+  /** Enable Filecoin Beam (CDN) for faster IPFS retrievals */
+  cdn?: boolean
 }
 
 export interface AddResult {
--- a/src/import/import.ts
+++ b/src/import/import.ts
@@ -145,7 +145,7 @@ export async function runCarImport(options: ImportOptions): Promise<ImportResult
   })
 
   // Check CDN status and warn if enabled
-  const withCDN = process.env.WITH_CDN === 'true' || options.cdn === true
+  const withCDN = options.cdn === true
   if (withCDN) {
     const proceed = await warnAboutCDNPricingLimitations()
     if (!proceed) {
--- a/src/import/import.ts
+++ b/src/import/import.ts
@@ -145,7 +145,7 @@ export async function runCarImport(options: ImportOptions): Promise<ImportResult
   })
 
   // Check CDN status and warn if enabled
-  const withCDN = process.env.WITH_CDN === 'true'
+  const withCDN = process.env.WITH_CDN === 'true' || options.cdn === true
   if (withCDN) {
     const proceed = await warnAboutCDNPricingLimitations()
     if (!proceed) {
--- a/src/import/types.ts
+++ b/src/import/types.ts
@@ -12,6 +12,8 @@ export interface ImportOptions extends CLIAuthOptions, CLIAutoFundOptions {
   dataSetMetadata?: Record<string, string>
   /** Skip IPNI advertisement verification after upload */
   skipIpniVerification?: boolean
+  /** Enable Filecoin Beam (CDN) for faster IPFS retrievals */
+  cdn?: boolean
 }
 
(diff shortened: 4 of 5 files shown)


SgtPooki (changes requested):
copilot has some good feedback, and we should also update the `addUploadOptions` for --cdn to read the WITH_CDN env var there.

rvagg:
> Enable Filecoin Beam (CDN) for faster IPFS retrievals

You can't say this, and we really shouldn't be making a big deal of `cdn` for IPFS uploads; which is why I'm uncomfortable in general with even adding this flag. I know it's been requested but it's a place-holder for possible future work that hasn't been prioritised and may never materialise.

Let me try and be clear about the problem here:

* Filecoin Beam operates on piece retrievals, `/piece/baga...` only, nothing about `/ipfs/bafy...` retrievals.
* Filecoin Pin is IPFS-focused; we expect our users will be thinking in IPFS terms, we give them an IPFS Root CID for this reason. They (mostly, maybe) expect that when they upload with Filecoin Pin, that it's available via "the IPFS network" (gateway, Kubo, inbrowser.link, etc.). That all depends on `/ipfs/bafy...` retrievals.
* If you are to use Filecoin Beam today to retrieve a piece uploaded with Filecoin Pin, all you're getting is the whole CAR in which the content was packed. Some users may know what to do with it, for some workflows this makes perfect sense, I can imagine a few of them, but it's pretty advanced.

So, advertising that a CDN flag and suggesting it's going to be helpful today for your IPFS data is wrong. There is a possible future world where we make Filecoin Beam do something special with IPFS data, but that's not today and there's no prioritisation of that.

There's a secondary problem here for users: enabling CDN means you're on the hook for piece retrieval egress. That means that if someone wants to grab the CAR of your data then you'll be charged for it. Anyone can look on chain and find pieces and perform Filecoin Beam retrievals, users are on the hook for it. Not the per-IPLD block retrievals that /ipfs/ retrievals are going to do, but the whole piece. As per Slack converstations we're apparently OK with doing default-on for CDN and opting users into this world, but I want it to be clear to everyone what the implications of that are - for the most part, if a user of Filecoin Pin opts in to CDN and they get charged a CDN bill, then it's very unlikely they're paying for egress that contributes to their data being on "the IPFS network".

So if we go ahead with this, document it accordingly, ideally we background this and not make it something  that gives users the wrong impression.

BigLep:
@beck-8 : the done criteria in https://github.com/filecoin-project/filecoin-pin/issues/464 have changed (see updated issue). As a result, I'm going to put this back into draft/todo until it's been updated. Or feel free to close if you think a fresh start should be taken.

beck-8:
https://github.com/filecoin-project/filecoin-pin/pull/486

PR #482, opened 2026-05-13
feat: accept 'calibnet' as alias for --network calibration

### LINKED ISSUE #480: Accept "calibnet" as an alias for "calibration" in --network
Don't need to advertise it, just accept it, smile and nod.

Jeff-Kazzee:
## What this changes

Accepts `calibnet` as an input alias for `calibration` everywhere the `--network` flag and `NETWORK` env var are parsed. The alias is never surfaced in `--help`; the canonical names (`mainnet`, `calibration`, `devnet`) remain the documented choices, per @rvagg's comment on the issue.

## Implementation

A small shared helper `normalizeNetworkName(input)` in `src/common/get-rpc-url.ts` lowercases, trims, and applies an alias table (currently just `calibnet → calibration`). It is called from:

- `getRpcUrl()` — replaces the inline `.toLowerCase().trim()` so library consumers and the CLI both pick up aliases.
- `resolveChain()` in `src/config.ts` — same reasoning for the chain-hint path.
- The Commander `--network` option's `argParser` in `src/utils/cli-options.ts` — runs *before* `.choices(['mainnet', 'calibration', 'devnet'])` validates, so `--network calibnet` is rewritten to `calibration` and accepted without the alias appearing in `--help` output.

No documentation changes. No new public exports the alias table itself — only the helper.

## Tests

Five new test cases:

- `src/test/unit/get-rpc-url.test.ts` — `getRpcUrl({ network: 'calibnet' })` resolves to the calibration RPC, including uppercase/whitespace variants.
- `src/test/unit/cli-options.test.ts` — `--network calibnet` and `NETWORK=calibnet` both produce `options.network === 'calibration'`. Plus a negative-coverage test that `--help` output contains `mainnet`, `calibration`, `devnet` but *not* `calibnet`, locking in the "don't advertise it" requirement.

Run locally with `pnpm run lint && pnpm run typecheck && pnpm run test:unit`. (Will paste container output in the PR thread if helpful.)

## Closes

Closes #480.

## AI assistance disclosure

Drafted with AI assistance; reviewed line-by-line by the contributor against an adversarial reviewer prompt before submission. Happy to iterate on any feedback.

### DIFF
--- a/src/common/get-rpc-url.ts
+++ b/src/common/get-rpc-url.ts
@@ -11,6 +11,19 @@ const NETWORK_CHAINS = {
   mainnet,
   calibration,
 } as const
+
+// Aliases accepted on input and rewritten to a canonical network name.
+// Not surfaced in --help; the canonical names remain the documented choices.
+const NETWORK_ALIASES: Record<string, string> = {
+  calibnet: 'calibration',
+}
+
+/** Lowercase, trim, and apply NETWORK_ALIASES. Undefined for empty input. */
+export function normalizeNetworkName(input: string | undefined): string | undefined {
+  const trimmed = input?.toLowerCase().trim()
+  if (!trimmed) return undefined
+  return NETWORK_ALIASES[trimmed] ?? trimmed
+}
 function getDefaultDevnetInfoPath(): string {
   const baseDir = process.env.FOC_DEVNET_BASEDIR?.trim() || join(homedir(), '.foc-devnet')
   return join(baseDir, 'state', 'latest', 'devnet-info.json')
@@ -85,7 +98,7 @@ export function getRpcUrl(options: CLIAuthOptions): string {
     return options.rpcUrl
   }
 
-  const network = options.network?.toLowerCase().trim()
+  const network = normalizeNetworkName(options.network)
   if (network) {
     if (network === 'devnet') {
       const devnet = resolveDevnetConfig()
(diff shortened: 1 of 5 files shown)


rvagg:
check failing tests @Jeff-Kazzee, it's the new one you added here

The problem you're facing is `choices()` is overriding your parser, it does two things: sets the help text and sets an `argsParser` which then overrides the one you added. You want the help text but you also want a custom `argsParser` so you have to do both but in the other order: `choices()` first, then an `argsParser()` that does what the `choices()` args parser does and more, i.e. it has to validate the correct list as well. So set up a `const ALLOWED_NETWORKS = ['mainnet', 'calibration', 'devnet']`, use that for a `choices()` then also use it for an `argsParser()` function after you've normalised.

Jeff-Kazzee:
Thanks, that was exactly it. `.choices()` was clobbering the `argParser`. Applied your suggested structure: shared `ALLOWED_NETWORKS`, `.choices()` first, then an `argParser` that normalizes the alias and re-validates against the same list. Tests pass locally (the `--network calibnet` / `NETWORK=calibnet` cases and the `--help` negative test). Pushed.

SgtPooki (approved):
lgtm, going to re-request a copilot review

BigLep:
@SgtPooki : good to merge now?

rjan90:
I did a quick skim/review of this PR as well, and can't see any reasons to block merge here. So I´m going forward and will squash merge it now.

PR #488, opened 2026-05-18
fix: calculate storage from active piece CIDs

### LINKED ISSUE #281: fix: calculate-actual-storage.ts can be improved.

shrirajpawar4:
Closes #281

  Updates the actual storage calculation to walk active pieces and sum raw sizes from PieceCIDs via `getSizeFromPieceCID()`, instead of using FR32-expanded PDP leaf counts.

  This builds on @ameeetgaikwad’s original optimization work in #383 while incorporating the sizing correction from review feedback.

  Tested with:

  - `pnpm test`


BigLep:
Thank you for creating. This may be delayed in getting reviewed for a couple of weeks as we focus on critical deliverables for GA (M4.2).

shrirajpawar4:
no worries. Can i assist in someway for those critial deliverables?

BigLep:
> Can i assist in someway for those critial deliverables?

Thanks for asking @shrirajpawar4 . We'll surface if we see anything. I'm mostly "treading water" for staying on top of issues so haven't been able to think to clearly on ways for others to come in and help, but we'll let you know. Thanks!

beck-8 (dismissed):
Thank you for your contribution. But most of the logic in this PR is outdated.

We should fix the header to only handle excluding the `OFFCHAIN_ORPHANED` field

`fix: exclude off-chain-orphaned pieces from actual storage total`

shrirajpawar4:
oh okay, will make changes according to updated files

BigLep:
@shrirajpawar4: go ahead and mark as "ready for review" once feedback has been incoporated. We'll be sure to reviewer sooner this next time around. Thank you!

beck-8:
I want others to confirm it too

BigLep:
@BravoNatalie : can you please review, and feel free to pull in @SgtPooki if needed.

BravoNatalie (changes requested):
I don't think this PR actually closes #281.

The core ask in #281 was to stop walking every piece just to compute the total size, but we're still doing that here. 

`getDataSetSizes` (in `synapse-core`) already does what #281 proposed. It takes all dataset IDs, performs a single multicall with `getDataSetLeafCount` for each dataset, and returns the total bytes. That makes it O(1) RPC round-trips regardless of how many pieces each dataset contains. It also naturally excludes `OFFCHAIN_ORPHANED`, since those pieces were never written on-chain.

One small thing that's worth documenting in the code is that `getDataSetSizes` returns the **padded** byte size, not the raw content size. I think that's fine, since it's what is actually stored and billed.

rvagg:
> the padded byte size, not the raw content size. I think that's fine, since it's what is actually stored and billed

Not quite, leaf count is the fr32 expanded size rounded up to the next 32 bytes. So you (a) need to unexpand it to find the actual size otherwise it's going to be nearly 0.8% overestimated, and (b) decide that being off by 32 is OK.

Per https://github.com/FilOzone/filecoin-services/pull/471 we charge by the unexpanded leaf-count approximated size, so we do actually charge by the nearest-32 number, so it's probably OK to do the same here, just don't use the raw leaf count, we've got ourselves in trouble with that before. See https://github.com/FilOzone/pdp/pull/266, we can redo the same thing here. Although I think actually the right thing to do is fix it in Synapse, it looks like we let https://github.com/FilOzone/synapse-sdk/pull/632 slip in the same problem with `getDataSetSizes` that we had to fix in https://github.com/FilOzone/filecoin-services/issues/451

rvagg:
https://github.com/FilOzone/synapse-sdk/issues/859

shrirajpawar4:
Updated this to use aggregate dataset sizes instead of walking pieces. Also unexpands the current Synapse leaf-byte result with 127/128 so we don’t overcount while the upstream issue is pending.

BravoNatalie (approved):

PR #501, opened 2026-05-21
feat: instrument multi-copy upload outcomes

### LINKED ISSUE #363: Instrument piece adds to gage multi-copy success rate
## Done Criteria
1. We can see what proportion of multicopy "piece additions" succeed
2. In the case where they fail, we have visibility into what step failed and which SP failed

## Why Important
As identified during ORR, we are lacking visbility into how successful we are the multi-copy golden path.  Filecoin Pin is a place we can actually measure this.
 
## Notes
1. Before picking this up, lets do a bit more desing about the specific metric names/definitions emitted to make sure we don't overengineer and are on the same page.

juliangruber:
## Summary
- Adds a single `uploadCopyStatus` counter around `executeUpload`, tagged with `value: success | failure.pull | failure.commit | failure.other`, `spId`, `role`, plus the common tags `affordance` (which surface emitted the metric: `CLI`, `GitHub Action`, `Library`, or `Filecoin Pin Website`) and `network`.
- Each `executeUpload` call fires one HTTP POST to [BetterStack's metrics ingestion endpoint](https://betterstack.com/docs/logs/ingesting-data/http/metrics/) containing one counter point per copy and per failed attempt — no in-memory buffer, no periodic timer. CLI, pinning server, and Action `await shutdownTelemetry()` before exit so any in-flight request finishes.
- The CLI sets `affordance: 'CLI'` and the GitHub Action sets `affordance: 'GitHub Action'` at entrypoint. Library consumers default to `'Library'`; the website passes `'Filecoin Pin Website'`. `configureTelemetry` validates the value at runtime and throws on anything else.
- Wires opt-out (`FILECOIN_PIN_TELEMETRY_DISABLED=true` / `DO_NOT_TRACK=1`) and overrides (`FILECOIN_PIN_METRICS_ENDPOINT`, `FILECOIN_PIN_METRICS_TOKEN`).
- Adds [`documentation/events-and-metrics.md`](documentation/events-and-metrics.md) as the source-of-truth spec for the events Filecoin Pin emits and the metrics they produce, modelled on [FilOzone/dealbot's events-and-metrics doc](https://github.com/FilOzone/dealbot/blob/main/docs/checks/events-and-metrics.md) (per https://github.com/filecoin-project/filecoin-pin/issues/363#issuecomment-4514202570).
- BetterStack endpoint and source token are embedded in [`src/core/telemetry/constants.ts`](src/core/telemetry/constants.ts). The token ships in source by design; the data source is treated as public/untrusted.

Implements the proposal in https://github.com/filecoin-project/filecoin-pin/issues/363#issuecomment-4506524939.

## Why direct HTTP, not OpenTelemetry?

The original implementation used `@opentelemetry/sdk-metrics` + `OTLPMetricExporter`. That's the OTel-idiomatic path, but for our needs it was overkill:

- Four runtime deps (`@opentelemetry/api`, `…/exporter-metrics-otlp-http`, `…/resources`, `…/sdk-metrics`) for a single counter.
- BetterStack accepts the same metric semantics over its own [JSON HTTP endpoint](https://betterstack.com/docs/logs/ingesting-data/http/metrics/), so the OTel layer added abstraction without adding capability.
- Per-event submission is fine at our cadence (one POST per upload). No buffer, no interval timer, no exit-flush hook needed — pending fetches naturally keep the Node loop alive, and `shutdownTelemetry()` awaits them for explicit teardown.

The replacement is ~150 lines: a single `post()` helper, a `Set` of in-flight promises, and a sync `recordUploadResult` that fires-and-tracks. Same public API (`configureTelemetry`, `recordUploadResult`, `flushTelemetry`, `shutdownTelemetry`).

If we later need higher-frequency metrics, histograms, or multi-backend support, we can revisit OTel — but the current shape is one low-frequency counter and direct HTTP fits it well.

## Test plan
- [x] `pnpm run typecheck`
- [x] `pnpm run test:unit` — 461 passed, 11 skipped (incl. 11 telemetry tests, 6 configure-from-env tests)
- [x] `pnpm run test:integration` — 22 passed
- [ ] Manual: confirm BetterStack ingests `uploadCopyStatus` end-to-end

Closes #363

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/cli.ts
```diff
@@ -69,7 +74,14 @@ program.hook('postAction', async (_thisCommand, actionCommand) => {
...
-    process.exit(process.exitCode ?? 0)
+    try {
+      await flushTelemetry()
+    } catch (err) {
+      // Never let a telemetry flush failure block the forced exit below.
+      console.error('Telemetry flush failed:', err)
+    } finally {
+      process.exit(process.exitCode ?? 0)
+    }
   }
```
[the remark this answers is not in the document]
juliangruber: No need as no telemetry would be written yet

src/core/telemetry/index.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Map a SDK failure `error` string to a coarse outcome value. Keep this
+ * short list of buckets — high-cardinality strings would defeat the purpose
+ * of the counter.
+ */
+function classifyFailure(error: string): string {
+  const lower = error.toLowerCase()
+  if (lower.startsWith('pull failed')) return 'failure.pull'
+  if (lower.startsWith('commit failed')) return 'failure.commit'
```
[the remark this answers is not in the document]
SgtPooki: @juliangruber how do you feel about tying a unit test to this so we know if things break during build rather than in runtime?

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,64 @@
...
+- All status count metrics carry an additional `value` tag that attributes the count to a specific outcome.
+
+| Metric | Relevant Events | When Emitted | `value` Values | Source of truth |
+|---|---|---|---|---|
+| <a id="uploadCopyStatus"></a>`uploadCopyStatus` | [`uploadCopyResolved`](#uploadCopyResolved) | Once per copy in the upload result — one per entry in `result.copies` and one per entry in `result.failedAttempts`. | `success` (copy committed by the SP), `failure.pull` (Synapse pull step failed), `failure.commit` (Synapse commit step failed), `failure.other` (any other Synapse-reported failure). | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
+
+In addition to the [common tags](#common-tags) and `value`, `uploadCopyStatus` carries:
+
+- `spId` — [Storage Provider](glossary.md#service-provider) ID, stringified from the SDK's `providerId`.
+- `role` — `primary` or `secondary`, per the Synapse SDK.
```
SgtPooki: should we have an `other` role for 3+ copies?

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,64 @@
...
+
+To compute the **multi-copy success rate** over a window, divide `sum(uploadCopyStatus{value="success"})` by `sum(uploadCopyStatus)`. To isolate a particular failure mode (e.g. commit-step regressions across SPs), filter on `value="failure.commit"` and break down by `spId`.
```
SgtPooki: FYI, I confirmed based on the existing data in betterstack that this will not work as documented.

```suggestion
To compute the **multi-copy success rate** over a window, divide the number of points tagged `value=success` by the total number of `uploadCopyStatus` points. To isolate a failure mode (e.g. commit-step regressions across SPs), filter on `value=failure.commit` and group by the `spId` tag.

> Querying in BetterStack: `uploadCopyStatus` is ingested as a counter, but each point is a flat `value:1`, so the default counter aggregation (`avgMerge(rate_avg)`) reads ~0. Count points with `sum(metrics_count)` and read tags via `label('value')` / `label('spId')` instead.
```

src/core/telemetry/index.ts
```diff
@@ -0,0 +1,161 @@
...
+ * `value` tag of `success | failure.pull | failure.commit | failure.other`.
+ * See `documentation/events-and-metrics.md` for the full schema (events, tags,
+ * and how the metric relates to the SDK's per-upload result).
+ *
+ * Metrics ship via direct HTTP POST to BetterStack
+ * (https://betterstack.com/docs/logs/ingesting-data/http/metrics/). Each call
+ * to {@link recordUploadResult} fires its own HTTP POST — there is no
+ * in-memory buffer or periodic flush. Use {@link flushTelemetry} to await
+ * pending requests before exit. To turn subsequent record calls into no-ops,
+ * call `configureTelemetry({ disabled: true })`.
```
SgtPooki: this is duplicated a few times and could use a "humanize" skill or something for your claude. we should link to documentation instead of repeating in jsdoc where appropriate, and we already link to it above, so this feels superfluous.

src/core/telemetry/index.ts
```diff
@@ -0,0 +1,161 @@
...
+
+import type { UploadResult } from '@filoz/synapse-sdk'
+import { DEFAULT_METRICS_ENDPOINT, DEFAULT_METRICS_TOKEN, METRIC_UPLOAD_COPY_STATUS } from './constants.js'
+
+/**
+ * Which surface the metric was emitted from. Every metric carries this as a
+ * tag so we can slice success/failure rates per UI. Defaults to `'Library'`
+ * when {@link configureTelemetry} is not called with an explicit value.
+ */
+export const AFFORDANCES = ['CLI', 'GitHub Action', 'Library', 'Filecoin Pin Website'] as const
```
SgtPooki: I thought we decided not to instrument library usage?

src/core/telemetry/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  inFlight.add(promise)
+}
+
+/**
+ * Await any in-flight HTTP submissions. Call before exit (or in tests) to
+ * guarantee delivery. Safe when nothing is pending.
+ */
+export async function flushTelemetry(): Promise<void> {
+  if (inFlight.size === 0) return
+  await Promise.all(inFlight)
```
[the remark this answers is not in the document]
SgtPooki: I believe this is a real potential race request. We should add a test for this, cause even if we don't care about flushing everything, it could cause hanging promises for users who aren't process.exit-ing.

upload-action/src/run.mjs
```diff
@@ -1,12 +1,19 @@
...
 import { runBuild } from './build.js'
 import { getErrorMessage, handleError } from './errors.js'
 import { completeCheck, createCheck } from './github.js'
+import { getInput, parseBoolean } from './inputs.js'
 import { getOutputSummary } from './outputs.js'
 import { runUpload } from './upload.js'
 
+configureTelemetry({
+  affordance: 'GitHub Action',
+  disabled: parseBoolean(getInput('disableTelemetry', 'false')),
```
SgtPooki: should we also read env var for the upload-action?

src/read-telemetry-config-from-env.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * CLI-side helper that reads environment variables into a partial
+ * {@link TelemetryConfiguration}, so the caller can compose it with their own
+ * fields (e.g. `configureTelemetry({ ...readTelemetryConfigFromEnv(), affordance: 'CLI' })`).
+ *
+ * The telemetry library never reads `process.env` itself; this file lives
+ * outside `src/core/` because env-var handling is a CLI concern, not a
+ * library responsibility. Other Node hosts (the GitHub Action, embedding
+ * apps) can either reuse this helper or apply their own configuration
+ * policy.
```
SgtPooki: >  * library responsibility. Other Node hosts (the GitHub Action, embedding
>  * apps) can either reuse this helper or apply their own configuration
>  * policy.

can the upload-action call it, it doesn't look like its in our exports map? why aren't we using it in the upload-action?

README.md
```diff
@@ -117,9 +117,34 @@ Filecoin Pin's CLI collects telemetry.  A few things:
...
+* **Error reports** via Sentry (no PII).
+* **Per-upload copy outcomes** posted directly to [BetterStack's HTTP metrics ingestion endpoint](https://betterstack.com/docs/logs/ingesting-data/http/metrics/), so we can measure the success rate of multi-copy uploads and identify which storage providers (or pipeline steps) are failing. One counter is emitted:
+  * `uploadCopyStatus` — incremented once per resolved copy attempt. Tags: `spId` (storage provider ID), `role` (`primary`/`secondary`), `value` (`success`, `failure.pull`, `failure.commit`, `failure.other`), `network`, `affordance` (which surface emitted the metric: `CLI`, `GitHub Action`, `Library`, `Filecoin Pin Website`).
 
-Set the environment variable `FILECOIN_PIN_TELEMETRY_DISABLED=true`.
+  See [`documentation/events-and-metrics.md`](documentation/events-and-metrics.md) for the full schema, including the underlying events and the relationship between this metric and the Synapse SDK's upload result.
+
+  The BetterStack source token is shipped in source, so the data source is treated as public and untrusted. The CLI lets you send your own copy with `FILECOIN_PIN_METRICS_ENDPOINT` and `FILECOIN_PIN_METRICS_TOKEN`.
```
SgtPooki: idk if we should have this in the README
```suggestion
```

README.md
```diff
@@ -117,9 +117,34 @@ Filecoin Pin's CLI collects telemetry.  A few things:
...
+  configureTelemetry({ endpoint: '…', token: '…' })            // send elsewhere
+  configureTelemetry({ affordance: 'Filecoin Pin Website' })   // tag the surface (default 'Library')
+  ```
+
+  The CLI's env-var support is built on top of this API (see `src/read-telemetry-config-from-env.ts`); other Node hosts can follow the same pattern.
+
+### How to disable telemetry
+
+- **CLI / pinning server:** set `FILECOIN_PIN_TELEMETRY_DISABLED=true` (or the cross-tool standard `DO_NOT_TRACK=1`).
+- **GitHub Action:** pass `disableTelemetry: true` to the action (see [`upload-action/action.yml`](upload-action/action.yml)).
```
SgtPooki: we should support the same env vars as CLI here. I'm fine with leaving the action input as well..

SgtPooki (commented):
in general LGTM, but I do have some comments and questions

src/core/telemetry/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  }))
+  try {
+    await fetch(config.endpoint, {
+      method: 'POST',
+      headers: {
+        Authorization: `Bearer ${config.token}`,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(body),
+    })
```
SgtPooki: we should have a signal here as well so fetch isn't unbounded.

src/core/telemetry/index.ts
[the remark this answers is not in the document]
SgtPooki: the loop copilot suggests wouldn't be great for non-exiting callers.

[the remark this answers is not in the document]
juliangruber: The flush method doesn't promise no further requests will be made - this is in control of the library user. If they make additional requests while flushing, the process won't quit either. I understand the suggestion but think the motivation isn't quite right.

In prior iterations of this PR, there was a `shutdownTelemetry` method, which prevented any further requests from being made. I found this to be overengineered, as there was little signal this is actually useful.

src/core/telemetry/index.ts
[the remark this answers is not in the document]
juliangruber: I don't think a unit test would give us that safety, as we can't make Synapse easily produce the relevant error strings, for us to test whether the classifier works correctly. Do you have a suggestion?

documentation/events-and-metrics.md
juliangruber: Synapse only defines primary and secondary https://github.com/FilOzone/synapse-sdk/blob/e699e604f2c765100144efeed326181db17ee706/packages/synapse-sdk/src/storage/manager.ts#L93-L94

src/core/telemetry/index.ts
juliangruber: See https://github.com/filecoin-project/filecoin-pin/issues/363#issuecomment-4551616056:

> "filecoin pin" telemetry is on by default, but this is clearly documented and easy to opt-out.
> If we can instrument at the library level, that's fine/good - lets just make sure we can comply with the above (easy to disable, easy to annotate what affordance we're using).

upload-action/src/run.mjs
juliangruber: I don't see why we should :thinking: Easy to add later on too

src/read-telemetry-config-from-env.ts
juliangruber: I think a github action should be configured primarily via inputs, not environment

README.md
juliangruber: Can you explain why? I don't want to duplicate all settings when the primary configuration method of action inputs works just fine

src/core/telemetry/index.ts
juliangruber: Addressed in 87ddec4: each post is now capped at 10s via `AbortSignal.timeout(POST_TIMEOUT_MS)`. A timeout-abort rejection takes the same swallow path as any other fetch error in `post()`, so the bound never breaks the host.

src/core/telemetry/index.ts
juliangruber: Addressed in dd82e72: trimmed the top-of-file JSDoc to just the spec-link + the env/precondition note, and tightened `recordUploadResult`'s JSDoc the same way. The delivery model now lives only in [`documentation/events-and-metrics.md`](../blob/add/upload-copy-metrics-363/documentation/events-and-metrics.md) and the README.

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+### Event List
+
+| Event | Definition | Source of truth |
+|------|------------|-----------------|
+| <a id="uploadStarted"></a>`uploadStarted` | Filecoin Pin invokes `executeUpload` for one piece. | [`src/core/upload/index.ts`](../src/core/upload/index.ts) |
+| <a id="uploadCopyAttempted"></a>`uploadCopyAttempted` | The Synapse SDK begins one copy attempt against a specific SP. Implicit marker — emitted once per entry that later shows up in `result.copies` or `result.failedAttempts`. | [`@filoz/synapse-sdk`](https://github.com/FilOzone/synapse-sdk) |
+| <a id="uploadCopyResolved"></a>`uploadCopyResolved` | A copy attempt produces a terminal outcome (`success` or one of the `failure.*` values). Drives [`uploadCopyStatus`](#uploadCopyStatus). | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
+| <a id="uploadResultReceived"></a>`uploadResultReceived` | `executeUpload` returns with `{copies, failedAttempts}`. This is the point at which `recordUploadResult` is called and the [`uploadCopyStatus`](#uploadCopyStatus) points are submitted. | [`src/core/upload/index.ts`](../src/core/upload/index.ts) |
```
BigLep: > recordUploadResult

This is the only place this is mentioned. From this doc, it wasn't obvious what this is.

BigLep: > uploadCopyStatus points are submitted

I don't know what these "`uploadCopyStatus` points" are.

BigLep (commented):
I'm out of steam for today and don't have a clear head. I'll look better during my 2026-05-28. 

A couple of things coming to mind (drawing from https://github.com/filecoin-project/filecoin-pin/issues/363#issuecomment-4514202570 )

- From a CLI regard, it would be ideal to which CLI action is being called and to ideally instrument all of them.
- I think we want to know how long the CLI action takes.
- I think we also would be interested in data sizes being used with filecoin-pin (e.g., how much data is being uploaded via filecoin-pin. 

Again, I'll make sure to do better review tomorrow and sorry not to engage better before 2026-05-28 EU started.

README.md
juliangruber: will add support, ChatGPT convinced: https://chatgpt.com/s/t_6a18286aa2288191ba021e2dedf33f59

juliangruber: 019ddc4

upload-action/src/run.mjs
juliangruber: 019ddc4

juliangruber:
> I'm out of steam for today and don't have a clear head. I'll look better during my 2026-05-28.
> 
> A couple of things coming to mind (drawing from [#363 (comment)](https://github.com/filecoin-project/filecoin-pin/issues/363#issuecomment-4514202570) )
> 
>     * From a CLI regard, it would be ideal to which CLI action is being called and to ideally instrument all of them.
> 
>     * I think we want to know how long the CLI action takes.
> 
>     * I think we also would be interested in data sizes being used with filecoin-pin (e.g., how much data is being uploaded via filecoin-pin.
> 
> 
> Again, I'll make sure to do better review tomorrow and sorry not to engage better before 2026-05-28 EU started.

I hear you, but this is scope creep wrt the original issue. The goal - as far as I understand it - wasn't to add generic instrumentation of Pin actions, it was to understand how successful the multi copy upload strategy is.

For understanding multi copy success, it makes sense to add the data size, as that might impact success rate. I'll add that to this PR.

If you agree, could you open a follow-up issue with the other requests?

juliangruber:
> For understanding multi copy success, it makes sense to add the data size, as that might impact success rate. I'll add that to this PR.

bc141c8

documentation/events-and-metrics.md
juliangruber: > > recordUploadResult
> 
> This is the only place this is mentioned. From this doc, it wasn't obvious what this is.

That's the main function introduced in this PR, submitting upload results. It's an internal function.



> > uploadCopyStatus points are submitted
> 
> I don't know what these "`uploadCopyStatus` points" are.

See `### Status Count Related Metrics`

src/core/telemetry/index.ts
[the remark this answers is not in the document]
SgtPooki: a unit test could prevent us from missing the fact that synapse-sdk updated their error message and our runtime telemetry is no longer tagging errors with these cases. 

the unit test would need to be mocking enough of synapse to trigger the error (this seems like the part you're pushing back on) so that when we assert the error contains X, we are sure it does, and runtime telemetry will emit correct details.

src/core/telemetry/index.ts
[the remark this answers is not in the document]
SgtPooki: i'm fine with that reasoning, I guess "Safe when nothing is pending" is the implicit flag for users to know "do not trigger additional requests". maybe a more explicit callout could be helpful.

documentation/events-and-metrics.md
SgtPooki: right. so do we flag n2+ all as "secondary"?

SgtPooki (approved):
code LGTM, but there are some open threads that need addressed:

* recordUploadResult/points language clarification
* what the 3rd copy will show up as in metrics
* jsdoc comment clarification for consumers of `flushTelemetry`
* free-form error string matching should at least be looked into for something more robust, or this will eventually fail on us and we won't remember this gotcha

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+
+Metrics ship as [direct HTTP POSTs](https://betterstack.com/docs/logs/ingesting-data/http/metrics/) to BetterStack from [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts).
+
+### Common Tags
+
+Every metric emitted by Filecoin Pin carries the following tags:
+
+| Tag | Values | Definition |
+|---|---|---|
+| <a id="tag-affordance"></a>`affordance` | `CLI`, `GitHub Action`, `Library`, `Filecoin Pin Website` | Which [affordance](../README.md#affordances) emitted the metric. Set by the host via `configureTelemetry({ affordance })`; defaults to `Library` when not configured. `configureTelemetry` rejects any other value at runtime. |
```
BigLep:
```suggestion
| <a id="tag-affordance"></a>`affordance` | `CLI`, `GitHub Action`, `Library`, `Website` | Which [affordance](../README.md#affordances) emitted the metric. Set by the host via `configureTelemetry({ affordance })`; defaults to `Library` when not configured. `configureTelemetry` rejects any other value at runtime. |
```

For CLI, GitHub Action, and Library, the "Filecoin Pin" portion is implied. Do we do the same for the website? Or maybe for the website we do pin.filecoiin.cloud to be clear?

What's the reason to reject any other values? What do we gain with the strictness?

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+| <a id="tag-network"></a>`network` | `mainnet`, `calibration`, `devnet` | The Filecoin network the upload targeted. |
+
+### Status Count Related Metrics
+
+- These count metrics track the occurrence of a particular outcome for an event.
+- All status count metrics carry an additional `value` tag that attributes the count to a specific outcome.
+
+| Metric | Relevant Events | When Emitted | `value` Values | Source of truth |
+|---|---|---|---|---|
+| <a id="uploadCopyStatus"></a>`uploadCopyStatus` | [`uploadCopyResolved`](#uploadCopyResolved) | Once per copy in the upload result — one per entry in `result.copies` and one per entry in `result.failedAttempts`. | `success` (copy committed by the SP), `failure.pull` (Synapse pull step failed), `failure.commit` (Synapse commit step failed), `failure.other` (any other Synapse-reported failure). | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
```
BigLep, line 59:
```suggestion
| <a id="uploadCopyStatus"></a>`uploadCopyStatus` | [`uploadCopyResolved`](#uploadCopyResolved) | Once per copy in the upload result — one per entry in `result.copies` and one per entry in `result.failedAttempts`. | `success` (copy committed by the SP), `failure.commit` (Synapse commit step failed),`failure.pull` (Synapse pull step failed), `failure.other` (any other Synapse-reported failure). | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
```

(nit) but putting commit before pull because the pull usually happens after commit.

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+
+In addition to the [common tags](#common-tags) and `value`, `uploadCopyStatus` carries:
+
+- `spId` — [Storage Provider](glossary.md#service-provider) ID, stringified from the SDK's `providerId`.
+- `role` — `primary` or `secondary`, per the Synapse SDK.
+
+To compute the **multi-copy success rate** over a window, divide the number of points tagged `value=success` by the total number of `uploadCopyStatus` points. To isolate a failure mode (e.g. commit-step regressions across SPs), filter on `value=failure.commit` and group by the `spId` tag.
+
+> Querying in BetterStack: `uploadCopyStatus` is ingested as a counter, but each point is a flat `value:1`, so the default counter aggregation (`avgMerge(rate_avg)`) reads ~0. Count points with `sum(metrics_count)` and read tags via `label('value')` / `label('spId')` instead.
```
BigLep: This is probably my ignorance, but is the `value:1` different than the `value` label we have defined?

I'm also not following why something that is a flat `value:1` would have an average that read as ~0 instead of ~1.

------
Oh from looking at https://betterstack.com/docs/logs/ingesting-data/http/metrics/ I see the "value" is a first class term for the counter. Maybe we shouldn't do this to avoid confusion between counter "value" and label/tag "value"? 

I think this goes away if we go with my idea below of just having a single `uploadCopyBytes` gauge metric. 

Or if we don't like that, maybe we don't use a label/tag called "value" but instead set the label/tag as "status"? That deviates from what dealbot does though...

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+| Metric | Relevant Events | When Emitted | `value` Values | Source of truth |
+|---|---|---|---|---|
+| <a id="uploadCopyStatus"></a>`uploadCopyStatus` | [`uploadCopyResolved`](#uploadCopyResolved) | Once per copy in the upload result — one per entry in `result.copies` and one per entry in `result.failedAttempts`. | `success` (copy committed by the SP), `failure.pull` (Synapse pull step failed), `failure.commit` (Synapse commit step failed), `failure.other` (any other Synapse-reported failure). | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
+
+In addition to the [common tags](#common-tags) and `value`, `uploadCopyStatus` carries:
+
+- `spId` — [Storage Provider](glossary.md#service-provider) ID, stringified from the SDK's `providerId`.
+- `role` — `primary` or `secondary`, per the Synapse SDK.
+
+To compute the **multi-copy success rate** over a window, divide the number of points tagged `value=success` by the total number of `uploadCopyStatus` points. To isolate a failure mode (e.g. commit-step regressions across SPs), filter on `value=failure.commit` and group by the `spId` tag.
```
BigLep: I don't love this from the regard that it isn't answering: "what proportion of golden-path 'adds' succeed" where "succeed" means all the copies are stored. 

If someone specifies only one copy, that shouldn't pollute the "golden path" metric. Similarly, if someone specifies they want 100 copies but 99 of those fail, I don't want that tanking the overall success rate of the multi-copy flow. 

This was the motivation behind the `multicopy` proposal in https://github.com/filecoin-project/filecoin-pin/issues/363#issuecomment-4514202570

Does that make sense? 

Anyways, from a pragmatic regard though, what you have is probably good enough for now. I think I would then just want to document it accordingly.

```suggestion
To compute the **upload success rate of all copies (independent of "primary" or "secondary")** over a window, divide the number of data points tagged `value=success` by the total number of `uploadCopyStatus` data points. To isolate a failure mode (e.g. commit-step regressions across SPs), filter on `value=failure.commit` and group by the `spId` tag.

Note that this computation doesn't directly answer "what proportion of golden-path 'adds' succeed" since:
1. "adds" that only want a single upload (no extra copies) will be included AND
2. "adds" that specify more than a single extra copy will add more weight to the metric
[#516](https://github.com/filecoin-project/filecoin-pin/issues/516) is tracking more accurately answer this question.
```

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
+# Filecoin Pin Events & Metrics
+
+This document is the intended **source of truth** for the events emitted by Filecoin Pin and the metrics computed from them. It is intended for Filecoin Pin dashboard consumers and maintainers who need to understand what each metric means and where it comes from.
+
+> **Note on "events":** the entries in the [Event List](#event-list) are named **timing markers** used to define when metrics are emitted — they are not all emitted as discrete events or log lines. Each marker is anchored in code (as a timestamp variable, log line, or status transition) and used to compute the metrics in the [Metrics](#metrics) section.
+
+## Upload Event Model
+
+An upload (`executeUpload` in the [JavaScript library](../README.md#-javascript-library)) asks the [Synapse SDK](glossary.md#synapse) to replicate the same [Piece](glossary.md#piece) across one primary and zero or more secondary copies. Each copy is an independent attempt against a single [Storage Provider](glossary.md#service-provider); the SDK reports one outcome per copy in a single result object. Filecoin Pin records one telemetry event per resolved copy outcome.
```
BigLep: I realize now we're being consistent with Synapse terminology, but the word "copy" was confusing me when thinking about these metrics.  As a result, the text below is an attempt to clarify it.

```suggestion
An upload (`executeUpload` in the [JavaScript library](../README.md#-javascript-library)) asks the [Synapse SDK](glossary.md#synapse) to replicate the same [Piece](glossary.md#piece) from Synapse across one primary and zero or more secondary copies. A "copy" is an independent attempt against a single [Storage Provider](glossary.md#service-provider). Even the the initial upload from the Synapse client to a Storage Provider is a "copy". The SDK reports one outcome per "copy" in a [single result object](https://docs.filecoin.cloud/reference/filoz/synapse-sdk/synapse/interfaces/uploadresult/#copies). Filecoin Pin records one telemetry event per resolved "copy" outcome. (Learn more about [copies in FOC](glossary.md#piece-copy).)
```

I'm thinking we could also add something like the following to the glossary:

```
## Piece Copy
In Filecoin Pin and Synapse, a copy is one independent storage placement of the same Piece on a single Service Provider (usually in its own Data Set). All copies share the same PieceCid; they differ by which SP holds the bytes and records the piece on chain.

This is not the everyday sense of “original plus one duplicate.” Here, N copies means N stored instances across N providers, not N extras on top of an unnamed original. Even the first upload target is called a copy with role "primary".  (See [Synapse upload docs](https://docs.filecoin.cloud/reference/filoz/synapse-sdk/storage/classes/storagemanager/#upload).)
```

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Caller as Filecoin Pin caller<br/>(CLI / GitHub Action /<br/>Library / Website)
+  participant Synapse as Synapse SDK
+  participant SP as PDP Storage Provider
+
+  Caller->>Synapse: uploadStarted
+  loop For each copy
```
BigLep:
```suggestion
  loop For each copy (primary or secondary)
```

documentation/events-and-metrics.md
BigLep:
```suggestion
| <a id="uploadResultReceived"></a>`uploadResultReceived` | `executeUpload` returns with `{copies, failedAttempts}`. This is the point at which `recordUploadResult` is called and the [`uploadCopyStatus`](#uploadCopyStatus) data points are submitted. | [`src/core/upload/index.ts`](../src/core/upload/index.ts) |
```

I think I had confusion with the 
>"uploadCopyStatus" points
(particularly the "points" portion)

I was confused because of 
1. English "point in time"
2. BetterStack doesn't use the term in https://betterstack.com/docs/logs/ingesting-data/http/metrics/

I think we should use "metric" (or ingested metric) in events-and-metrics.md or use "data point" if want want OTel familiarity.

For now I just suggested "data point".

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+| <a id="uploadCopyAttempted"></a>`uploadCopyAttempted` | The Synapse SDK begins one copy attempt against a specific SP. Implicit marker — emitted once per entry that later shows up in `result.copies` or `result.failedAttempts`. | [`@filoz/synapse-sdk`](https://github.com/FilOzone/synapse-sdk) |
+| <a id="uploadCopyResolved"></a>`uploadCopyResolved` | A copy attempt produces a terminal outcome (`success` or one of the `failure.*` values). Drives [`uploadCopyStatus`](#uploadCopyStatus). | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
+| <a id="uploadResultReceived"></a>`uploadResultReceived` | `executeUpload` returns with `{copies, failedAttempts}`. This is the point at which `recordUploadResult` is called and the [`uploadCopyStatus`](#uploadCopyStatus) points are submitted. | [`src/core/upload/index.ts`](../src/core/upload/index.ts) |
+
+## Metrics
+
+Metrics ship as [direct HTTP POSTs](https://betterstack.com/docs/logs/ingesting-data/http/metrics/) to BetterStack from [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts).
```
BigLep:
```suggestion
Metrics ship as [direct HTTP POSTs](https://betterstack.com/docs/logs/ingesting-data/http/metrics/) to BetterStack from [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts). One "upload" typically produces several metric data points (one per "copy" outcome).
```

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+> Querying in BetterStack: `uploadCopyStatus` is ingested as a counter, but each point is a flat `value:1`, so the default counter aggregation (`avgMerge(rate_avg)`) reads ~0. Count points with `sum(metrics_count)` and read tags via `label('value')` / `label('spId')` instead.
+
+### Gauge Metrics
+
+- These metrics carry a single numeric reading per point.
+- Each gauge point is emitted alongside the corresponding [`uploadCopyStatus`](#uploadCopyStatus) point, sharing its tag set so the two metrics can be joined at query time.
+
+| Metric | Relevant Events | When Emitted | Gauge Value | Source of truth |
+|---|---|---|---|---|
+| <a id="uploadCopySize"></a>`uploadCopySize` | [`uploadCopyResolved`](#uploadCopyResolved) | Once per copy in the upload result, paired with [`uploadCopyStatus`](#uploadCopyStatus). | Piece size in bytes (`UploadResult.size`). All copies of one upload share the same size — the value identifies the upload that produced the outcome. | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
```
BigLep:
```suggestion
| <a id="uploadCopyBytes"></a>`uploadCopyBytes` | [`uploadCopyResolved`](#uploadCopyResolved) | Once per copy in the upload result, paired with [`uploadCopyStatus`](#uploadCopyStatus). | Piece size in bytes (`UploadResult.size`). All copies of one upload share the same size — the value identifies the upload that produced the outcome. | [`src/core/telemetry/index.ts`](../src/core/telemetry/index.ts) |
```

Make this self-documenting when looking at the metric name?

documentation/events-and-metrics.md
```diff
@@ -0,0 +1,77 @@
...
+- `role` — `primary` or `secondary`, per the Synapse SDK.
+
+To compute the **multi-copy success rate** over a window, divide the number of points tagged `value=success` by the total number of `uploadCopyStatus` points. To isolate a failure mode (e.g. commit-step regressions across SPs), filter on `value=failure.commit` and group by the `spId` tag.
+
+> Querying in BetterStack: `uploadCopyStatus` is ingested as a counter, but each point is a flat `value:1`, so the default counter aggregation (`avgMerge(rate_avg)`) reads ~0. Count points with `sum(metrics_count)` and read tags via `label('value')` / `label('spId')` instead.
+
+### Gauge Metrics
+
+- These metrics carry a single numeric reading per point.
+- Each gauge point is emitted alongside the corresponding [`uploadCopyStatus`](#uploadCopyStatus) point, sharing its tag set so the two metrics can be joined at query time.
```
BigLep: I'm wondering if instead of `uploadCopyStatus` and `uploadCopySize` we just do `uploadCopyBytes` (rename of `uploadCopyBytes`). I think `uploadCopyBytes` alone can give us everything we're wanting here. We then have a label/tag for "status" which can be `success` (copy committed by the SP), `failure.commit` (Synapse commit step failed),`failure.pull` (Synapse pull step failed), `failure.other` (any other Synapse-reported failure).

Does that make sense?

README.md
```diff
@@ -117,9 +117,33 @@ Filecoin Pin's CLI collects telemetry.  A few things:
 * With our [end user affordance](#affordances) we expect to make telemetry on by default, requiring a consumer/user to opt out.  We are defaulting as "enabled" to help make sure we have a good pulse on the user experience and can address issues correctly.
 * In this [pre-v1 season](https://github.com/filecoin-project/filecoin-pin/issues/187), we are particularly focused on helping maintainers validate functionality and iron out problems throughout the whole Filecoin Onchain Cloud stack that `filecoin-pin` relies on.
```
BigLep:
```suggestion
## Telemetry

Filecoin Pin collects telemetry.  A few things:
* Telemetry always [has a way to be disabled](#how-to-disable-telemetry).
* We don't collect Personal identifiable information (PII).
* [Affordance](#affordances) make telemetry on by default, requiring a consumer/user to opt out.  We are defaulting as "enabled" to help make sure we have a good pulse on the user experience and can address issues correctly. Maintainers are particularly focused on validating functionality and ironing out problems throughout the whole Filecoin Onchain Cloud stack that `filecoin-pin` relies on.
```

README.md
```diff
@@ -117,9 +117,33 @@ Filecoin Pin's CLI collects telemetry.  A few things:
...
+What we collect:
+* **Error reports** via Sentry (no PII).
+* **Per-upload copy outcomes** posted directly to [BetterStack's HTTP metrics ingestion endpoint](https://betterstack.com/docs/logs/ingesting-data/http/metrics/), so we can measure the success rate of multi-copy uploads and identify which storage providers (or pipeline steps) are failing. Two metrics are emitted per resolved copy:
+  * `uploadCopyStatus` counter — one increment per resolved copy attempt. Tags: `spId` (storage provider ID), `role` (`primary`/`secondary`), `value` (`success`, `failure.pull`, `failure.commit`, `failure.other`), `network`, `affordance` (which surface emitted the metric: `CLI`, `GitHub Action`, `Library`, `Filecoin Pin Website`).
+  * `uploadCopySize` gauge — the piece size in bytes for the same outcome, sharing the counter's tag set so dashboards can join the two (e.g. p99 size of commit-step failures).
 
-Set the environment variable `FILECOIN_PIN_TELEMETRY_DISABLED=true`.
+  See [`documentation/events-and-metrics.md`](documentation/events-and-metrics.md) for the full schema, including the underlying events and the relationship between this metric and the Synapse SDK's upload result.
```
BigLep:
```suggestion
* **Per-upload copy outcomes** posted directly to [BetterStack's HTTP metrics ingestion endpoint](https://betterstack.com/docs/logs/ingesting-data/http/metrics/), so we can measure the success rate of multi-copy uploads and identify which storage providers (or pipeline steps) are failing. See [`documentation/events-and-metrics.md`](documentation/events-and-metrics.md) for the full schema, including the underlying events and the relationship between this metric and the Synapse SDK's upload result.
```

Have one place where we're documenting the specifics...

README.md
```diff
@@ -238,6 +262,12 @@ HOST=localhost                 # Daemon server host
 DATABASE_PATH=./pins.db        # SQLite database location
 CAR_STORAGE_PATH=./cars        # CAR file storage directory
 LOG_LEVEL=info                 # Logging verbosity (info, debug, error)
+
+# Optional - Telemetry (see "CLI Telemetry" above)
```
BigLep:
```suggestion
# Optional - Telemetry (see "Telemetry" above)
```

README.md
```diff
@@ -238,6 +262,12 @@ HOST=localhost                 # Daemon server host
...
+FILECOIN_PIN_TELEMETRY_DISABLED=true        # Disable all telemetry
+DO_NOT_TRACK=1                              # Standard cross-tool opt-out
+FILECOIN_PIN_METRICS_ENDPOINT=https://…     # Override BetterStack metrics endpoint
+FILECOIN_PIN_METRICS_TOKEN=…                # Override BetterStack source token
```
BigLep: Are there options for the "Sentry" metrics endpoint?

README.md
```diff
@@ -117,9 +117,33 @@ Filecoin Pin's CLI collects telemetry.  A few things:
...
+  * `uploadCopySize` gauge — the piece size in bytes for the same outcome, sharing the counter's tag set so dashboards can join the two (e.g. p99 size of commit-step failures).
 
-Set the environment variable `FILECOIN_PIN_TELEMETRY_DISABLED=true`.
+  See [`documentation/events-and-metrics.md`](documentation/events-and-metrics.md) for the full schema, including the underlying events and the relationship between this metric and the Synapse SDK's upload result.
+
+
+  **Delivery model.** Each `executeUpload` fires its own HTTP POST containing one counter point per copy and per failed attempt — there is no in-memory buffer or periodic flush. The CLI, pinning server, and GitHub Action `await flushTelemetry()` before exit so any in-flight request finishes. **Long-running consumers that terminate via `process.exit()`, `SIGINT`, or `SIGTERM` should do the same** (`flushTelemetry` is exported from `filecoin-pin/core/telemetry`). To silence subsequent `recordUploadResult` calls without exiting the process, call `configureTelemetry({ disabled: true })`.
```
BigLep: Head is feeling blurry, but I was struggling a little on reasoning why these specific details about executeUpload are in the readme vs. events-and-metrics.md. 

I don't have strong convictions here...

upload-action/README.md
```diff
@@ -17,10 +17,12 @@ See the two-workflow approach in the [examples directory](./examples/) for compl
 See [action.yml](./action.yml) for complete input documentation including:
 - **Core**: `path`, `walletPrivateKey`, `network`
 - **Financial**: `minStorageDays`, `filecoinPayBalanceLimit`
-- **Advanced**: `egressProvider`, `dryRun`
+- **Advanced**: `egressProvider`, `dryRun`, `disableTelemetry`
 
 **Outputs**: `ipfsRootCid`, `dataSetId`, `pieceCid`, `providerId`, `providerName`, `carPath`, `uploadStatus`
 
+> **Disabling telemetry.** Either set the `disableTelemetry: true` input, or set `FILECOIN_PIN_TELEMETRY_DISABLED=true` / `DO_NOT_TRACK=1` in the job's `env:` block. Either signal opts out; see the [CLI Telemetry section of the root README](../README.md#cli-telemetry) for what is collected.
```
BigLep:
```suggestion
> **Disabling telemetry.** Either set the `disableTelemetry: true` input, or set `FILECOIN_PIN_TELEMETRY_DISABLED=true` / `DO_NOT_TRACK=1` in the job's `env:` block. See the [CLI Telemetry section of the root README](../README.md#telemetry) for more info.
```

documentation/events-and-metrics.md
BigLep: Also, will there be a PR in filecoin-pin-website after this?

BigLep (commented):
Hi @juliangruber. I've had brain fog today so I don't know if the comments below are coherent. I was tempted just to do a PR on top with what I'm proposing, but I think it's best if I don't spend more time on this at the moment. If there's anything I'm saying that isn't clear, I can try drafting it up.

At this point, I'm good with what you decide.

BigLep:
@juliangruber :

On

> add generic instrumentation of Pin actions

That's fair on the scope creep. I think I was just mentally assuming we had that when we stripped out sentry from Synapse since the Sentry integration in Synapse before was giving this by proxy.

That said, I can't justify that we need this. I'm good to stay focused on what we specifically need, and wait to add more when specific questions come in.

The way I see it biting us to not have all actions implemented is we'll miss signal on what functionality is or isn't used. But regardless, you are right, that belonga in a separate PR. 

Backlog item created: https://github.com/filecoin-project/filecoin-pin/issues/515

documentation/events-and-metrics.md
BigLep: I'm good to ignore this. I was thinking about the "primary" case as the first case where there isn't "pull". But for the "secondary" case, your order is right. I"m fine to ignore this.

documentation/events-and-metrics.md
BigLep: I keep feeling torn on what to do here. I think it's most clear if we change from using the label/tag of "value" to instead be "status".

documentation/events-and-metrics.md
BigLep: Stepping away and coming back to this with a fresh head, I think this is probably a bad idea.

A gauge alone makes counting successes awkward; the counter+gauge pair is dealbot-aligned and supports “count failures” and “p99 failure size” cleanly. If anything, maybe we just rename the "value" label/tag to "status" as the highest-leverage fix.

src/core/telemetry/index.ts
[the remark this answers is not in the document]
juliangruber: Follow up issue: https://github.com/FilOzone/synapse-sdk/issues/807

documentation/events-and-metrics.md
juliangruber: It's ultimately out of scope for this PR, I think it only makes sense to follow Synapse here. To confirm: Any copy that is pulled from the primary will be called secondary, even if there are multiple secondary copies.

src/core/telemetry/index.ts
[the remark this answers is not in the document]
juliangruber: 2c5159e

documentation/events-and-metrics.md
juliangruber: +1 renamed affordance to `pin.filecoin.cloud` in d3fb5a3 e365b07

> What's the reason to reject any other values? What do we gain with the strictness?

It's to ensure that one doesn't submit with a typo, mixed case, etc.

> Also, will there be a PR in filecoin-pin-website after this?

Good catch! https://github.com/filecoin-project/filecoin-pin-website/issues/187

documentation/events-and-metrics.md
juliangruber: Agreed, good enough for now, and straight forward to improve later

README.md
juliangruber: Events and metrics seems more like a data type definition, while this section is relevnt for library users wanting to ensure all telemetry is submitted

README.md
juliangruber: No. I'm going to remove this configuratino, I don't see us needing it

juliangruber: b8b5b5d

documentation/events-and-metrics.md
juliangruber: ba271d6

documentation/events-and-metrics.md
juliangruber: ba271d6

documentation/events-and-metrics.md
juliangruber: Done in e6b545e — applied your suggested rewording to the Upload Event Model paragraph (with the link to the new glossary entry) and added the `Piece Copy` glossary entry. Thanks!

documentation/events-and-metrics.md
juliangruber: Done in 10b3390 — renamed `uploadCopySize` → `uploadCopyBytes` so the unit is self-documenting in the metric name (Prometheus/OTel `*Bytes` convention), updated the constant, emitter, tests, and this doc. Thanks!

BigLep (approved):
Looks good to me - thanks !

I pushed a couple of small doc changes with
https://github.com/filecoin-project/filecoin-pin/pull/501/commits/5b008ff9f6dcdccc8d3535ada819eed214d3192c
https://github.com/filecoin-project/filecoin-pin/pull/501/commits/b30b8c695eef3f93ff5ac734daf54bdcf7e048c4

PR #412, opened 2026-04-23
refactor(payments)!: harmonize runway/coverage with synapse-sdk 0.41.0

SgtPooki:
## Summary
- fix runway calculations to use total deposited Filecoin Pay balance rather than balance minus lockup
- align funding and depletion helpers with the same runway model
- add regression coverage for issue #385 and related edge cases

## Context
Issue #385 reported a mainnet deposit showing about 10.5225 USDFC deposited, 0.4670 USDFC/day spend, and `Runway: ~0 day(s)` even though the account had roughly 22 days of funds remaining.

This PR replaces the earlier Copilot attempt in #409 with a focused fix from `master`.

## Testing
- `pnpm install`
- `pnpm run typecheck`
- `pnpm test -- --run src/test/unit/payments.compute.test.ts`
- `pnpm exec vitest run src/test/unit/payments.compute.test.ts src/test/unit/payments-funding.test.ts src/test/unit/floor-pricing.test.ts`
- `pnpm biome check src/core/payments/index.ts src/core/payments/funding.ts src/core/payments/types.ts src/test/unit/payments.compute.test.ts src/test/unit/payments-funding.test.ts src/test/unit/floor-pricing.test.ts`

### manual run

```
> dist/cli.js payments deposit --network mainnet --private-key $WALLET_PRIVATE_KEY_REAL --amount 1  

┌  Filecoin Onchain Cloud Deposit
│
{"level":30,"time":1776967837649,"pid":63484,"hostname":"MacBookPro","event":"synapse.init","mode":"private-key","msg":"Initializing Synapse"}
{"level":30,"time":1776967837650,"pid":63484,"hostname":"MacBookPro","event":"synapse.init.success","chain":"Filecoin - Mainnet","msg":"Synapse initialized"}
◇  ✓ Connected
│
◇  ✓ Deposit complete
│
│  Transaction details:
│    Deposit: 0xb3d557cb685769ff609fecbb5b6e893452c8fc926f25c3e43fcb0ba006e46d7d
│
│
│  Deposit Summary
│    Total deposit: 2.4467 USDFC
│    Current spend: 0.0080 USDFC/day
│    Runway: ~10 month(s) 5 day(s) at current spend
│
└  Deposit completed

> dist/cli.js payments deposit --network calibration --private-key $WALLET_PRIVATE_KEY_REAL --amount 1

┌  Filecoin Onchain Cloud Deposit
│
[...]

PR #505, opened 2026-05-26
fix(payments): size auto-setup by available funds

SgtPooki:
## What changed

`payments setup --auto` sizes its default deposit from current available funds instead of the gross deposited balance.

A new pure helper, `computeAutoSetupTargetBalance` in `core/payments`, works out two values from live on-chain pricing: the funds that must be available (per data set: sybil fee, minimum monthly price, and the CDN lockup the default FilCDN upload path needs, plus 1 USDFC of runway) and the balance to deposit up to. `runAutoSetup` reads the account summary and deposits the shortfall between that requirement and current available funds.

## Why

The auto setup deposit decision compared the target against `status.filecoinPayBalance`, which is gross deposited funds and includes amounts locked by active rails. An account with funds already locked could report "deposit already sufficient" while leaving too little available to create new data sets, so the next upload could fail. Comparing against available funds (net of lockup and debt) keeps the available balance correct. Fresh accounts behave as before, where available equals gross.

## Tests

`computeAutoSetupTargetBalance`: fresh account, active lockup, and available funds already sufficient.

## Notes

Stacked on #412 (base `codex/fix-issue-385-runway`), which introduced the runtime auto-setup deposit sizing this builds on. Retarget to `master` after #412 merges.


src/payments/auto.ts
```diff
@@ -90,7 +90,7 @@ export async function runAutoSetup(options: PaymentSetupOptions): Promise<void>
     }
 
     // Now safe to get payment status since we know account exists
-    const status = await getPaymentStatus(synapse)
+    const [status, accountSummary] = await Promise.all([getPaymentStatus(synapse), synapse.payments.accountSummary({})])
```
juliangruber:
```suggestion
    const [status, accountSummary] = await Promise.all([getPaymentStatus(synapse), synapse.payments.accountSummary()])
```

juliangruber (approved):

PR #514, opened 2026-05-28
chore(ci): split build and publish jobs

SgtPooki:
## What changed

Splits `release-please.yml` into four jobs (`release-please`, `validate-dispatch`, `build`, `publish`) so the job that holds `id-token: write` for npm trusted publishing no longer also runs `pnpm install` and third-party build scripts. Closes the TanStack-class OIDC token extraction vector ([analysis](https://www.endorlabs.com/learn/how-a-misconfigured-ci-workflow-became-an-npm-supply-chain-compromise)) without changing the developer-facing release flow.

`build` runs install, build, and `npm pack`, then uploads a tarball artifact. It has no `id-token` and no `cache: pnpm` restore, so a compromised dependency cannot mint or scrape an npm publishing token and cannot rely on poisoned cache bytes. `publish` runs on a fresh runner with no project checkout or install, downloads the tarball, and runs `npm publish <tarball>` against npm's trusted publisher binding. Provenance attestation continues to attach automatically via OIDC.

Additional hardening: workflow-level `permissions: contents: read` with per-job grants, `persist-credentials: false` on checkout in jobs that execute dependency scripts, `concurrency` group to serialize releases, `npm@11` pin on publish-path runners, and a tag-ancestry check in `validate-dispatch` so manual dispatch can only target tags that point at commits already on `master`.

## How to verify

1. Merge.
2. `git log --oneline -3` on master shows this commit, which triggers a 0.22.1 release-please PR.
3. Merge the release PR. The four jobs run sequentially: `release-please` → `build` → `publish`. `validate-dispatch` is skipped on push.
4. `npm view filecoin-pin version` reports `0.22.1` with SLSA provenance attestation.
5. Manual recovery still works: Actions → release-please → Run workflow → pick a release tag. `validate-dispatch` runs (release-please is skipped on dispatch), then `build`, then `publish`. `dry_run=true` rehearses without publishing.

## Notes

- No npm trusted publisher reconfiguration required: workflow filename is unchanged and no environment is introduced.
- `environment: release` is intentionally not added in this PR. The job split already prevents the in-memory token extraction path; an approval gate would fire after `build` produces the tarball, which is too late to catch a poisoned artifact without reviewer-side tarball diffs.
- `--ignore-scripts` is not used.


SgtPooki (commented):
self review.

.github/workflows/release-please.yml
```diff
@@ -73,36 +95,102 @@ jobs:
...
+    needs: [release-please, validate-dispatch]
+    if: |
+      always() && (
+        needs.release-please.outputs.release_created == 'true' ||
+        needs.validate-dispatch.result == 'success'
+      )
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
```
rvagg, line 110:
```suggestion
    timeout-minutes: 20
    outputs:
```

.github/workflows/release-please.yml
```diff
@@ -73,36 +95,102 @@ jobs:
...
+          retention-days: 7
+          if-no-files-found: error
+
+  publish:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write  # OIDC for npm trusted publishing
+      contents: read
+    steps:
```
rvagg, line 161:
```suggestion
    timeout-minutes: 20
    steps:
```

rvagg (approved):

.github/workflows/release-please.yml
```diff
@@ -62,6 +73,17 @@ jobs:
             exit 1
           fi
 
+          # Confirm the tag commit is on master via GitHub's compare API
+          # (works regardless of local clone depth).
+          STATUS="$(gh api "repos/$GITHUB_REPOSITORY/compare/master...$SHA" --jq .status)"
```
juliangruber: nit: read default branch from `github.event.repository.default_branch`

juliangruber (approved):

PR #520, opened 2026-06-01
fix: keep node-only devnet config out of browser bundles

juliangruber:
## Summary

`src/common/get-rpc-url.ts` statically imports `node:fs`/`node:os`/`node:path` (added with devnet support). The browser only reaches it via the guarded lazy `import()` in `resolveChainFromRpc`, but Rollup/Vite still bundle lazy-import targets and choke on the node built-ins — breaking browser builds for downstream consumers.

This restores browser/Rollup support by isolating the node-only devnet code and letting bundlers swap it for a browser stub via the package `browser` field — the same mechanism the repo already uses for its other browser variants.

- New `src/common/devnet-config.ts` holds the `node:fs/os/path` imports, `resolveDevnetConfig`, cache, and `DevnetConfig`.
- New `src/common/devnet-config.browser.ts` stub throws; it's caught by the existing guard in `resolveChainFromRpc`.
- `package.json` `browser` field maps `./dist/common/devnet-config.js` → the stub.
- `get-rpc-url.ts` drops the node imports and re-exports `resolveDevnetConfig`, so node callers (`config.ts`, `cli-auth.ts`, `resolve-network.ts`) are unchanged.

Found via filecoin-project/filecoin-pin-website#188, which had to work around this with a Vite `resolveId` stub plugin. That workaround can be dropped once the site bumps to a release containing this change.

## Verification

- Bundled the browser-reachable path (`resolveChainFromRpc`) with esbuild targeting the browser: builds clean, pulls in no `node:fs/os/path`/`readFileSync`, resolves to the stub. The node target still uses the real devnet module.
- Added `devnet-config-browser-stub.test.ts` locking in the `browser` mapping + stub throw.
- `pnpm run lint`, `pnpm run typecheck`, and the unit project all pass.

## Test plan

- [x] CI green (lint, typecheck, unit, integration, browser)
- [x] Confirm a browser bundle of filecoin-pin no longer requires a devnet stub

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
   "packageManager": "pnpm@10.32.1",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "browser": {
+    "./dist/common/devnet-config.js": "./dist/common/devnet-config.browser.js"
+  },
   "exports": {
     ".": {
       "browser": {
--- a/src/common/devnet-config.browser.ts
+++ b/src/common/devnet-config.browser.ts
@@ -1,0 +1,20 @@
+/**
+ * Browser stub for ./devnet-config.js.
+ *
+ * Devnet support reads devnet-info.json from disk via node:fs/os/path, which has
+ * no meaning in the browser. Bundlers resolve to this file via the "browser" field
+ * in package.json, keeping node built-ins out of browser bundles. The only browser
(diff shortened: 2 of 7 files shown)


rvagg (approved):
oh snap, good catch

SgtPooki (commented):
we could do what libp2p did with devtools browser plugin for devnet support if we ever get time to do that.

Issue #522, opened 2026-06-01
GA: CLI cleanup and consistency


SgtPooki:
Part of #279. GA v1 CLI cleanup. Each item is non-breaking unless tagged.

## Tasks

- [x] Return a distinct non-zero exit code for unconfirmed or cancelled operations.
      `rm --wait-for-confirmation` on timeout (`src/core/piece/remove-piece.ts:111`) and
      `data-set terminate` on cancel (`src/data-set/run.ts:233`) return normally, so the postAction
      hook exits `0` (`src/cli.ts:71`). Use exit code `2`, distinct from the `1` used for caught
      errors. **Done when:** both paths exit `2`, and tests assert `2` instead of `0`.
- [x] Wire `-v/--verbose` (`src/cli.ts:15`, declared but unread) to set the log level to `debug` for
      the run. **Done when:** running any command with `-v` emits debug logs; the flag is no longer
      dead.
- [x] Suppress the update-check banner (`src/cli.ts:61`) when stdout is not a TTY, using the
      `isTTY()` helper (`src/utils/cli-logger.ts:16`). **Done when:** the banner prints on a TTY and
      not when stdout is piped.
- [x] Surface auth env var names in `--help`. Move `--private-key`, `--wallet-address`,
      `--session-key`, `--provider-ids`, and `--data-set-ids` to `new Option().env(...)`, and have
      `server` use the shared auth-options builder. **Done when:** `--help` (including `server --help`)
      shows the `[env: ...]` annotation for each of these flags.
- [x] Rename `rm` to `remove`, keeping `rm` as an alias. **Done when:** `remove` is the primary
      command in help and `rm` still works.
- [x] Rename `--8004-type` / `--8004-agent` to `--erc8004-type` / `--erc8004-agent`, keeping the old
      names as hidden aliases (`src/utils/cli-options-metadata.ts`). **Done when:** the new flags
      work, the old ones still parse, and help shows only the new names.
- [x] Add a read command for piece status: `data-set piece-status <dataSetId> [pieceCid]` (name
      open to change). Export `reconcilePieceStatus` from `core/piece` (`src/core/piece/piece-status.ts:38`,
      currently unexported); there is no piece-to-dataset reverse lookup, so the command scopes to a
      dataset. **Done when:** the command prints the status of a dataset's pieces (or one piece), and
      `reconcilePieceStatus` is exported.

## Breaking changes (included for GA)

[...]

Issue #279, opened 2025-12-04
Review/update API for GA


BigLep:
DOH - I meant to put this in Synapse.  Just a sec....

## Done Criteria
The public API has been updated to smooth out rough edges. align names/semantics, etc.  The specific changes will be added as part of X.

## Why Important
The SDK *works now*, but it’s unclear if the **external API shape** is what we want to commit to for GA.  We need to take an **architectural lens** on the SDK/Synapse APIs, not just individual issues.

## User/Customer
Consumers
Contributors so they have a consistent pattern to follow and reason about

## Notes
We have likely gotten into this state because in the sprint to Buenos Aires, tasks oftengot done in isolation, and things “slip through” in a absence of a clear API owner guarding consistency.

SgtPooki:
I had claude and a swarm of agents do a quick pass:

https://bafkreibmcynlo2d7ghmsad3zqv3bdmpwkhrzja2lqr6mja5die46cgjhqq.ipfs.inbrowser.link/

BigLep:
I zero'd out the estimate since the subtasks have estimates.

BigLep:
I'll close this since the API updates have been handled that we wanted for GA. We just have cost related things, but those are tracked separately in
https://github.com/filecoin-project/filecoin-pin/issues/559
https://github.com/filecoin-project/filecoin-pin/issues/402

PR #526, opened 2026-06-01
feat(payments)!: make deposit one-way, fund sets targets

SgtPooki:
Part of #522 (CLI cleanup for GA), under #279.

## What changed

`payments deposit --days N` duplicated `payments fund --days N --mode minimum`, and `fund` could withdraw despite its additive-sounding name. This splits the two commands by direction of effect:

- `payments deposit` now requires `--amount` and only adds funds. `--days` is removed.
- `payments fund` keeps the exact-target behavior (deposits or withdraws to hit a runway or total), with a clearer description.

## How to verify

- `filecoin-pin payments deposit --help` shows `--amount` as required and no `--days`.
- `filecoin-pin payments fund --help` describes setting an exact target.
- `pnpm test` (payments unit tests pass; typecheck clean).

## Notes / risks

BREAKING (CLI): `payments deposit --days` is removed. The runway top-up use case is `payments fund --days N --mode minimum`. The CLI is invoked via npx / global install, so this lands pre-GA per the #279 discussion.


SgtPooki (commented):
self review

src/payments/deposit.ts
```diff
@@ -1,22 +1,15 @@
...
+ * use `payments fund`.
  */
 
 import pc from 'picocolors'
 import { parseUnits } from 'viem'
 import { CliFatal, isCliFatal } from '../common/cli-errors.js'
-import {
-  checkFILBalance,
-  checkUSDFCBalance,
-  computeTopUpForDuration,
```
juliangruber, line 15: `computeTopUpForDuration` looks to be unused now

juliangruber (approved):

rvagg (approved):
heh yeah .. probably a good idea

PR #530, opened 2026-06-02
feat: upload-action docs, CLI recipe, and input rename

### LINKED ISSUE #524: GA: Upload action doc fixes
Part of #279. GA v1 upload action documentation fixes. The action is now the authoritative GitHub Action we support for uploading to Filecoin (IP Shipyard is not taking on this work in ipfs-deploy-action; see ipshipyard/ipfs-deploy-action#39). These are doc-drift fixes and framing updates with no behavior change.

## Tasks

- [ ] Use one wallet secret name across docs. The READMEs use `secrets.FILECOIN_WALLET_KEY`; the example workflows use `secrets.WALLET_PRIVATE_KEY`. The action input (`walletPrivateKey`) is consistent. **Done when:** READMEs and example workflows use the same secret name.
- [ ] Fix the provider override docs. The code parses only `PROVIDER_IDS` (`upload-action/src/inputs.js:150`); the README documents `PROVIDER_ID` and a `PROVIDER_ADDRESS` "highest priority" option that is not implemented. **Done when:** the README documents only `PROVIDER_IDS` and the unimplemented priority section is removed.
- [ ] Resolve the network mismatch: the README says the action runs on Calibration, `action.yml` defaults `network` to `mainnet`. **Done when:** the README and `action.yml` agree on the default.
- [ ] Correct the `uploadStatus` enum docs. `action.yml:78` lists `uploaded`, `reused-cache`, `reused-artifact`; the code emits `uploaded`, `dry-run`, `fork-pr-blocked`, and `pending-upload`. **Done when:** the docs list the emitted values and drop `reused-cache` / `reused-artifact`.
- [ ] Document the egress default difference: the action defaults `egressProvider` to `none`, the CLI defaults `--egress-provider` to `beam`. Defaults stay as they are. **Done when:** the action README and CLI help both note the difference.
- [ ] Remove "educational demo" framing and ipfs-deploy-action deferral from upload-action docs. The upload-action README describes itself as "an educational demo" not expected to be depended on for production use, and defers to `ipshipyard/ipfs-deploy-action` via [issue #39](https://github.com/ipshipyard/ipfs-deploy-action/issues/39). IP Shipyard has decided not to take on this work and is instead pursuing a recipe-based approach ([PR #53](https://github.com/ipshipyard/ipfs-deploy-action/pull/53)). We are treating the upload-action as the authoritative GitHub Action we support for uploading to Filecoin. **Done when:** the upload-action README opening paragraph positions it as the supported Filecoin upload action (no "educational demo", no deferral to ipfs-deploy-action); the Calibration testnet note is updated (default is now mainnet); the main README "Examples" entry at line 84 treats it as a supported tool rather than "Example GitHub Action workflows demonstrating automated IPFS/Filecoin uploads".
- [ ] Add a CLI-based recipe for users who prefer a custom workflow over the composite action. IP Shipyard's [PR #53](https://github.com/ipshipyard/ipfs-deploy-action/pull/53) advocates a pattern where `ipfs-deploy-action` produces a CAR and downstream steps consume it via CLI. We should document an equivalent pattern in our repo: a recipe showing how to install and run the `filecoin-pin` CLI directly in a GitHub Actions workflow (e.g. `npx filecoin-pin add`) without using the composite action. This gives users full control over their workflow while still using our tooling, and is the pattern we'd point ipfs-deploy-action users to. **Done when:** a recipe (in the upload-action examples dir or a new docs location) shows a complete workflow that installs `filecoin-pin` from npm and runs `filecoin-pin add` / `filecoin-pin import`, covering both "pack + upload a directory" and "upload a pre-built CAR" cases.

## Breaking change (included for GA)

- [ ] Rename the financial inputs to match the CLI vocabulary: `minStorageDays` to `minRunwayDays` and `filecoinPayBalanceLimit` to `maxBalance` (the CLI uses `--min-runway-days` / `--max-balance`). **Done when:** the action accepts the new input names; the old names work as deprecated aliases or are removed; the README and examples use the new names.

beck-8:
Closes https://github.com/filecoin-project/filecoin-pin/issues/524

GA upload-action cleanup, in four commits:

1. **Fix doc drift** — one wallet secret name (`FILECOIN_WALLET_KEY`) across the example workflows; document only the implemented `PROVIDER_IDS` override and drop the never-implemented `PROVIDER_ADDRESS` / `PROVIDER_ID`; state the mainnet default for `network` to match action.yml; correct the `uploadStatus` enum to the emitted values (`uploaded`, `dry-run`, `fork-pr-blocked`, `pending-upload`); note in the action README that the action defaults egress to `none` while the CLI defaults to `beam`.
2. **Position as the supported upload action** — drop the "educational demo" framing and the ipfs-deploy-action deferral; update the root README Examples entry.
3. **Add a CLI recipe** (`examples/cli-recipe/`) — workflow + README showing how to run the filecoin-pin CLI directly (`filecoin-pin add` / `import`) without the composite action.
4. **`feat!`: rename financial inputs** — `minStorageDays` -> `minRunwayDays`, `filecoinPayBalanceLimit` -> `maxBalance` (matching the CLI's `--min-runway-days` / `--max-balance`). Old names remain as deprecated aliases (with a deprecation warning); new names take precedence. All first-party usages (README, examples, FLOW.md, and the test-upload-action CI workflow) are migrated to the new names.

> [!WARNING]
> **BREAKING CHANGE (commit 4):** the inputs `minStorageDays` and `filecoinPayBalanceLimit` are renamed to `minRunwayDays` and `maxBalance`. The old names keep working as deprecated aliases for now and will be removed in a future release.

**Beyond the issue checklist:** FLOW.md is brought in sync (new input names + alias note) and drops a stale reference to a nonexistent `providerAddress` setting, documenting the real `PROVIDER_IDS` env override instead.

**Notes for maintainers:** `secrets.WALLET_PRIVATE_KEY` in `test-upload-action.yml` is left as-is (it maps to a real configured repo secret; renaming it to `FILECOIN_WALLET_KEY` would require renaming the repo secret too) — left as a follow-up. The egress default difference is documented in the action README only, not in the CLI `--egress-provider` help, to keep CLI help free of GitHub-Action-specific concerns. No `src/` changes; CLI behavior is unchanged.

### DIFF
--- a/upload-action/README.md
+++ b/upload-action/README.md
@@ -16,7 +16,7 @@ See the two-workflow approach in the [examples directory](./examples/) for compl
 
 See [action.yml](./action.yml) for complete input documentation including:
 - **Core**: `path`, `walletPrivateKey`, `network`
-- **Financial**: `minRunwayDays`, `maxBalance` (the old names `minStorageDays` / `filecoinPayBalanceLimit` still work as deprecated aliases)
+- **Financial**: `minRunwayDays`, `maxBalance`
 - **Advanced**: `egressProvider`, `dryRun`, `disableTelemetry`
 
 **Outputs**: `ipfsRootCid`, `dataSetId`, `pieceCid`, `providerId`, `providerName`, `carPath`, `uploadStatus`
--- a/upload-action/README.md
+++ b/upload-action/README.md
@@ -95,6 +95,7 @@ The action checks npm for a newer `filecoin-pin` release at the start of each ru
 ## Examples & Documentation
 
 - **[examples/](./examples/)** - Ready-to-use workflow files and setup instructions
+- **[examples/cli-recipe/](./examples/cli-recipe/)** - Drive the `filecoin-pin` CLI directly from your own workflow steps, without this composite action
 - **[Actual usage in filecoin-pin-website repo](https://github.com/filecoin-project/filecoin-pin-website/blob/main/.github/workflows/filecoin-pin-upload.yml)** ([🎥 demo recording](https://www.youtube.com/watch?v=_2ZsMYXfgwI))
 - **[Filecoin Pin + ENS Demo](https://github.com/FIL-Builders/filecoin-pin-ens-demo)** ([🎥 demo recording](https://www.youtube.com/watch?v=tkDwXAVtnDA)) - A minimal demo showing a static website deployed with the Filecoin Pin Upload Action and an ENS update that points the ENS name to the latest IPFS CID after each push to main.
 - **[FLOW.md](./FLOW.md)** - Internal architecture for contributors and maintainers
--- a/upload-action/README.md
+++ b/upload-action/README.md
@@ -2,7 +2,7 @@
 
 The Filecoin Pin Upload Action is a composite GitHub Action that packs a file or directory into a UnixFS CAR, uploads it to Filecoin, and publishes artifacts and context for easy reuse.
 
-This GitHub Action is provided to illustrate how to use filecoin-pin, a new IPFS pinning workflow that stores to the Filecoin decentralized storage network.  It's not expected to be the action that other repos will depend on for their production use case of uploading to Filecoin.  Given the emphasis on this being an educational demo, breaking changes may be made at any time.  For robust use, the intent is to add filecoin-pin functionality to the ipshipyard/ipfs-deploy-action, which is being tracked in [issue #39](https://github.com/ipfs/ipfs-deploy-action/issues/39).
+This is the GitHub Action that Filecoin Pin supports for uploading content to the Filecoin decentralized storage network from CI. It packs a file or directory into a UnixFS CAR and uploads it via filecoin-pin, so you can publish to Filecoin on every push or pull request.
 
 *Note: The action defaults to Filecoin Mainnet (`network: mainnet`). Set `network: calibration` to target the Calibration testnet, where data isn't permanent and infrastructure resets regularly.*
 
--- a/upload-action/README.md
+++ b/upload-action/README.md
@@ -12,11 +12,11 @@ See the two-workflow approach in the [examples directory](./examples/) for compl
 
 ## Inputs & Outputs
 
-> **Using the CLI directly?** `filecoin-pin add` and `filecoin-pin import` accept `--auto-fund`, `--min-runway-days`, and `--max-balance` flags that mirror `minStorageDays` and `filecoinPayBalanceLimit` here. See `filecoin-pin add --help`.
+> **Using the CLI directly?** `filecoin-pin add` and `filecoin-pin import` accept `--auto-fund`, `--min-runway-days`, and `--max-balance` flags that mirror `minRunwayDays` and `maxBalance` here. See `filecoin-pin add --help`.
 
 See [action.yml](./action.yml) for complete input documentation including:
 - **Core**: `path`, `walletPrivateKey`, `network`
-- **Financial**: `minStorageDays`, `filecoinPayBalanceLimit`
+- **Financial**: `minRunwayDays`, `maxBalance` (the old names `minStorageDays` / `filecoinPayBalanceLimit` still work as deprecated aliases)
 - **Advanced**: `egressProvider`, `dryRun`, `disableTelemetry`
 
 **Outputs**: `ipfsRootCid`, `dataSetId`, `pieceCid`, `providerId`, `providerName`, `carPath`, `uploadStatus`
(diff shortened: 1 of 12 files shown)


upload-action/src/inputs.js
```diff
@@ -66,6 +66,30 @@ export function getInput(name, fallback = '') {
...
+  const current = getInput(name, '')
+  if (current !== '') return current
+
+  const deprecated = getInput(deprecatedName, '')
+  if (deprecated !== '') {
+    console.warn(
+      `::warning::Input "${deprecatedName}" is deprecated; use "${name}" instead. ` +
+        `Support for "${deprecatedName}" will be removed in a future release.`
+    )
+    return deprecated
```
juliangruber:
```suggestion
    return toStringValue(deprecated).trim()
```

juliangruber (approved):

BigLep:
@beck-8 : I'll take a look too, particularly around
- Remove "educational demo" framing and ipfs-deploy-action deferral from upload-action docs
- Add a CLI-based recipe for users who prefer a custom workflow over the composite action

upload-action/examples/cli-recipe/upload-with-cli.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Import CAR to Filecoin
+        env:
+          PRIVATE_KEY: ${{ secrets.FILECOIN_WALLET_KEY }}
+        run: |
+          npx -y filecoin-pin@0.22.3 import build.car \
+            --network mainnet \
+            --auto-fund \
+            --min-runway-days 30 \
+            --max-balance 5.00
```
SgtPooki: do we need users to build a car themselves, or can we just use `filecoin-pin add /build-folder` ?

SgtPooki (changes requested):
in general lgtm but it would be nice for the recipe to just be a drop-in that will work for most users.

upload-action/examples/cli-recipe/upload-with-cli.yml
beck-8: I put it in the readme so that the default ci will not fail

upload-action/examples/cli-recipe/README.md
```diff
@@ -0,0 +1,58 @@
...
+    npx -y filecoin-pin@0.22.3 import build.car \
+      --network mainnet \
+      --auto-fund \
+      --min-runway-days 30 \
+      --max-balance 5.00
+```
+
+## Composing with a CAR-producing action
+
+If you produce the CAR with [`ipfs-deploy-action`](https://github.com/ipshipyard/ipfs-deploy-action) (or any step that emits one), pass that CAR's path to `filecoin-pin import`. That project also maintains a [`filecoin-pin` recipe](https://github.com/ipshipyard/ipfs-deploy-action/tree/main/docs/recipes) for this exact pattern.
```
BigLep: I think this could be merged into one section rather than being two. At the minimum, `Composing with a CAR-producing action` should be a subheading `###`.

upload-action/examples/cli-recipe/README.md
BigLep, line 1: There's a decent amount of duplication with the upload-with-cli.yml file. Do we just put all this extra readme content as comments within "upload-with-cli.yml"?

upload-action/action.yml
```diff
@@ -23,27 +23,37 @@ inputs:
...
+      Specifies a target Filecoin Pay balance in USDFC that this action will never exceed.
       Actions (manual or other) outside of this action may cause the Filecoin Pay balance to exceed this limit,
       but this limit prevents the action from using the wallet to top up the Filecoin Pay balance beyond the limit.
-      As a protection, this must be set if `minStorageDays` is set.
+      As a protection, this must be set if `minRunwayDays` is set.
+      Mirrors the CLI's `--max-balance`.
       SECURITY: Hardcode this in trusted workflows.
     required: false
 
+  # Deprecated aliases (kept for backwards compatibility; will be removed in a future release)
```
BigLep: Any reason to not just remove them now? I am ok with breakages here, but if you think we should allow some transition period, that is fine.
Maybe add a comment that these can get removed whenever. We just wanted to have at least one release to provide transition.
But again, I'm also good just doing the breakage now.

upload-action/README.md
```diff
@@ -2,25 +2,27 @@
...
 
 ## Inputs & Outputs
 
-> **Using the CLI directly?** `filecoin-pin add` and `filecoin-pin import` accept `--auto-fund`, `--min-runway-days`, and `--max-balance` flags that mirror `minStorageDays` and `filecoinPayBalanceLimit` here. See `filecoin-pin add --help`.
+> **Using the CLI directly?** `filecoin-pin add` and `filecoin-pin import` accept `--auto-fund`, `--min-runway-days`, and `--max-balance` flags that mirror `minRunwayDays` and `maxBalance` here. See `filecoin-pin add --help`.
 
 See [action.yml](./action.yml) for complete input documentation including:
 - **Core**: `path`, `walletPrivateKey`, `network`
-- **Financial**: `minStorageDays`, `filecoinPayBalanceLimit`
+- **Financial**: `minRunwayDays`, `maxBalance` (the old names `minStorageDays` / `filecoinPayBalanceLimit` still work as deprecated aliases)
```
BigLep: I don't know if it's critical to list the old names. I don't really want to leave more stuff for us to clean up later. Ideally we clean now why we're thinking about it.

BigLep (approved):
Thanks @beck-8 . I'm fine to look again but please don't block on me.

upload-action/action.yml
beck-8: There seems to be no problem with upload-action

SgtPooki (approved):

PR #531, opened 2026-06-02
feat(server): GA cleanup for pinning server (beta)

### LINKED ISSUE #523: GA: Pinning server cleanup (beta)
Part of #279. GA v1 pinning server cleanup. The server ships as beta; state loss across restarts is
accepted, so this is cleanup only (no persistence work).

## Tasks

- [x] Refuse startup when no access token is configured, unless an explicit `--allow-no-auth` flag is
      passed (name open to change). Today the server runs open to all requests with no token
      (`src/filecoin-pinning-server.ts:150`). Update the unit and integration test fixtures, which
      start with no `accessToken` (the integration tests already send `Bearer test-token`, so they can
      set a real token). **Done when:** startup fails with a clear error when no token and no flag;
      starts with the flag; tests updated. Small but breaking.
- [x] Use the IPFS Pinning Service error shape `{ error: { reason, details } }` in place of the
      current `{ error: string }`. Spec: https://ipfs.github.io/pinning-services-api-spec/. **Done
      when:** all error responses use the spec shape, and a test asserts it for 400 / 401 / 404 / 500.
- [x] Add request validation and remove the `any`-typed handlers for `POST` and
      `DELETE /pins/:requestId`. **Done when:** both handlers are typed, a malformed body returns
      `400`, and no `any` remains in those handlers.
- [x] Document that `delegates` is always empty. Each pin's Helia node is created and stopped within
      one pin operation (`src/filecoin-pin-store.ts:211`, stopped in `finally` near `:402`), so there
      is no long-lived node to advertise. **Done when:** the server docs state this and why.

juliangruber:
Resolves #523 (part of #279). GA v1 pinning server cleanup. The server ships as beta; state loss across restarts is accepted, so this is cleanup only (no persistence work).

### Refuse startup without an access token (breaking)
The server previously ran open to all requests when no token was configured. It now refuses to start unless an access token is set via `--access-token` / `ACCESS_TOKEN`, or the operator explicitly opts out with `--allow-no-auth` / `ALLOW_NO_AUTH`. The check runs after synapse-config validation but before the network call, with a friendly stderr hint on failure.

### IPFS Pinning Service spec error shape
All error responses now use `{ error: { reason, details } }` (per the [spec](https://ipfs.github.io/pinning-services-api-spec/)) instead of `{ error: string }`, via a shared `sendError` helper. Covers 400 / 401 / 403 / 404 / 500.

### Typed + validated `POST` and `DELETE /pins/:requestId`
Removed the `any`-typed handlers and added `FastifyRequest` generics. A shared `parsePinMutationBody` validates `name` / `origins` / `meta` (also applied to `POST /pins`); malformed bodies return `400`. Added auth-context (`401`) guards.

### Docs
README pinning-server section now notes the beta status (in-memory, state lost across restarts), the access-token requirement + `--allow-no-auth`, and that `delegates` is always `[]` (each pin uses a short-lived Helia node, so there's no long-lived node to advertise). Added `ACCESS_TOKEN` / `ALLOW_NO_AUTH` to the env-var list.

## Tests
- New unit tests for the startup guard (no token → throws; `--allow-no-auth` → starts; token → starts).
- New integration tests asserting the spec error shape for 400 / 401 / 404 / 500 (500 via a `pinStore.get` spy) and malformed update bodies.
- Updated existing fixtures: integration config sets `accessToken: 'test-token'`; unit session-key success case sets a token.
- Full suite green: 524 unit + 13 integration, lint + typecheck pass.

## Reviewer notes
- **Breaking**: existing deployments relying on token-less startup must now pass `--allow-no-auth` or configure a token.
- `--allow-no-auth` flag name is open to change (per the issue).

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,9 +64,11 @@ Use Filecoin Pin programmatically in your Node.js or browser applications. The l
 ### 📡 IPFS Pinning Server (Daemon Mode)
 Run a localhost IPFS Pinning Service API server that implements the [IPFS Pinning Service API specification](https://ipfs.github.io/pinning-services-api-spec/). This allows you to use standard IPFS tooling (like `ipfs pin remote`) while storing data on Filecoin.
 
-- **Status**: Works and is tested, but hasn't received as many features as the CLI. If it would benefit your use case, please comment on the [tracking issue](https://github.com/filecoin-project/filecoin-pin/issues/46) so we can be better informed when it comes to prioritizing.
+- **Status**: Beta. Works and is tested, but hasn't received as many features as the CLI. State is held in memory and is lost across restarts. If it would benefit your use case, please comment on the [tracking issue](https://github.com/filecoin-project/filecoin-pin/issues/46) so we can be better informed when it comes to prioritizing.
 - **Repository**: This repo (`filecoin-pin server` command in CLI)
-- **Usage**: `PRIVATE_KEY=0x... npx filecoin-pin server` (or use session key auth — see [Configuration](#configuration))
+- **Usage**: `PRIVATE_KEY=0x... ACCESS_TOKEN=... npx filecoin-pin server` (or use session key auth — see [Configuration](#configuration))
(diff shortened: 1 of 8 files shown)


SgtPooki (approved):
good enough for now

PR #533, opened 2026-06-03
fix(cli): distinct exit code 2 for cancelled or unconfirmed operations

juliangruber:
## Summary

`rm --wait-for-confirmation` on a confirmation timeout and `data-set terminate` on a declined confirmation both returned normally, so the `postAction` hook exited `0`. Scripts had no way to distinguish "the operation didn't complete" from success.

Both paths now set exit code `2` (`EXIT_CODE_INCOMPLETE`), distinct from the `1` the Commander wrappers use for caught errors.

## Changes

- Add `EXIT_CODE_INCOMPLETE = 2` in `src/common/cli-errors.ts`.
- `data-set terminate`: set exit code `2` when the user declines/cancels the confirmation prompt.
- `rm`: set exit code `2` when `--wait-for-confirmation` was requested but the wait timed out (removal unconfirmed).

## Test plan

- [x] Unit test: `rm` exits `2` when a requested confirmation wait times out, and stays `0` when confirmation wasn't requested.
- [x] Unit test: `data-set terminate` exits `2` when the user cancels the confirmation prompt.
- [x] `tsc --noEmit` and Biome pass on changed files.

### DIFF
--- a/src/rm/remove-piece.ts
+++ b/src/rm/remove-piece.ts
@@ -114,7 +114,7 @@ export async function runRmPiece(options: RmPieceOptions): Promise<RmPieceResult
       waitForConfirmation: options.waitForConfirmation ?? false,
     })
 
-    // A requested confirmation wait that timed out leaves the removal
+    // Time-out waiting for requested confirmation, leaving the removal
     // unconfirmed. Signal that distinctly so scripts can tell it apart from
     // both success (0) and a caught error (1).
     if (options.waitForConfirmation === true && !isConfirmed) {
--- a/src/rm/remove-piece.ts
+++ b/src/rm/remove-piece.ts
@@ -9,6 +9,7 @@
  */
 import pc from 'picocolors'
 import pino from 'pino'
+import { EXIT_CODE_INCOMPLETE } from '../common/cli-errors.js'
 import { type RemovePieceProgressEvents, removePiece } from '../core/piece/index.js'
 import { initializeSynapse } from '../core/synapse/index.js'
 import { parseCLIAuth } from '../utils/cli-auth.js'
@@ -112,6 +113,13 @@ export async function runRmPiece(options: RmPieceOptions): Promise<RmPieceResult
       onProgress,
       waitForConfirmation: options.waitForConfirmation ?? false,
     })
+
+    // A requested confirmation wait that timed out leaves the removal
+    // unconfirmed. Signal that distinctly so scripts can tell it apart from
+    // both success (0) and a caught error (1).
+    if (options.waitForConfirmation === true && !isConfirmed) {
+      process.exitCode = EXIT_CODE_INCOMPLETE
+    }
 
     // Ensure spinner is stopped before displaying results
     spinner.stop(`${pc.green('✓')} Piece removed${isConfirmed ? ' and confirmed' : ' (confirmation pending)'}`)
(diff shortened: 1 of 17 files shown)


src/rm/remove-piece.ts
```diff
@@ -113,6 +114,13 @@ export async function runRmPiece(options: RmPieceOptions): Promise<RmPieceResult
       waitForConfirmation: options.waitForConfirmation ?? false,
     })
 
+    // A requested confirmation wait that timed out leaves the removal
```
juliangruber:
```suggestion
    // Time-out waiting for requested confirmation, leaving the removal
```

beck-8 (changes requested):
It seems we have other things to handle together
 
 | Path | Mechanism | Exit code |
  |---|---|---|
  | `data-set terminate` — confirmation declined | `exitCode = 2; return` (this PR) | **2** ✅ |
  | `rm --piece --wait-for-confirmation` — wait times out | `exitCode = 2` (this PR) | **2** ✅ |
  | `rm --all` — **confirmation declined** | `throw` → wrapper `catch { exit 1 }` | **1** ❌ |
  | `rm --all --wait-for-confirmation` — wait times out | not handled, returns normally | **0** ❌ |
  | `session authorize` — confirmation declined | `throw new Error('Authorization cancelled')` | **1** ❌ |
  | `payments setup` (interactive) — cancelled | `throw new CliFatal('Setup cancelled')` | **1** ❌ |

SgtPooki:
I ran a deep review pass on this and pushed the fixes directly (see the last commit) rather than leaving a comment trail. What it covers:

**Bug worth knowing about:** the `payments fund` deposit/withdraw prompts checked `if (!proceed)` without `isCancel`. clack's cancel symbol is truthy, so Ctrl-C/ESC at "Deposit X USDFC?" skipped the throw and proceeded with the transaction. Every other confirm site in the repo guards with `isCancel`; fund.ts was the outlier. Fixed and covered by tests.

**Completeness:** the three `fund.ts` confirm declines still threw plain `Error`, so the wrapper exited 1, contradicting the CONTRIBUTING table this PR adds. They now exit 2 via a new `CliIncomplete` error that the runner's catch maps. Also: `remove --all` now exits 1 when any piece fails (failures previously only changed console text and exited 0, including the all-failed case where `confirmedCount < removedCount` was `0 < 0`), and `data-set terminate --wait` maps a receipt timeout to exit 2 to match `remove --wait`.

**Mechanism:** the `(process.exitCode ?? 0) === 0` guard was hand-rolled at 8 sites, which is exactly how the downgrade bug Copilot kept finding got in. It's now `setIncompleteExitCode()` in cli-errors.ts.

**Output:** the `remove` outros said "completed successfully" while exiting 2; they now report "confirmation still pending".

One API note for the release: `RemoveAllPiecesResult` (exported via `./core/piece`) gains a required `confirmedCount` field. Readers are unaffected; TS consumers constructing the type need the new field.

The two latest Copilot comments about unconditional exit-code sets were already fixed in d7b6416 and can be dismissed.

SgtPooki (approved):

BigLep:
@juliangruber : I assume you'll merge after looking at https://github.com/filecoin-project/filecoin-pin/pull/533#issuecomment-4635219498 and https://github.com/filecoin-project/filecoin-pin/pull/533/commits/f1f9d81db16c3c3d8a7d2ef761bf9a5e8b391f1c

beck-8 (approved):

PR #535, opened 2026-06-03
fix(cli): suppress update banner when stdout is not a TTY

juliangruber:
## Summary

The "update available" banner printed unconditionally from the `postAction` hook, which pollutes piped/CI output (e.g. `filecoin-pin ... | jq`). It now prints only on an interactive TTY.

## Changes

- Add `printUpdateBanner(result, { tty, print })` in `src/common/version-check.ts`. It gates on the `isTTY()` helper from `src/utils/cli-logger.ts` and is no-op for non-`update-available` statuses. `tty`/`print` are injectable for testing.
- `src/cli.ts` postAction now calls `printUpdateBanner` instead of building and logging the banner inline (removes the now-unused `picocolors` import).

## Test plan

- [x] Unit tests: banner prints on TTY, is suppressed when not a TTY, and is a no-op when no update is available.
- [x] `node dist/cli.js --help` still loads.
- [x] `tsc --noEmit` and Biome pass.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,10 +1,9 @@
 #!/usr/bin/env node
 import './instrument.js'
 import { Command } from 'commander'
-import pc from 'picocolors'
 
 import { ALL_CLI_COMMANDS } from './commands/index.js'
-import { checkForUpdate, type UpdateCheckStatus } from './common/version-check.js'
+import { checkForUpdate, printUpdateBanner, type UpdateCheckStatus } from './common/version-check.js'
 import { configureTelemetry, flushTelemetry } from './core/telemetry/index.js'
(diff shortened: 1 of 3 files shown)


src/cli.ts
beck-8, line 40: 我们要不要在这里直接让他不发起请求？
```
if (options.updateCheck === false || !isTTY()) {
```

beck-8: If they really need it, do they want non-TTY output to sterr?

beck-8 (approved):
There is one question, no other questions

SgtPooki (approved):

src/cli.ts
juliangruber: not sure I understand 🤔

PR #536, opened 2026-06-03
feat(cli): surface auth env var names in --help

juliangruber:
## Summary

The auth/context flags hinted their env vars in free-text descriptions (or not at all). Declaring them via `new Option().env(...)` makes Commander render a real `[env: ...]` annotation in `--help`.

Affected flags: `--private-key`, `--wallet-address`, `--session-key`, `--provider-ids`, `--data-set-ids`.

## Changes

- Extract `addSigningAuthOptions()` (the three signing flags) so it can be shared. `addAuthOptions()` builds on it (adds `--view-address`, `--network`, `--rpc-url`).
- `--provider-ids` / `--data-set-ids` now use `.env('PROVIDER_IDS')` / `.env('DATA_SET_IDS')`.
- The `server` command uses `addSigningAuthOptions` instead of duplicating the three flags, so `server --help` shows the annotations. It omits `--view-address` because the server has no view-only auth path (`buildSynapseConfig` supports only private-key or wallet+session) — adding it would be a dead flag.

Behavior is unchanged: the runner-side `|| process.env.X` fallbacks remain, and Commander's `.env()` keeps the same precedence (CLI flag wins over env).

## Test plan

- [x] Unit tests assert the `envVar` bindings and that `--help` shows the env names.
- [x] `node dist/cli.js server --help` and `add --help` show `(env: ...)` for the flags; server has no `--view-address`.
- [x] Full unit suite passes (pre-existing `upload-action.test.ts` env-resolution failures are unrelated).
- [x] `tsc --noEmit` and Biome pass.


src/utils/cli-options.ts
```diff
@@ -70,16 +82,14 @@ export function addAuthOptions(command: Command): Command {
 export function addContextSelectionOptions(command: Command): Command {
   return command
     .addOption(
-      new Option(
-        '--provider-ids <ids>',
-        'Target specific providers by ID, comma-separated (can also use PROVIDER_IDS env)'
-      ).conflicts('dataSetIds')
+      new Option('--provider-ids <ids>', 'Target specific providers by ID, comma-separated')
```
beck-8: I know this has nothing to do with your modification, but would it be better if we delete the `.conflicts` here?
Will it produce such an effect and make users have some doubts? At the same time, `parseContextSelectionOptions` will become dead code
```
  PROVIDER_IDS=1 DATA_SET_IDS=2  →  THREW:
    "environment variable 'PROVIDER_IDS' cannot be used with environment variable 'DATA_SET_IDS'"
```

beck-8 (dismissed):
There is an omission in the `addOwnerAuthOptions` method of `session create`

src/utils/cli-options.ts
juliangruber: https://github.com/filecoin-project/filecoin-pin/pull/540

beck-8:
and `session create --session-key`

SgtPooki (approved):

PR #539, opened 2026-06-03
feat(cli): add data-set piece-status command, export reconcilePieceStatus

juliangruber:
## Summary
- Adds a read-only `data-set piece-status <dataSetId> [pieceCid]` command that prints the reconciled status of a data set's pieces.
- When a `pieceCid` is supplied, the output is filtered to the matching piece (by PieceCID or IPFS root CID); otherwise every piece is listed.
- Exports `reconcilePieceStatus` (and its `PieceStatusContext` / `PieceStatusResult` types) from `core/piece`, which was previously unexported.

## Details
The command reuses `getDetailedDataSet` (which already reconciles on-chain and provider-reported piece status via `reconcilePieceStatus`) and renders results through a new `displayPieceStatuses` helper that shares the existing per-piece renderer. Because there is no piece-to-dataset reverse lookup, the command is scoped to a single dataset, as called out in the issue.

Part of #522.

## Test plan
- [x] `data-set --help` lists `piece-status`
- [x] `data-set piece-status --help` shows both arguments
- [x] `reconcilePieceStatus` is importable from `filecoin-pin/core/piece`
- [x] unit tests: lists all pieces, filters by pieceCid, empty-message on miss, rejects invalid id
- [x] cli-network-defaults includes the new leaf (has `--network`)
- [x] typecheck + biome clean

### DIFF
--- a/src/data-set/run.ts
+++ b/src/data-set/run.ts
@@ -62,12 +62,13 @@ export async function runDataSetDetailsCommand(dataSetId: number, options: DataS
 /**
  * Display the reconciled status of a data set's pieces.
  *
- * When `pieceCid` is provided, only the matching piece (by PieceCID or IPFS root
+ * When `cid` is provided, only the matching piece (by PieceCID or IPFS root
  * CID) is shown; otherwise every piece in the data set is listed.
  */
 export async function runDataSetPieceStatusCommand(
   dataSetId: number,
-  pieceCid: string | undefined,
+  // pieceCid or ipfsRootCid
+  cid: string | undefined,
   options: DataSetCommandOptions
 ): Promise<void> {
   if (Number.isNaN(dataSetId) || dataSetId <= 0) {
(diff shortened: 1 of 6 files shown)


src/data-set/run.ts
```diff
@@ -58,6 +59,68 @@ export async function runDataSetDetailsCommand(dataSetId: number, options: DataS
...
+
+    spinner.message('Fetching piece status...')
+
+    const dataSet: DataSetSummary = await getDetailedDataSet(synapse, BigInt(dataSetId))
+    const allPieces = dataSet.pieces ?? []
+
+    let pieces: PieceInfo[] = allPieces
+    let emptyMessage: string | undefined
+    if (pieceCid != null) {
+      pieces = allPieces.filter((piece) => piece.pieceCid === pieceCid || piece.rootIpfsCid === pieceCid)
```
SgtPooki:
```suggestion
      pieces = allPieces.filter((piece) => piece.pieceCid === cid || piece.rootIpfsCid === cid)
```

should we change the arg name to cid instead of `pieceCid` since we're treating it as either a pieceCID or a non pieceCid?

src/data-set/run.ts
```diff
@@ -58,6 +59,68 @@ export async function runDataSetDetailsCommand(dataSetId: number, options: DataS
...
 
+/**
+ * Display the reconciled status of a data set's pieces.
+ *
+ * When `pieceCid` is provided, only the matching piece (by PieceCID or IPFS root
+ * CID) is shown; otherwise every piece in the data set is listed.
+ */
+export async function runDataSetPieceStatusCommand(
+  dataSetId: number,
+  pieceCid: string | undefined,
```
SgtPooki:
```suggestion
  // pieceCid or ipfsRootCid
  cid: string | undefined,
```

SgtPooki (approved):
just some nits, otherwise lgtm

PR #540, opened 2026-06-03
feat(cli)!: unify data-set and provider ID flags

juliangruber:
Part of #279. Implements the "unify the data-set and provider ID flags" breaking-change item from #522.

## Summary
- Converge on repeatable `--data-set-id` / `--provider-id` across `add`, `import`, `rm`, and `data-set list`.
- Keep the old names working as **deprecated aliases** (with a warning):
  - `--provider-ids` / `--data-set-ids` (comma-separated) on `add` / `import`
  - `--data-set` (single value) on `rm`
  - `--provider-id` is now repeatable on `data-set list` (previously single)
- `rm` now accepts **multiple data sets** and processes each (a failure on one no longer aborts the rest; exit code is non-zero if any fail).
- Read the `PROVIDER_IDS` / `DATA_SET_IDS` env vars that the help text already advertised.
- `--help` shows only the new canonical flags; deprecated aliases are hidden.

## Implementation notes
- New flag builders `addProviderIdOption` / `addDataSetIdOption` in `cli-options.ts`; `addContextSelectionOptions` now composes them.
- `parseContextSelectionOptions` refactored into reusable `parseProviderIdSelection` / `parseDataSetIdSelection`, which gather from canonical flags, deprecated aliases, and env (in that precedence), validate/dedupe, and emit deprecation warnings. Provider and data-set selection remain mutually exclusive.
- rm runners stay single-dataset; the command loops over the parsed IDs.

## Breaking change
`rm --data-set` and the comma-separated `--provider-ids` / `--data-set-ids` flags are deprecated in favor of the repeatable `--provider-id` / `--data-set-id`. Old flags still parse but print a deprecation warning.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `biome check` passes on changed files
- [x] Unit suite passes (pre-existing `upload-action.test.ts` worktree module-linking failures are unrelated)
- [x] Added unit coverage for canonical flags, deprecated aliases, env vars, mutual exclusivity, and duplicate detection
- [x] Verified `add` / `rm` / `data-set list` `--help` shows the canonical flags and hides the aliases
- [x] Verified deprecated aliases and repeatable flags parse correctly


src/data-set/run.ts
```diff
@@ -66,20 +66,17 @@ export async function runDataSetListCommand(options: DataSetListCommandOptions):
...
-    // Parse and validate provider ID. Reject non-integers so BigInt() does
-    // not throw a low-level RangeError downstream.
-    const providerIdRaw = options.providerId != null ? Number(options.providerId) : undefined
-    if (providerIdRaw != null && (!Number.isInteger(providerIdRaw) || providerIdRaw <= 0)) {
-      throw new Error(`Invalid provider ID: '${options.providerId}' (must be a positive integer)`)
-    }
-    const providerId = providerIdRaw != null ? BigInt(providerIdRaw) : undefined
+    // Provider IDs come from the repeatable --provider-id flag (plus the
+    // deprecated --provider-ids alias and PROVIDER_IDS env). Validation and
+    // numeric parsing happen in parseProviderIdSelection.
```
juliangruber:
```suggestion
```

beck-8 (approved):
Non-blocking: I noticed there are still existing error messages in `src/add/add.ts` and `src/import/import.ts` that suggest `--data-set-ids` when metadata matching is too broad/narrow. 
`but expected ${resolution.expected} (narrow the filter or pass --data-set-ids to pin the target).`

PR #542, opened 2026-06-03
feat(payments)!: disambiguate fund and deposit day flags

juliangruber:
Part of #279. Addresses the final task of #522 (GA: CLI cleanup and consistency).

## Problem

`payments fund --days` and `payments deposit --days` shared a flag name but meant different things:

- **fund** reconciles your balance to an exact runway — it deposits **or withdraws** to hit the target.
- **deposit** is an additive top-up that **never withdraws**.

Same flag, opposite behavior. Easy to reach for the wrong command and get a surprising withdraw (or a no-op top-up).

## Change

Rename the flags so intent is explicit:

| Command | Old | New |
| --- | --- | --- |
| `payments fund` | `--days` | `--target-days` |
| `payments deposit` | `--days` | `--cover-days` |

- The old `--days` remains as a **hidden, deprecated alias** on both commands and prints a deprecation warning when used, so existing scripts keep working.
- Command descriptions and validation error messages now state the deposit/withdraw direction explicitly.

## Verification

- `npx tsc --noEmit` and `npx biome check` clean on changed files.
- `payments fund --help` shows `--target-days` (hides `--days`); `payments deposit --help` shows `--cover-days` (hides `--days`).
- Confirmed the deprecated `--days` still parses on both commands.
- Unit suite passes (the 4 unrelated `upload-action` failures are a pre-existing worktree package-resolution issue, not touched here).

## Breaking change

`--days` on `payments fund` and `payments deposit` is deprecated. Use `--target-days` and `--cover-days` respectively. The old flag still works but emits a deprecation warning.

### DIFF
--- a/src/commands/payments.ts
+++ b/src/commands/payments.ts
@@ -1,4 +1,4 @@
-import { Command } from 'commander'
+import { Command, Option } from 'commander'
 import { runAutoSetup } from '../payments/auto.js'
 import { runDeposit } from '../payments/deposit.js'
 import { runFund } from '../payments/fund.js'
@@ -6,6 +6,7 @@ import { runInteractiveSetup } from '../payments/interactive.js'
 import { showPaymentStatus } from '../payments/status.js'
 import type { FundOptions, PaymentSetupOptions } from '../payments/types.js'
 import { runWithdraw } from '../payments/withdraw.js'
+import { log } from '../utils/cli-logger.js'
 import { addAuthOptions } from '../utils/cli-options.js'
 
 export const paymentsCommand = new Command('payments').description('Manage payment setup for Filecoin Onchain Cloud')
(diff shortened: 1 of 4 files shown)


src/test/unit/payments-day-flags.test.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Make a subcommand test-safe: throw instead of exiting on parse errors, and
+// swallow the error output Commander writes to stderr.
+function parseSafe(cmd: Command, args: string[]): Promise<Command> {
+  return cmd
+    .exitOverride()
+    .configureOutput({ writeErr: () => undefined })
+    .parseAsync(args, { from: 'user' })
+}
```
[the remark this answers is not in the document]
juliangruber: [a010116](https://github.com/filecoin-project/filecoin-pin/pull/542/commits/a010116f8de0b81e5211446a5d2b8eb06ef9456f)

src/commands/payments.ts
```diff
@@ -45,23 +46,29 @@ const setupCommand = new Command('setup')
 addAuthOptions(setupCommand)
 paymentsCommand.addCommand(setupCommand)
 
-// Fund command - adjust funds to an exact runway or deposited total
+// Fund command - reconciles the balance to an EXACT target; can deposit OR withdraw.
```
juliangruber:
```suggestion
// Fund command - reconciles the balance to an exact target; can deposit or withdraw.
```

src/commands/payments.ts
```diff
@@ -106,17 +113,23 @@ const statusCommand = new Command('status')
 addAuthOptions(statusCommand)
 paymentsCommand.addCommand(statusCommand)
 
-// Deposit command
+// Deposit command - additive top-up ONLY; never withdraws.
```
juliangruber:
```suggestion
// Deposit command - additive top-up only; never withdraws.
```

SgtPooki (changes requested):
@juliangruber this conflicted a bit with #526 which I had two approvals on, so we might not need it? I think to keep the surface area simple, deposit should probably just be amount, and just be additive.. and fund should handle the more complex cases.

PR #518, opened 2026-05-29
feat: add session revoke command

shrirajpawar4:
## Summary

  Adds `filecoin-pin session revoke <session-address>` so wallet owners can revoke Filecoin Pin session-key permissions from the CLI.

  ## Changes

  - Added a core `revokeSessionAddress()` helper using `@filoz/synapse-core/session-key` `revokeSync`
  - Revokes the same Filecoin Pin FWSS permission set granted by session authorization
  - Added typed revoke progress events and result metadata
  - Added `src/session` runner and output formatter for the CLI flow
  - Wired `session revoke` into the existing `session` command tree
  - Added unit coverage for revoke call payload, progress events, registry resolution, tx hash/block, and checksummed session address
  - Added CLI help coverage for `session revoke`

  ## Verification

  ```bash
  pnpm exec vitest run src/test/unit/session-revoke.test.ts src/test/unit/session-authorize.test.ts src/test/unit/cli-network-defaults.test.ts
  
  pnpm exec biome check src/core/session/revoke-session.ts src/session/run-revoke.ts src/commands/session.ts src/test/unit/session-revoke.test.ts src/test/unit/cli-network-defaults.test.ts
```

Targeted tests and Biome checks pass.

Closes #504

PR #552, opened 2026-06-05
feat(core)!: remove authorizeSession:resolving event

SgtPooki:
## Problem

`authorizeSession:resolving` does not mark an observable phase. It fires synchronously back-to-back with `authorizeSession:submitting`; the only statement between the two emissions is a synchronous expiry calculation, and the resolution work it claims to mark (`getAddress` checksumming, registry address lookup) completes before the event is emitted. Neither CLI runner (`run-create.ts`, `run-authorize.ts`) consumes it; both progress switches start at `submitting`.

## Change

Removes the event from `AuthorizeSessionProgressEvents`, its emission in `authorizeSessionAddress`, and the test assertion. `ownerAddress` remains available in the function's return value.

BREAKING CHANGE: `AuthorizeSessionProgressEvents` no longer includes `authorizeSession:resolving`. Progress handlers should treat `authorizeSession:submitting` as the first event.

## Context

Came out of the #518 review, where the same vestigial event was about to be copied into `revokeSessionAddress` for symmetry. #518 drops `revokeSession:resolving`; this PR removes the original.

### DIFF
--- a/src/core/session/authorize-session.ts
+++ b/src/core/session/authorize-session.ts
@@ -65,11 +65,6 @@ export async function authorizeSessionAddress(
     throw new Error(`No session key registry address configured for chain id ${client.chain.id}`)
   }
 
-  options.onProgress?.({
-    type: 'authorizeSession:resolving',
-    data: { sessionAddress, ownerAddress },
-  })
-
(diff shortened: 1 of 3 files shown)


SgtPooki (commented):
self review

juliangruber (approved):

beck-8 (approved):

Issue #553, opened 2026-06-05
server command silently overrides PORT and HOST env vars


SgtPooki:
## Description

The `server` command declares `--port` and `--host` with Commander defaults (`'3000'`, `'127.0.0.1'` at `src/commands/server.ts:7-8`). A default means `options.port` is always set, so the action's copy-back block (`if (options.port) { process.env.PORT = options.port }`) runs on every invocation and overwrites `PORT`/`HOST` before `createConfig()` reads them in `src/config.ts`.

`PORT` and `HOST` are documented as supported env vars, but they are dead for the server command.

## Steps to reproduce

1. `PORT=8080 HOST=0.0.0.0 filecoin-pin server`

## Expected behavior

Server binds `0.0.0.0:8080`.

## Actual behavior

Server binds `127.0.0.1:3000`.

## Proposed fix

Same pattern #536 applied to the signing flags: declare both via `new Option(...).env('PORT')` / `.env('HOST')` with no Commander default, so env binds at parse time and flags still win.

One decision needed before the fix: the Commander defaults (3000/127.0.0.1) currently shadow the config defaults in `src/config.ts` (3456/localhost), so removing them changes the effective default port from 3000 to 3456. Pick one pair as canonical and align the two files.

PR #554, opened 2026-06-05
feat(cli): finish Option.env() rollout

SgtPooki:
## What changed

Completes the `Option.env()` convention #536 establishes:

- Shared option factories `privateKeyOption` / `sessionKeyOption` / `rpcUrlOption` in `cli-options.ts`, so each flag/env pairing is declared once and only descriptions vary per command. `addSigningAuthOptions`, `addOwnerAuthOptions`, `addAuthOptions`, `session create`, and `server` now share them.
- `server --access-token` migrates to `Option.env('ACCESS_TOKEN')`, removing the last hand-written `(env: ...)` prose on that help screen for a string flag.
- Comments document the intentional exemptions: `--allow-no-auth` (Commander treats a defined env var as true for boolean options regardless of value, while the config loader requires `ALLOW_NO_AUTH === 'true'`) and `--provider-id` / `--data-set-id` (env handling lives in `gatherIdSelection`, which comma-splits and lets a flag fully replace the env value).
- Removes the `options.x || process.env.X` fallbacks in `cli-auth.ts`, `run-create.ts`, `run-authorize.ts`, and `payments/interactive.ts`. Every command that reaches these paths binds the same env vars through Commander, which owns flag-over-env precedence; the devnet private-key fallback stays.
- Env-binding tests for `addOwnerAuthOptions`, the `server` command, and `session create --session-key`.

## Notes / risks

Based on #536; merge that first (this PR targets its branch and will retarget to master when it merges). Behavior is unchanged for every flag/env combination; the fallback removal relies on the `.env()` bindings delivering the same values at parse time.

`--port` / `--host` are not migrated here; #553 tracks that separately because it changes the effective default port.

### DIFF
--- a/src/commands/server.ts
+++ b/src/commands/server.ts
@@ -1,6 +1,6 @@
 import { Command, Option } from 'commander'
 import { startServer } from '../server.js'
-import { addNetworkOptions, addSigningAuthOptions } from '../utils/cli-options.js'
+import { addNetworkOptions, addSigningAuthOptions, rpcUrlOption } from '../utils/cli-options.js'
 
 export const serverCommand = new Command('server')
   .description('Start the IPFS Pinning Service API server')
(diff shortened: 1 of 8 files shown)


SgtPooki (commented):
self review

juliangruber (approved):

Issue #559, opened 2026-06-08
migrate payments code for synapse-sdk GA costs reshape (synapse-sdk#823)


SgtPooki:
### Description

FilOzone/synapse-sdk#823 (~~open, unreleased~~) reshapes the Costs API for GA. Three filecoin-pin surfaces depend on what it removes or changes:

1. `getServicePrice()` is removed in favor of `getPriceList()`. Call sites: `src/payments/auto.ts:107`, `src/payments/status.ts:132`, `src/core/payments/index.ts:932`, `src/core/payments/funding.ts:437`. `src/core/payments/index.ts:50` also re-exports it, so the published core surface inherits the break (relevant to #279).
2. Operator-approval `maxLockupPeriod` moves from the hardcoded 30-day `LOCKUP_PERIOD` constant to a chain read (`getPriceList().lockups.defaultLockupPeriod`). `depositUSDFC` and `setServiceApprovals` pass an explicit 30-day value; that stays valid post-#823 but under-covers if a chain's default ever exceeds 30 days. `checkAllowances` mirrors the old `isFwssMaxApproved` comparison and will drift from the SDK's view; #823 adds a `requiredMaxLockupPeriod` option to `isFwssMaxApproved`, so the mirror can be deleted in favor of calling the SDK directly.
3. `calculateAdditionalLockupRequired` is reshaped; the funding paths that call it need adjusting.

### Impact

Nothing to do today: #823 is unmerged, blocked on contract redeploys, and the latest published SDK (0.41.0) matches our dependency. Once it ships, bumping `@filoz/synapse-sdk` / `@filoz/synapse-core` will not compile until these are migrated.

### Next steps

- Replace `getServicePrice()` call sites and the core re-export with `getPriceList()`.
- Drop the explicit `maxLockupPeriod` in `depositUSDFC` / `setServiceApprovals` and let the SDK resolve the chain default.
- Replace the comparison inside `checkAllowances` with the SDK's `isFwssMaxApproved`.

BigLep:
@SgtPooki : how does this relate to https://github.com/filecoin-project/filecoin-pin/issues/402 ? Are they overlapping and should be combined? I assume we just need one issue that is about "update filecoin pin to use latest cost changes and get all of those from synapse rather than bake into fileocin-pin directly".

SgtPooki:
[...]

PR #560, opened 2026-06-08
fix!: gate payments preflight on needed transactions

### LINKED ISSUE #557: payments preflight rejects already-configured accounts and reports no numbers
### Description

`payments setup` validates the wallet before reading on-chain state. `validatePaymentRequirements` (src/core/payments/index.ts:209) hard-fails when the wallet holds under 0.1 FIL (`MIN_FIL_FOR_GAS`, src/core/payments/constants.ts:19) or zero USDFC, and it runs before `getPaymentStatus` (src/payments/auto.ts:79). An account that is already fully configured (sufficient deposit, FWSS operator approved) gets blocked even though the run would send zero transactions. The rest of the flow is already idempotent: it deposits only the shortfall, `checkAndSetAllowances` skips unneeded updates, and it reports "Payment setup already configured - ready to use". A configured account with a low-gas wallet can never reach that path. A wallet that deposited all of its USDFC hits the same wall via the `No USDFC tokens found` branch.

The failure also prints no numbers. `✗ Insufficient FIL for gas fees` reads as "you have no FIL" whether the balance is 0 or 0.09892.

### Impact

Gas from a successful approve + deposit can drop a wallet from just above 0.1 FIL to just below it, locking a fully funded account out of every command. A mainnet user hit exactly this during onboarding; diagnosing it required reading contract state manually.

### Expected behavior

1. Read on-chain state first; require gas only when a transaction will actually be sent.
2. When the check does fail, print the numbers and the shortfall so users know exactly how much to add: `✗ Insufficient FIL for gas fees (balance: 0.0989 FIL, minimum: 0.1 FIL, add at least: 0.0011 FIL)`. `checkFILBalance` already returns the balance; `validatePaymentRequirements` takes only a boolean, so it needs the balance passed in. Same for `src/payments/deposit.ts:53` and `src/payments/withdraw.ts:47`.

SgtPooki:
### What changed

`payments setup` now reads on-chain state (deposit, operator allowances) before validating wallet funding, and only gates when the run will actually send transactions. A fully configured account reports "already configured" regardless of wallet FIL or USDFC balance. A deposit requires wallet USDFC and gas; an allowance update requires gas alone.

Gas failures now report the numbers: `Insufficient FIL for gas fees (balance: 0.0989 FIL, minimum: 0.1000 FIL, add at least: 0.0011 FIL)` via the new `validateGasRequirement`, used by setup, deposit, withdraw, fund, and upload preflight. `validatePaymentRequirements` takes the FIL balance instead of a precomputed boolean.

Fixes #557

### How to verify

Unit tests cover the message format and check ordering (`payment-validation.test.ts`). For the live path: with a configured account holding under 0.1 FIL, `payments setup --auto` completes instead of aborting; `payments deposit --amount 1` reports the exact shortfall.

### Notes / risks

The preflight gate and the allowance action both derive from `checkAllowances`, so they cannot disagree. Verified against FilOzone/synapse-sdk#823: no overlap with the Costs API reshape; the migration items it does create are tracked in #559.

### DIFF
--- a/src/core/payments/index.ts
+++ b/src/core/payments/index.ts
@@ -25,6 +25,7 @@ import {
 import { calibration, SIZE_CONSTANTS, type Synapse, TIME_CONSTANTS, TOKENS } from '@filoz/synapse-sdk'
 import { formatUnits, type Hash } from 'viem'
 import { getClientAddress, isSessionKeyMode } from '../synapse/index.js'
+import { formatFIL } from '../utils/format.js'
 import { assertPriceNonZero } from '../utils/validate-pricing.js'
 import {
   DEFAULT_LOCKUP_DAYS,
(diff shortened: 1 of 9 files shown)


src/core/payments/index.ts
```diff
@@ -206,20 +207,53 @@ export function getUsdfcAcquisitionHelpMessage(isCalibnet: boolean): string {
...
+ * allowance update), use {@link validateGasRequirement} so accounts that
+ * hold all their USDFC as deposits are not rejected.
+ *
+ * @param filBalance - Wallet FIL balance in attoFIL
+ * @param walletUsdfcBalance - Wallet USDFC balance (18 decimals)
+ * @param isCalibnet - Whether the network is Calibration testnet
+ */
 export function validatePaymentRequirements(
-  hasSufficientGas: boolean,
+  filBalance: bigint,
```
juliangruber, line 250: should this PR be marked as a breaking change?

juliangruber (approved):
lgtm, only concern is the pr/commit message

src/core/payments/index.ts
juliangruber: edit and merging, since we're rolling out a breaking release next anyway

PR #572, opened 2026-06-12
feat: prompt dataset selection on ambiguous --data-set-metadata match

### LINKED ISSUE #437: Guided dataset selection for add/import in multi-dataset wallets
Some users have multiple datasets on one wallet — Storacha-migrated users, users organizing content across spaces or projects, the pinning server, others we don't know about yet. For these users, `add` and `import` either smart-select silently (which can land in the wrong place — see #430 §2) or require the user to pre-know `--data-set-ids id1,id2`. There is no middle ground for "I want to add a file, help me pick the right destination."

Design a guided selection flow for `add` / `import` when the user has multiple plausible destinations and has not pinned a target.

Sketch:
1. User runs `filecoin-pin add file.json`.
2. If a sane default exists (e.g. one matching dataset pair, or a single pre-configured destination), use it. Default path stays one-keystroke.
3. If the choice is ambiguous (multiple matching pairs, mixed metadata), prompt:
   - List the caller's datasets with metadata snippets and piece counts.
   - Accept dataset IDs to upload to.
4. Run the upload against the chosen IDs.

Constraints:
- Must not break the no-prompt default flow for users who do not care.
- Must respect non-interactive contexts (CI, scripts) — follow the existing `--auto` flag pattern (used by `payments setup`, etc.); skip the prompt and either fall back to current smart-select behavior or error with a clear "ambiguous, pass --data-set-ids or --auto" message.
- Must not duplicate the work tracked in #435 (explicit `--data-set-metadata` resolution) or the future profile system.

Open questions:
- What counts as "ambiguous enough" to prompt vs auto-select?
- Should the prompt persist the choice (link to profile work) or be one-shot?
- How to surface this in the GitHub Action and pinning-server contexts?

Parent context: #430
Related: #435 (explicit `--data-set-metadata` resolver, narrower scope)

BravoNatalie:
Closes #437 

### What
  
When `--data-set-metadata` matches more datasets than `--copies` requires, `add` and `import` now prompt the user to pick which datasets to upload to instead of throwing a hard error.

Each option shows the dataset ID, the metadata keys that differ across the matched candidates (values truncated if long, empty-string values shown as bare keys), and the active piece count. Non-interactive contexts (CI, GitHub
Actions, no TTY) keep the hard error, the prompt only runs when `isInteractive()` is `true`.

### Verification

- `pnpm test` - all 627 tests pass.
- Run `filecoin-pin add file.json --data-set-metadata "space-name=X" --data-set-metadata "space-did=did:key:..."` against a wallet with 2+ datasets matching. Confirm the multiselect prompt appears and upload proceeds to the chosen pair.
- Run `filecoin-pin add file.json --data-set-metadata "space-name=X" --data-set-metadata "space-did=did:key:..."` against a wallet with exactly one matching pair. Confirm no prompt, resolves directly via matched path.

image

###   Notes

- `isInteractive()` (not `isTTY()`) is the gate, it also blocks `CI=true` and `GITHUB_ACTIONS=true`, so the GitHub Action and pinning server are covered without additional handling.
- The prompt enforces exact selection count (`expectedCopies`). If the user selects the wrong number, it uses a loop that halts the prompt and ask again.

### DIFF
--- a/src/common/upload-flow.ts
+++ b/src/common/upload-flow.ts
@@ -5,10 +5,12 @@
  * including payment validation, storage context creation, and result display.
  */
 
+import { isCancel, multiselect } from '@clack/prompts'
 import type { CopyResult, FailedAttempt, Synapse } from '@filoz/synapse-sdk'
 import type { CID } from 'multiformats/cid'
 import pc from 'picocolors'
 import type { Logger } from 'pino'
+import type { DataSetSummary } from '../core/data-set/types.js'
 import { DEFAULT_LOCKUP_DAYS, type PaymentCapacityCheck } from '../core/payments/index.js'
 import {
   checkUploadReadiness,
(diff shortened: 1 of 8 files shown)


src/common/upload-flow.ts
```diff
@@ -21,11 +23,106 @@ import { formatUSDFC } from '../core/utils/format.js'
...
+ */
+export async function promptDataSetSelection(
+  matchedDataSets: DataSetSummary[],
+  expectedCopies: number,
+  spinner: Spinner
+): Promise<bigint[]> {
+  if (!isInteractive()) {
+    throw new Error(
+      `--data-set-metadata matched ${matchedDataSets.length} data sets (${matchedDataSets.map((d) => d.dataSetId).join(', ')}) ` +
+        `but expected ${expectedCopies}. Narrow the filter, pass --data-set-ids, or run in a TTY to pick interactively.`
```
juliangruber:
```suggestion
        `but expected ${expectedCopies}. Narrow the filter, pass --data-set-id, or run in a TTY to pick interactively.`
```

src/common/upload-flow.ts
```diff
@@ -21,11 +23,106 @@ import { formatUSDFC } from '../core/utils/format.js'
...
+
+  return { label }
+}
+
+/**
+ * Prompt the user to select exactly `expectedCopies` data sets from a list of candidates.
+ *
+ * Only called when `--data-set-metadata` matched more datasets than `--copies` requires and
+ * the process is running in an interactive TTY. In non-interactive contexts the caller must
+ * throw before reaching here.
```
juliangruber:
```suggestion
 * Only called when `--data-set-metadata` matched more datasets than `--copies` requires and
 * the process is running in an interactive TTY. Throws in non-interactive contexts.
```

src/core/data-set/resolve-by-metadata.ts
```diff
@@ -1,11 +1,12 @@
 import type { Synapse } from '@filoz/synapse-sdk'
 import type { Logger } from 'pino'
 import { listDataSets } from './list-data-sets.js'
+import type { DataSetSummary } from './types.js'
 
 export type MetadataResolution =
   | { kind: 'no-match' }
   | { kind: 'matched'; dataSetIds: bigint[] }
-  | { kind: 'too-many-matches'; matchedIds: bigint[]; expected: number }
+  | { kind: 'too-many-matches'; matchedIds: bigint[]; matchedDataSets: DataSetSummary[]; expected: number }
```
juliangruber, line 9: note: `matchedIds` is now redundant, can be deduced from `matchedDataSets`. Not necessary to fix now though

src/common/upload-flow.ts
```diff
@@ -21,11 +23,106 @@ import { formatUSDFC } from '../core/utils/format.js'
...
+  const visible = pairs.slice(0, MAX_LABEL_PAIRS)
+  const overflow = pairs.length - visible.length
+  const overflowSuffix = overflow > 0 ? `  (+${overflow} more)` : ''
+
+  const pieces = Number(ds.activePieceCount ?? 0n)
+  const piecesLabel = `(${pieces} piece${pieces !== 1 ? 's' : ''})`
+
+  const label = [`#${ds.dataSetId}`, ...visible, piecesLabel].join('  ') + overflowSuffix
+
+  return { label }
```
juliangruber: nit: return `label` string directly

src/common/upload-flow.ts
```diff
@@ -21,11 +23,106 @@ import { formatUSDFC } from '../core/utils/format.js'
 import { autoFund } from '../payments/fund.js'
 import type { AutoFundOptions } from '../payments/types.js'
 import type { Spinner } from '../utils/cli-helpers.js'
-import { cancel, formatFileSize } from '../utils/cli-helpers.js'
+import { cancel, formatFileSize, isInteractive } from '../utils/cli-helpers.js'
 import { log } from '../utils/cli-logger.js'
 import { createSpinnerFlow } from '../utils/multi-operation-spinner.js'
 import { CliFatal } from './cli-errors.js'
 
+function truncate(str: string, max: number): string {
```
juliangruber: nit: add a note or `assert()` because this is only safe to call for `max >= 7`

src/common/upload-flow.ts
```diff
@@ -21,11 +23,106 @@ import { formatUSDFC } from '../core/utils/format.js'
...
+  if (str.length <= max) return str
+  return `${str.slice(0, max - 7)}…${str.slice(-6)}`
+}
+
+/**
+ * Find the metadata keys whose values differ across the candidate set.
+ * Keys with identical values on every dataset add no signal to the prompt.
+ * Falls back to all keys when everything is uniform.
+ */
+function differentiatingKeys(dataSets: DataSetSummary[]): string[] {
```
juliangruber: `differentiatingKeys`, `buildOptionLabel` and `truncate` are untested. Would be nice to add tests, or add least some `assert()` to make them safer

juliangruber (changes requested):

src/common/upload-flow.ts
BravoNatalie: thanks! I've made it safe for `max < 7`

src/core/data-set/resolve-by-metadata.ts
BravoNatalie: yeah, I just didn't want to disrupt the existing pattern

juliangruber (approved):

PR #573, opened 2026-06-12
feat: add dataSetIds input parsing to upload action

### LINKED ISSUE #436: Add dataSetIds input to upload-action
`upload-action/action.yml` exposes `path, walletPrivateKey, network, minStorageDays, filecoinPayBalanceLimit, withCDN, dryRun` but no way to target specific datasets. CI consumers cannot route uploads to existing datasets — every workflow run either creates a new dataset or reuses the action's own previously-created `source=filecoin-pin` ones via SDK smart-select. Storacha-migrated users running CI cannot append to their migration pair without forking or wrapping the CLI directly.

Add a `dataSetIds` input to `action.yml`. Map it to the existing `executeUpload({ dataSetIds })` path used by the CLI's `--data-set-ids` flag (`upload-action/src/filecoin.js:190`).

Acceptance:
- `dataSetIds: "13260,13261"` in a workflow routes the upload to those datasets and skips creation.
- Omitted input preserves current behavior.
- Invalid input (non-numeric, mixed with `providerIds` if that ever surfaces) errors clearly.

Parent: #430
Investigation: https://gist.github.com/SgtPooki/d05e1ece2cdd9ae83ec003d2ec9e2979

BravoNatalie:
Closes #436 

### What
Add a dataSetIds input to the upload action so CI workflows can route uploads into existing datasets instead of always creating new ones or relying on SDK smart-select.

### Verification
- `pnpm vitest run src/test/unit/upload-action.test.ts` - all 12 tests pass

### Notes 
`dataSetIds` and `PROVIDER_IDS` are mutually exclusive, passing both fails at input parsing with a clear error.

### DIFF
--- a/src/test/unit/upload-action.test.ts
+++ b/src/test/unit/upload-action.test.ts
@@ -47,6 +47,7 @@ interface ParseInputsModule {
     contentPath: string
     network: 'mainnet' | 'calibration'
     dryRun: boolean
+    dataSetIds?: bigint[]
   }
 }
 
@@ -55,7 +56,7 @@ interface FilecoinModule {
     synapse: unknown,
     carPath: string,
     ipfsRootCid: string,
-    options: { withCDN: boolean; providerIds?: bigint[] },
+    options: { withCDN: boolean; providerIds?: bigint[]; dataSetIds?: bigint[] },
     logger: unknown
   ) => Promise<{
     dataSetId: string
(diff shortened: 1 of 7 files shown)


juliangruber (changes requested):
Implementation lgtm, https://github.com/filecoin-project/filecoin-pin/blob/master/upload-action/README.md needs to be updated

juliangruber (approved):

Issue #187, opened 2025-10-31
filecoin-pin v1


BigLep:
This issue captures notes about issues or tasks that need to be completed before v1 of filecoin-pin is released.  It's created as an issue vs. a milestone so it can be easily linked to by other issues in other repos.

Expectated time frame 202512.  This will come after https://github.com/filecoin-project/filecoin-pin/milestone/3 .  

In general, we are waiting to do a v1 release until:
1. We address some know immediate breaking changes (they'll be listed below)
2. We get a chance to pull our head up and identify if there is anything else in the short term that would be breaking that are wanting to change.  There will be a discussion in FDS-7 to cover this.

## Items that need to make it in for v1

- [ ] https://github.com/filecoin-project/filecoin-pin/issues/66
- [ ] https://github.com/filecoin-project/filecoin-pin/issues/306
- [x] https://github.com/filecoin-project/filecoin-pin/issues/91
- [ ] https://github.com/filecoin-project/filecoin-pin/issues/279
- [ ] crisp telemetry messaging
- [ ] #443

## Items that are unblocked by v1
- https://github.com/ipshipyard/ipfs-deploy-action/issues/39

rvagg:
Also need to default to mainnet before v1

BigLep:
Per 2026-05-26 conversations and in light of where https://github.com/ipshipyard/ipfs-deploy-action/issues/39 is at, I've updated the issue description with the minimum work required for the upload-action itself. I'll plan on handling the upload action work.

juliangruber:
@BigLep what does "crisp telemetry messaging" mean?

And are you still planning on handling the upload action work, or shall I take over?

BigLep:
> what does "crisp telemetry messaging" mean?

I'm meaning that we document that we have telemetry on by default, what is collected, and how to turn off. We ave it in https://github.com/filecoin-project/filecoin-pin#telemetry

> And are you still planning on handling the upload action work, or shall I take over?

That has been handled. I updated the checklist with the references

BigLep:
It's seeming like the synapse v1 update and the resulting costa api changes are what's remaining before we can cut the v1 release. I've updated the issue description.

BigLep:
@juliangruber : the Synapse v1 api changes are all in `main`, so we're unblocked to start updating filecoin-pin even as Synapse does the last remaining release items for Synapse v1.

BigLep:
@SgtPooki : anything more remaining here, or are we ok to resolve?

PR #574, opened 2026-06-15
feat!: upgrade to synapse-sdk v1.0.1 and synapse-core v0.7.0

### LINKED ISSUE #402: Use Synapse's costs API
filecoin-pin's costs / runway calculations predate the new APIs in Synapse, but now they should be treated as canonical. There's a whole lot of code in src/core/payments/funding.ts that has our custom logic. Maybe we have to keep some or most of it? But if we can defer the tricky logic to synapse, even just synapse-core, then that would be preferred so we don't have to maintain our own and hit things like the new sybil fee for clients.

### LINKED ISSUE #559: migrate payments code for synapse-sdk GA costs reshape (synapse-sdk#823)

juliangruber:
## Summary

Upgrades to `@filoz/synapse-sdk` **v1.0.1** and `@filoz/synapse-core` **v0.7.0**, adopting the GA per-operation pricing model and the v1 API surface.

Part of #187 (filecoin-pin v1).

### Closes #559 — payments migration for the GA costs reshape (synapse-sdk#823)
- `getServicePrice()` → `getPriceList()` at every call site and the published `core/payments` re-export
- dropped the hardcoded `maxLockupPeriod` from `depositUSDFC` / `setServiceApprovals`, the SDK now resolves the chain default (`priceList.lockups.defaultLockupPeriod`)
- `checkAllowances` defers the "is it at max?" decision to the SDK's `isFwssMaxApproved`

### Closes #402 — use Synapse's costs API
- all pricing now flows from the on-chain price list (`getPriceList` / `storageInfo.pricing.priceList`) instead of the removed `USDFC_SYBIL_FEE` / `CDN_FIXED_LOCKUP` constants and the `minimumPricePerMonth` floor
- adopted the reshaped `calculateAdditionalLockupRequired` (takes the price list, the per-data-set floor is now the additive `datasetFeePerMonth`), and the funding planner now threads a `priceList` through instead of `pricePerTiBPerEpoch` + `minimumPricePerMonth`
- new-data-set funding now defers to synapse-core's `calculateUploadFees({ isNewDataSet: true })` and adds the `lifecycleReserveTarget` lockup, parity-tested against `calculateDepositNeeded`. This widens the planned top-up for new data sets to cover add-piece fees and the lifecycle reserve, not just the create-data-set fee and CDN lockup

### Other v1.0.1 breaking changes handled
- `getSizeFromPieceCID(cid)` → `from(cid).size` (PieceCID now carries `.size`)
- session-key `DeleteDataSetPermission` → `TerminateServicePermission`
- `storage.terminateDataSet()` → `storage.terminateService({ dataSetId, skipProvider: true })` — `skipProvider` preserves the existing direct-termination UX

## Follow-ups

These are intentionally out of scope.

- fuller deferral of the funding planner to synapse-core's `calculateDepositNeeded` / `getUploadCosts` (this PR only routes the new-data-set fee component through `calculateUploadFees`)
- adopting provider-relayed termination (the new v1 default) for the `data-set terminate` command


src/test/unit/payments-setup.test.ts
```diff
@@ -45,6 +45,13 @@ vi.mock('@filoz/synapse-sdk', () => {
   }
 })
 
+// `checkAllowances` defers the "is this at max?" decision to the SDK helper.
+// Report "not max approved" so deposits take the approve-and-deposit path.
```
juliangruber:
```suggestion
// `checkAllowances` defers the maximum approved decision to the SDK helper.
// Report it's not, so deposits take the approve-and-deposit path.
```

src/payments/status.ts
```diff
@@ -126,15 +125,14 @@ export async function showPaymentStatus(options: StatusOptions): Promise<void> {
...
+    const [accountSummary, storageInfo] = await Promise.all([
       synapse.payments.accountSummary({}),
       synapse.storage.getStorageInfo(),
-      getServicePrice(synapse.client),
     ])
     const runway = toStorageRunwaySummary(accountSummary)
 
     const pricePerTiBPerEpoch = storageInfo.pricing.noCDN.perTiBPerEpoch
-    const minimumPricePerMonth = servicePrice.minimumPricePerMonth
+    const minimumPricePerMonth = storageInfo.pricing.priceList.rates.datasetFeePerMonth
```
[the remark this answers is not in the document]
juliangruber: Done in 08101af — renamed the variable to `datasetFeePerMonth` to match the GA per-data-set fee model.

src/test/unit/payments-funding.test.ts
```diff
@@ -231,11 +269,8 @@ describe('planFilecoinPayFunding', () => {
       synapse: makeSynapseStub() as any,
       targetRunwayDays: 10,
       pieceSizeBytes: 1024,
-      pricePerTiBPerEpoch: 1n,
-      minimumPricePerMonth: 60_000_000_000_000_000n,
     })
```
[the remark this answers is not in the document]
juliangruber: Done in 08101af — retitled to "adds create-data-set fees for new data sets in the shared funding plan".

src/test/unit/payments-funding.test.ts
```diff
@@ -617,17 +650,17 @@ describe('computeAutoSetupTargetBalance', () => {
       filecoinPayBalance: 0n,
       availableFunds: 0n,
       copies,
-      minimumPricePerMonth,
+      priceList,
     })
     expect(result.requiredAvailableFunds).toBe(3_320_000_000_000_000_000n)
```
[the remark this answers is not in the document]
juliangruber: Done in 08101af — updated the comment to use "create-data-set fee" instead of "sybil fee".

upload-action/src/filecoin.js
```diff
@@ -135,8 +133,7 @@ export async function handlePayments(synapse, options, logger) {
     allowWithdraw: false,
     targetRunwayDays: minStorageDays,
     pieceSizeBytes,
-    pricePerTiBPerEpoch: storageInfo.pricing.noCDN.perTiBPerEpoch,
-    minimumPricePerMonth: servicePrice.minimumPricePerMonth,
+    priceList: storageInfo.pricing.priceList,
     newDataSetCount,
     withCDN,
```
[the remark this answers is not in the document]
juliangruber: da862ab

src/data-set/run.ts
```diff
@@ -308,7 +308,13 @@ export async function runTerminateDataSetCommand(dataSetId: number, options: Dat
...
     spinner.start('Submitting termination transaction...')
 
-    const txHash = await synapse.storage.terminateDataSet({ dataSetId: BigInt(dataSetId) })
+    // Direct (client-submitted) termination preserves the existing confirmation
+    // UX: `skipProvider` keeps the old `terminateDataSet` behaviour where the
+    // owner wallet submits the on-chain transaction and always gets a tx hash.
+    const { txHash } = await synapse.storage.terminateService({ dataSetId: BigInt(dataSetId), skipProvider: true })
+    if (txHash == null) {
+      throw new Error('Termination did not return a transaction hash')
+    }
```
SgtPooki, line 317: i'm trying to remember what the tradeoff for users is with this.. if we skip provider, they can't release funds immediately and resolve things more quickly.. right?

do we want that to be the default in filecoin-pin?

upload-action/src/filecoin.js
```diff
@@ -148,9 +145,10 @@ export async function handlePayments(synapse, options, logger) {
   }
 
   if (newDataSetCount > 0) {
+    const feeDescription = withCDN ? 'create-data-set fee and CDN lockup' : 'create-data-set fee'
```
SgtPooki: i think this wording is slightly out of date now.. the deposit would also cover lifecycle reserve and add-piece fees. not sure how we want to word this right now

SgtPooki (approved):
mostly lgtm. I pushed up a change to fix an edge-case on pricing with test: [`730fcda` (this PR)](https://github.com/filecoin-project/filecoin-pin/pull/574/commits/730fcda955dff5f40e655e77059b980427f1d83c)

src/data-set/run.ts
juliangruber: Intentional, to keep this PR behaviour-preserving. `skipProvider: true` reproduces the old `terminateDataSet` UX where the owner wallet submits the on-chain tx and we always get a tx hash back to confirm/wait on.

The v1 default (provider-relayed) is the better long-term path. I've left "adopt provider-relayed termination for `data-set terminate`" as a follow-up rather than flipping the default in this upgrade PR.

Or should we break now?

upload-action/src/filecoin.js
juliangruber: a395e93

src/data-set/run.ts
SgtPooki: we are already breaking in this PR, right? if the question is "should we delay another breaking change outside of this breaking change" I would say no, if we're pretty confident we should do it in the future.

SgtPooki: I think maybe we should add a tracking issue for adding an option to CLI & lib to do the old functionality (skip provider) and just do the "right" path now

SgtPooki (approved):

src/data-set/run.ts
SgtPooki: opened https://github.com/filecoin-project/filecoin-pin/issues/581 to keep track of this

PR #580, opened 2026-06-17
feat: compact data-set list output

### LINKED ISSUE #578: feat: add compact summary view to `data-set list`
## Description

`data-set list` currently prints the full info for every data set, which makes the output hard to scan when users simply want to see which data sets they have.

Since `data-set show <datasetID>` already provides detailed information, `data-set list` could focus on a concise summary instead.

**Current output:**

```sh
│ #14093
│    Status: live
│    CDN add-on: enabled
│
│    Provider
│      ID: 9
│      Address: 0xa3171A7234a3379A1813d9867B531e7EeB20aw00
│
│    Metadata
│      source: "filecoin-pin"
│      withCDN: ""
│      withIPFSIndexing: ""
│
│    Payment
│      PDP rail ID: 10523
│      FilBeam rail ID: 10526
│      FilBeam cache-miss rail ID: 10525
│      Payer: 0xb4dd624f5ffb9a5a5144A67d596a4D2fA7c6140D
│      Payee: 0xa3171A7234a3379A1813d9867B531e7EeB20aw00
│
│    Pieces
│      Total pieces: 2
│
│  #14094
│    Status: live
│    CDN add-on: enabled
│
│    Provider
│      ID: 9
│      Address: 0xa3171A7234a3379A1813d9867B531e7EeB20aw00
│
│    Metadata
│      source: "filecoin-pin"
│      withCDN: ""
│      withIPFSIndexing: ""
│
│    Payment
│      PDP rail ID: 10527
│      FilBeam rail ID: 10530
│      FilBeam cache-miss rail ID: 10529
│      Payer: 0xb4dd624f5ffb9a5a5144A67d596a4D2fA7c6140D
│      Payee: 0xa3171A7234a3379A1813d9867B531e7EeB20aw00
│
│    Pieces
│      Total pieces: 0
```

## Proposed solution

Change the default output of `data-set list` to a compact one-row-per-dataset table showing only the necessary information.

```sh
ID      Status   Provider ID    Pieces     CDN
─────────────────────────────────────────────────────────────
14093   live          9           3       enabled
14190   live          2           3       enabled

2 data sets  ·  6 pieces  
Run `filecoin-pin data-set show <id>` for full details.
```

ps.: might need to add pagination

## Done criteria

- [ ] Default `data-set list` output is a compact table (one row per dataset)
- [ ] Table shows: ID, status, provider ID, piece count, CDN flag
- [ ] Footer shows aggregate totals

## Notes

This is distinct from the performance issue tracked in #362 (which covers timeouts at scale). This issue is about the default display for any list size.

shrirajpawar4:
## Summary

  Fixes #578

  This updates `filecoin-pin data-set list` / `ls` to use a compact one-row-per-data-set table while keeping detailed output for `data-set show`, `piece-status`, and termination flows.

  The compact list now includes:

  - `ID`
  - `Status`
  - `Provider ID`
  - `Pieces`
  - `Size`
  - `CDN`

  ## Implementation

  - Added an internal list-size enrichment helper under `src/core/data-set/`.
  - Calculates list row sizes by paging active on-chain pieces with `getActivePieces`.
  - Sums piece sizes using `getSizeFromPieceCID`.
  - Skips piece fetching for zero-piece or non-live data sets.
  - Treats size lookup failures as non-fatal and renders those rows as `unknown`.
  - Added `displayDataSetList()` for compact table rendering.
  - Updated only `runDataSetListCommand()` to use the compact renderer.

  ## Visual Evidence

  Before, `data-set list` used the same detailed multi-section renderer as `show`.

  After, `data-set list` renders compact rows like:

  ```text
  ID   Status  Provider ID  Pieces  Size     CDN
  158  live    2            3       3.0 MiB  disabled

  1 data sets, 3 active pieces, 3.0 MiB total known size
  Run `filecoin-pin data-set show <id>` for full details.
```
  ## Verification

  Passed:
```
  pnpm exec vitest run --project unit src/test/unit/core-data-set.test.ts src/test/unit/data-set.test.ts src/test/unit/data-set-display.test.ts
  pnpm exec biome check --no-errors-on-unmatched --files-ignore-unknown=true src/core/data-set/enrich-list-sizes.ts src/data-set/display.ts src/data-set/run.ts src/test/unit/core-data-set.test.ts src/
  test/unit/data-set.test.ts src/test/unit/data-set-display.test.ts
  git diff --check
  ```


src/core/data-set/enrich-list-sizes.ts
```diff
@@ -0,0 +1,76 @@
...
+export interface DataSetListRow extends Omit<DataSetSummary, 'totalSizeBytes'> {
+  totalSizeBytes?: bigint
+  sizeKnown: boolean
+}
+
+function getActivePieceCount(dataSet: DataSetSummary): bigint {
+  return dataSet.activePieceCount ?? 0n
+}
+
+async function calculateActivePieceSize(
```
BravoNatalie: After running it, I realized that the dataset size isn't a direct property of the dataset and has to be calculated from its pieces. This adds a significant amount of work to what should be a simple dataset ls command.

Given that, could you remove the size column?

BravoNatalie (changes requested):

src/data-set/display.ts
```diff
@@ -20,6 +20,18 @@ function statusLabel(dataSet: DataSetSummary): string {
   return pc.yellow('inactive')
 }
 
+function plainStatusLabel(dataSet: Pick<DataSetSummary, 'isLive' | 'pdpEndEpoch'>): string {
+  if (dataSet.isLive) {
+    return 'live'
+  }
+
+  if (dataSet.pdpEndEpoch > 0) {
+    return `terminated @ epoch ${dataSet.pdpEndEpoch}`
```
BravoNatalie: I think it's better to just show `Terminated` so we don't use too much space. If someone needs more details, like the termination epoch, they can get it from `dataset show`.

src/data-set/display.ts
```diff
@@ -209,7 +241,48 @@ export function displayPieceStatuses(
...
+): void {
+  if (dataSets.length === 0) {
+    log.line(pc.yellow(emptyMessage ?? 'No data sets managed by filecoin-pin were found for this account.'))
+    log.flush()
+    return
+  }
+
+  renderNetworkDetails(network, address)
+
+  const ordered = [...dataSets].sort((a, b) => (a.dataSetId < b.dataSetId ? -1 : a.dataSetId > b.dataSetId ? 1 : 0))
```
BravoNatalie, line 271: If possible, could we try centering the values? Only if it's not too much trouble.

Image

src/core/data-set/get-data-set-pieces.ts
```diff
@@ -17,6 +17,22 @@ import { type DataSetPiecesResult, type GetDataSetPiecesOptions, type PieceInfo,
 
 const ACTIVE_PIECES_BATCH_SIZE = 100n
 
+type PieceSizeTools = {
+  getSizeFromPieceCID?: (pieceCid: unknown) => number
+  from?: (pieceCid: unknown) => { size: number }
+}
+
+function getPieceSize(pieceCid: unknown): number {
```
BravoNatalie: Can we drop `getPieceSize`/`PieceSizeTools` here and call `pieceFromCID(...).size` directly, to match `master`?  
The `master` branch already does this since #574 
Unless there's a specific reason to keep it that I'm missing.

BravoNatalie (changes requested):
Everything looks good, but I have one point about the changes in `get-data-set-pieces.ts`.

src/core/data-set/get-data-set-pieces.ts
shrirajpawar4: I added this because some CI were failing. reverting back.

BravoNatalie (approved):
Looks great!

PR #566, opened 2026-06-09
chore(master): release 1.0.1

FilOzzy:
📦 Release Preparation
---

### ⚠ BREAKING CHANGES

* upgrade to synapse-sdk v1.0.1 and synapse-core v0.7.0 ([#574](https://github.com/filecoin-project/filecoin-pin/issues/574))

### Features

* add dataSetIds input parsing to upload action ([#573](https://github.com/filecoin-project/filecoin-pin/issues/573)) ([79b5d1e](https://github.com/filecoin-project/filecoin-pin/commit/79b5d1e9e37e3872ba06d130ce03c9587918bc38))
* add session revoke command ([#518](https://github.com/filecoin-project/filecoin-pin/issues/518)) ([48b6f64](https://github.com/filecoin-project/filecoin-pin/commit/48b6f64d96cb63e97313c6912aaece6606352974))
* prompt dataset selection on ambiguous --data-set-metadata match ([#572](https://github.com/filecoin-project/filecoin-pin/issues/572)) ([6778145](https://github.com/filecoin-project/filecoin-pin/commit/67781455ccf8d2b19a026531987b351ad069c5f4))
* upgrade to synapse-sdk v1.0.1 and synapse-core v0.7.0 ([#574](https://github.com/filecoin-project/filecoin-pin/issues/574)) ([d005cec](https://github.com/filecoin-project/filecoin-pin/commit/d005cec97805ed0db84ff301773ef4463b9e76a5))

### Bug Fixes

* run upload action test on push too ([#558](https://github.com/filecoin-project/filecoin-pin/issues/558)) ([7a996a5](https://github.com/filecoin-project/filecoin-pin/commit/7a996a58fdea57dbf17d7b56922ec6f5df02f78d))

### Chores

* configure release-please for proper semver ([#582](https://github.com/filecoin-project/filecoin-pin/issues/582)) ([bad16db](https://github.com/filecoin-project/filecoin-pin/commit/bad16dbd1969dd7761f452ab126bc3bbaa6b1c3d))
* release as 1.0.1 ([#586](https://github.com/filecoin-project/filecoin-pin/issues/586)) ([4f6ba80](https://github.com/filecoin-project/filecoin-pin/commit/4f6ba809f1743674313df17d7804395ef381b6dc))

### Documentation

[...]

PR #583, opened 2026-06-17
revert: configure release-please for proper semver (#582)

SgtPooki:
Reverts #582. That change was correct, but its squash merge dropped the commit body and the `Release-As: 1.0.1` trailer, so release-please recomputed PR #566 to 1.0.0 (already taken on npm and the v1.0.0 tag).

Reverting restores a clean base. The follow-up PR re-applies the same semver config together with the `Release-As: 1.0.1` trailer, merged via a merge commit so the trailer reaches master.

Do not merge release PR #566 until both this and the follow-up have landed.


SgtPooki:
Consolidating into a single one-line PR; see replacement.

PR #584, opened 2026-06-17
chore: configure release-please for proper semver

SgtPooki:
Re-applies the semver config from #582 (reverted in #583) together with the version-forcing trailer that the earlier squash merge dropped.

Sets `bump-minor-pre-major` and `bump-patch-for-minor-pre-major` to `false` so breaking changes bump major, features minor, fixes patch on the 1.x line. Drops the ignored `always-bump-patch` key. The commit carries a `Release-As: 1.0.1` trailer to pin the next release past the accidental 1.0.0 (#356); it self-expires after one release.

## Merge order

1. Merge #583 (revert) first. GitHub then retargets this PR to `master`.
2. Merge this PR with a **merge commit**, not squash. A squash drops the commit body and the `Release-As: 1.0.1` trailer, which is what sent #566 to 1.0.0 last time.

## How to verify

After merge, release-please reruns and regenerates #566 to 1.0.1. Confirm its changelog compares from `v0.23.2`.


SgtPooki:
Consolidating into a single one-line PR. master already has the correct semver config from #582 (the squash kept the file change, only dropped the trailer), so the revert+redo dance is unnecessary.

PR #586, opened 2026-06-17
chore: release as 1.0.1

SgtPooki:
Empty commit carrying a `Release-As: 1.0.1` footer to force the next release to 1.0.1.

1.0.0 was already claimed by the accidental release #356 (npm + the v1.0.0 tag), so the synapse-sdk v1 upgrade can't reuse it. The footer overrides only the version label; the changelog baseline stays at `v0.23.2`. The proper-semver config from #582 is already on master, so no file change is needed here, hence the empty commit. The footer self-expires after this release, so there is no config cleanup.

## Merge with a merge commit, not squash

A squash strips the commit body and drops the footer, which is what sent #566 to 1.0.0 last time. Use **Create a merge commit** (or rebase) so `Release-As: 1.0.1` reaches master.

## How to verify

After merge, release-please regenerates #566 to 1.0.1. Confirm its changelog compares from `v0.23.2`.


BigLep (approved):
Is this the approach synapse recently took? They had the same problem of needing to actually be 1.0.1 since they had accidentally released 1.0.0 as  well.

BigLep:
Synapse's PR: https://github.com/FilOzone/synapse-sdk/pull/834

rjan90 (approved):

SgtPooki:
> Is this the approach synapse recently took? They had the same problem of needing to actually be 1.0.1 since they had accidentally released 1.0.0 as well.

They did config change that needs a “revert commit”

i messed up and squash merged which removed the git commit body trailer message that does the same thing

PR #595, opened 2026-06-24
 feat: add `--dry-run` flag to `add` and `import` commands

### LINKED ISSUE #591: feat: add `--dry-run` flag to `filecoin-pin add` for pre-upload cost estimates
### Description
Users have no way to see the packed size or storage cost of a file or directory before committing to an upload. The only way to find out is to run the full `add` flow, which means cost is discovered after the upload has already started rather than before any commitment is made.

### Proposed solution

Add a `--dry-run` flag to `filecoin-pin add <path>`. With the flag set, the command packs the CAR and computes the packed size exactly as it does today, prints the estimated cost and the number of copies that cost covers, then exits before any upload or deposit transaction runs. Reuse the existing CAR-packing/sizing logic from the `add` flow rather than writing a separate estimator.

### Done criteria

 - [x] `filecoin-pin add <path> --dry-run` packs the CAR and prints the estimated cost without uploading
 - [x] Output includes the packed size and the number of copies the cost covers
 - [x] No deposit or upload transaction is triggered when `--dry-run` is set
 - [x] Existing add-flow tests are extended to cover the dry-run branch

BravoNatalie:
Closes #591

### What
Adds `--dry-run` to both `add` and `import`. Packs the CAR (add) or validates the file (import) as usual, then estimates cost across all requested copies using `calculateMultiContextCosts()` and exits without uploading or moving funds. Also fixes a pre-existing bug in `autoFund` where missing `WITH_IPFS_INDEXING` metadata caused all contexts to resolve as new data sets, inflating deposit estimates.

### Verification
- `pnpm test` - 634 unit tests pass
- `filecoin-pin add <file> --dry-run` prints cost breakdown, no upload
- `filecoin-pin import <car> --dry-run` same
- `--dry-run --auto-fund` rejected by Commander with a conflict error

image

### Notes
`--dry-run` skips the early WarmStorage allowance-approval check in `add`, which would otherwise submit an on-chain transaction before the dry-run exits.

`calculateMultiContextCosts()` was chosen over calling `getUploadCosts()` once per copy because `getUploadCosts()` independently re-derives account-level debt, runway, and buffer on every call. `calculateMultiContextCosts()` sums per-context lockup and fees while computing the account-level terms exactly once, giving a correct aggregate.

The current upload path's payment check (`validatePaymentCapacity` / `calculatePieceUploadRequirements`) still uses a single-context, copies-unaware estimator that hardcodes `isNewDataSet: false`. A follow-up could replace it with `calculateMultiContextCosts()` or `prepare()` directly for consistent, accurate pre-upload estimates.

### DIFF
--- a/src/common/upload-flow.ts
+++ b/src/common/upload-flow.ts
@@ -6,12 +6,14 @@
  */
 
 import { isCancel, multiselect } from '@clack/prompts'
-import type { CopyResult, FailedAttempt, Synapse } from '@filoz/synapse-sdk'
+import type { CopyResult, FailedAttempt, Synapse, UploadCosts } from '@filoz/synapse-sdk'
+import { METADATA_KEYS } from '@filoz/synapse-sdk'
 import type { CID } from 'multiformats/cid'
 import pc from 'picocolors'
 import type { Logger } from 'pino'
 import type { DataSetSummary } from '../core/data-set/types.js'
 import { DEFAULT_LOCKUP_DAYS, type PaymentCapacityCheck } from '../core/payments/index.js'
+import { DEFAULT_COPIES } from '../core/synapse/constants.js'
 import {
   checkUploadReadiness,
   executeUpload,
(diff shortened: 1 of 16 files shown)


src/core/upload/index.ts
```diff
@@ -162,7 +162,7 @@ export async function checkUploadReadiness(options: UploadReadinessOptions): Pro
 
   onProgress?.({ type: 'validatingCapacity' })
 
-  const capacityCheck = await validatePaymentCapacity(synapse, fileSize)
+  const capacityCheck = await validatePaymentCapacity(synapse, fileSize) // TODO: checks allowances and set max allowances
```
SgtPooki:
```suggestion
  const capacityCheck = await validatePaymentCapacity(synapse, fileSize) // <tracking issue if not implemented>
```

src/common/upload-flow.ts
```diff
@@ -422,6 +424,139 @@ function displayPaymentIssues(capacityCheck: PaymentCapacityCheck, fileSize: num
...
+  const requestedCopies = options.providerIds?.length ?? options.dataSetIds?.length ?? options.copies ?? DEFAULT_COPIES
+
+  // Mirror uploadToSynapse's metadata injection: when not targeting specific data
+  // sets by ID, prepend WITH_IPFS_INDEXING so metadataMatches() finds existing
+  // data sets that the real upload would also resolve to (it uses exact key-count
+  // matching — omitting this key causes all contexts to resolve as new data sets).
+  const resolvedMetadata: Record<string, string> =
+    options.dataSetIds != null
+      ? (options.metadata ?? {})
+      : { [METADATA_KEYS.WITH_IPFS_INDEXING]: '', ...(options.metadata ?? {}) }
```
SgtPooki: this code is duplicated three times, we should probably abstract a helper

SgtPooki (changes requested):
two small changes needed i think

SgtPooki (approved):

PR #596, opened 2026-06-25
docs: pin upload-action examples to @master, recommend specific version or SHA for production

BigLep:
## Summary

Updates the upload-action README examples and example workflows to pin to `@master`, matching what [filecoin-pin-website's CI](https://github.com/filecoin-project/filecoin-pin-website/blob/main/.github/workflows/filecoin-pin-upload.yml) actually uses. Each `@master` pin has an inline comment pointing readers to a specific version or commit SHA for production.

Background: the prior README examples (and this branch's first commit) used `@v0` / `@v1`, but no floating major-version ref exists on the repo today, so those pins fail to resolve. Caught in review by @rjan90.

## Changes

- README code examples (3 of them) and both example workflows: `@v1` → `@master` with a leading comment recommending specific version / SHA pins for production.
- README **Versioning and Updates** section: restructured to lead with specific version tag (production), then commit SHA (max supply-chain security), then `@master` (with a clear caveat). Added a NOTE that floating major-version tags like `@v1` are not currently maintained.
- README **Security Checklist**: rewritten to match the new ordering.

## Follow-ups (not in this PR)

- If we want the floating-major-tag UX (`@v1`) at some point, file a follow-up to either (a) manually create and maintain `v1`, or (b) automate it via the release-please job.

## Test plan

- [ ] Visual: render the README on the branch and confirm the **Versioning and Updates** section reads clearly
- [ ] Verify the two example workflow YAMLs still parse (no incidental indentation drift)
- [ ] Sanity-check that `@master` actually resolves for the action (it should; that's what filecoin-pin-website already uses)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/upload-action/README.md
+++ b/upload-action/README.md
@@ -32,7 +32,7 @@ If `path` points to a regular file whose name ends in `.car`, the action skips i
 
 ```yaml
 - name: Upload pre-built CAR to Filecoin
-  uses: filecoin-project/filecoin-pin/upload-action@v0
+  uses: filecoin-project/filecoin-pin/upload-action@v1
   with:
     path: build.car
     walletPrivateKey: ${{ secrets.FILECOIN_WALLET_KEY }}
@@ -46,7 +46,7 @@ For most users, automatic provider selection is recommended. However, for advanc
 
 ```yaml
 - name: Upload to Filecoin
-  uses: filecoin-project/filecoin-pin/upload-action@v0
+  uses: filecoin-project/filecoin-pin/upload-action@v1
   env:
     PROVIDER_IDS: "1,2"  # Target specific providers by their numeric on-chain IDs
   with:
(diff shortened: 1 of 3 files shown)


upload-action/README.md
```diff
@@ -32,7 +32,7 @@ If `path` points to a regular file whose name ends in `.car`, the action skips i
 
 ```yaml
 - name: Upload pre-built CAR to Filecoin
-  uses: filecoin-project/filecoin-pin/upload-action@v0
+  uses: filecoin-project/filecoin-pin/upload-action@v1
```
rjan90: little late/not-so-fast brain on my side here, but do we actually have a moving v1 ref?

I’m only seeing exact version tags like v1.0.0, v1.0.1, and v1.1.0, not refs/tags/v1 or a v1 branch. If that’s right, the new uses: filecoin-project/filecoin-pin/upload-action@v1 examples would still fail to resolve until we create/update that ref? Or am I missing something substantial here

rjan90 (approved):
One question from my side here due to it being late, but apart from that nothing that should be blocking for this PR

upload-action/README.md
BigLep: Doh - you are right. I didn't realize this. I'm getting this fixed.

Issue #470, opened 2026-05-08
Add CLI error-scenario smoke test matrix for devs


SgtPooki:
## Description

Manually verifying CLI error scenarios across all commands is slow and ad-hoc. During PR #468 (CliFatal sentinel work) we hand-rolled a 15-scenario bash sweep to confirm error output and exit codes. We should turn this into a reusable script or test suite so future CLI changes can re-run the matrix in one command.

## Goals

- Cover every command + every documented error scenario (invalid args, missing flags, bad RPC, no auth, insufficient funds, user cancel, etc.)
- Capture stdout, stderr, exit code per scenario
- Run without live RPC where possible (use mocks or no-network failure paths)
- Clear pass/fail per scenario; easy to add new scenarios

## Acceptance criteria

- [ ] `pnpm run smoke` (or similar) executes the full matrix
- [ ] Output format makes it easy to spot regressions (snapshot diff or labeled per-scenario block)
- [ ] At minimum covers: `payments setup/fund/deposit/withdraw/status`, `add`, `import`, `data-set show/list/terminate`, `provider list/show/ping`, `rm`
- [ ] Argument-validation scenarios (no network, no auth) run in CI
- [ ] Network-dependent scenarios (insufficient funds, bad RPC) are clearly tagged and skipped in CI by default
- [ ] README or `CONTRIBUTING.md` documents how to run + add scenarios

## Notes

- See PR #468 conversation for the manual sweep template
- Reuse existing `src/test/mocks/synapse-sdk.js` where mocking is preferable to live RPC
- Consider snapshot testing on captured stderr (vitest `toMatchInlineSnapshot`)
- Out of scope: replacing existing unit/integration tests

CodeByD3v:
Hi, I'd like to take a stab at this. My plan is to start with the argument-validation and no-network scenarios using the existing synapse-sdk mock setup and Vitest snapshot testing for stderr/stdout + exit codes. Then I'll look into structuring the matrix so network-dependent scenarios can be tagged/skipped in CI by default.

Let me know if there are any existing conventions or edge cases from the PR #468 sweep that I should keep in mind before I begin.

SgtPooki:
@CodeByD3v I think network operations would be okay against calibration network, and ran manually by users.. 

the more we can get into CI, the better, but I think we will ultimately want a non-stubbed/mocked backend running through the entire flow that a single user can run and analyze

SgtPooki:
[...]

PR #604, opened 2026-07-01
test: add CLI error-scenario smoke test matrix

CodeByD3v:
Relates to #470 (PR 1 of 2)

Adds a CI-safe smoke test suite that spawns the built CLI as a real
subprocess and asserts exit codes + snapshots for arg-validation,
missing-flag, and no-auth error scenarios across payments, add,
import, data-set, provider, rm, and session.

Notes:
- Script named `test:smoke` to match existing `test:unit` / `test:browser` convention
- Snapshots use `result.combined` (stdout + stderr merged) because clack's UI writes to stdout, not stderr
- Network-dependent scenarios (bad RPC, insufficient funds, live provider queries) deferred to PR 2

## Test run
image
image
image
image
image


BigLep:
Thanks @CodeByD3v. Some folks have been away on holiday. We'll look at this next week if not sooner. We'll also need to figure out the test failures.

CodeByD3v:
@BigLep Thanks for the update! No worries at all, I completely understand. I'll take a look at the test failures from my side as well. Looking forward to your feedback whenever you get a chance.

SgtPooki (changes requested):
Thanks so much for getting this out! after looking at this, I feel like I shouldn't have mentioned `snapshots` in the tracking issue at all.. lets go ahead and remove those because the eventual rubber-stamping of changed output won't do us any good in the long run.

1. Swap exact snapshots for exit-code + one toContain per scenario to assert the expected error happened.. we want low-maintenance overhead while also getting coverage.
2. Wire `test:smoke` into CI (ubuntu only), we should make sure that we're aware of how changes to commands is altering failure scenarios.. but those shouldn't be required for 
3. pnpm lint:fix to unblock CI

CodeByD3v:
Thanks for the feedback! That makes sense. I'll update the PR accordingly.

CodeByD3v:
**Addressed Issues**
---
1. **Snapshots removed:** Swapped all `toMatchInlineSnapshot()` calls for `exitCode` + one `toContain()` per scenario asserting the key error phrase. This removes snapshot maintenance overhead going forward.

2. **Smoke tests wired into CI:** Added a `smoke-test` job to `test.yml` running on `ubuntu-latest` with both `lts/*` and `current` Node versions, matching the pattern of the existing `browser-test` job.

3. **Lint fixed:** Ran `pnpm lint:fix` to resolve the Biome import ordering and trailing newline issues in `utils.ts` and `cli-error-scenarios.test.ts`.

src/test/smoke/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    for (const key of STRIP_AUTH) {
+      delete env[key]
+    }
+
+    const child = spawn('node', [CLI_PATH, '--no-update-check', ...args], {
+      cwd: PROJECT_ROOT,
+      env,
+      stdio: 'pipe',
+    })
```
SgtPooki, line 36: is there any reason we can't just clear out the environment variables entirely?, and allow passing env to `runCli` ?

```suggestion
export async function runCli(args: string[] = [], env: Record<string, string> = {}): Promise<CliRunResult> {
  return new Promise((resolve) => {
    const child = spawn('node', [CLI_PATH, '--no-update-check', ...args], {
      cwd: PROJECT_ROOT,
      env: {},
      stdio: 'pipe',
    })
```

src/test/smoke/cli-error-scenarios.test.ts
```diff
@@ -0,0 +1,447 @@
...
+    expect(result.exitCode).toBe(1)
+    expect(result.combined).toContain('Interactive mode requires a TTY terminal')
+  })
+
+  it('unknown flag', T, async () => {
+    const result = await runCli(['payments', 'setup', '--bogus-flag'])
+    expect(result.exitCode).toBe(1)
+    expect(result.combined).toContain("unknown option '--bogus-flag'")
+  })
+})
```
SgtPooki, line 138: I don't think we need all these unknown flag tests.. they're basically just testing that commander works.

src/test/smoke/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    child.on('close', (exitCode) => {
+      resolve({
+        exitCode,
+        stdout,
+        stderr,
+        combined: stdout + stderr,
+      })
+    })
```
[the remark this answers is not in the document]
SgtPooki: we should at least add an error listener here.

SgtPooki (changes requested):

CodeByD3v:
Done:

1. runCli now accepts an env param (default {}) — callers can inject specific vars for PR 2 network tests. Auth keys are always stripped from the base env for hermeticity.
2. Error listener added — child.on('error', reject) so a spawn failure rejects immediately instead of hanging.
3. `combined` now tracks in real-time — built incrementally as data arrives from both streams, preserving actual interleaving order.
4. All unknown flag tests removed — agreed, they were just testing Commander.

Note on the env approach: went with inherit-then-strip rather than `env: {}` — a completely clean environment causes `--help` to hang because the `CLI` needs `HOME` for config file resolution on startup. The spirit is the same: auth vars never leak in, callers control what gets injected.

SgtPooki (approved):

PR #628, opened 2026-07-23
fix: make payment preflight deposit-aware

### LINKED ISSUE #616: fix: upload preflight and payments status reject wallets holding all USDFC as deposits
### Description

#560 fixed #557 for `payments setup`: the auto path now reads on-chain state first and only gates on wallet funding when the run will actually send a transaction (src/payments/auto.ts). Two call sites did not get the same treatment and still hard-fail on zero wallet USDFC before reading the deposit:

1. **Upload preflight** — `checkUploadReadiness` (src/core/upload/index.ts:133) calls `validatePaymentRequirements(filStatus.balance, walletUsdfcBalance, isCalibnet)` as step 1, before `validatePaymentCapacity` ever reads the account. A wallet holding all of its USDFC as deposits returns `status: 'blocked'` with `No USDFC tokens found`, and `add`/`import` cancel with "Payment setup incomplete" regardless of the deposited balance.
2. **`payments status`** — `showPaymentStatus` (src/payments/status.ts:112) cancels and throws on `walletUsdfcBalance === 0n`; `accountSummary` is not read until line 128. The diagnostic command the upload error tells the user to run is itself blocked by the same condition.

`validatePaymentRequirements` takes only `(filBalance, walletUsdfcBalance, isCalibnet)`, so no caller of it can account for deposited funds.

### Impact

A mainnet user hit this in July: ~36.7 USDFC deposited, ~34 available, years of runway at their data size, wallet USDFC 0 because they had deposited everything — exactly the account state the tooling encourages. `filecoin-pin add` refused to upload and `payments status` refused to run, while `payments setup` (fixed) correctly reported the healthy account. Fully funded accounts are locked out of uploads and diagnostics; the error message points at acquisition links for tokens the user does not need.

### Expected behavior

1. `checkUploadReadiness` should read on-chain state first and gate on wallet USDFC only when the upload will actually require a deposit, mirroring the #560 approach in `payments setup`. The deposit-aware `validatePaymentCapacity` check later in the same function already answers "can this account pay for the upload".
2. `payments status` is read-only and sends no transactions; it should never gate on wallet balances. Zero wallet USDFC is a display line, not an error.
3. When a wallet check does fail, print the numbers, per #557's convention.

beck-8:
## Summary

- allow upload readiness checks to use deposited Filecoin Pay funds when wallet USDFC is zero
- calculate funding plans before deciding whether wallet USDFC is required
- preserve FIL gas validation for deposit, withdrawal, and allowance transaction paths
- avoid configuring allowances before a funding plan has passed wallet validation

## Testing

- `pnpm run build`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run test:unit` (685 passed, 11 skipped)

Closes #616

### DIFF
--- a/src/core/payments/funding.ts
+++ b/src/core/payments/funding.ts
@@ -8,6 +8,7 @@ import {
   computeAdjustmentForExactDeposit,
   depositUSDFC,
   getPaymentStatus,
+  validateGasRequirement,
   validatePaymentRequirements,
   withdrawUSDFC,
 } from './index.js'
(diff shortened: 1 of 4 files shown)


SgtPooki (changes requested):
@beck-8 looks mostly good, but there were a few gaps I meant to push up to this PR but couldn't because its on your fork. check out #631 if you would

PR #631, opened 2026-07-28
fix: make payment preflight deposit-aware

### LINKED ISSUE #616: fix: upload preflight and payments status reject wallets holding all USDFC as deposits

SgtPooki:
Builds on #628 by @beck-8, whose commits are included here with authorship preserved, and adds the fixes that came out of its review.

## From #628

Upload preflight rejected wallets holding all their USDFC as deposits (#616). The preflight now validates gas only up front and lets the deposit-aware capacity check answer whether the account can pay for the upload. Funding plans keep gas validation before any transaction.

## Added in this PR

- **Fail fast for fully unfunded accounts.** A wallet with no USDFC anywhere (wallet or deposit) can never upload, so `checkUploadReadiness` blocks it with acquisition help before any allowance transaction spends gas. This also restores the `add` zero-size pre-flight, which otherwise passes vacuously and packs the whole CAR before failing.
- **Acquisition help when there is nothing to deposit.** When capacity is insufficient and the wallet holds no USDFC, the "Deposit at least X USDFC" suggestion now comes with links for acquiring USDFC.
- **`payments status` links gated on deposits.** Acquisition links only print when there is no USDFC anywhere. An account holding all its USDFC as deposits, the healthy state the tooling encourages, gets a clean status.
- **Regression tests** for each case above, including the first tests covering `showPaymentStatus`. All were verified to fail against the pre-fix code.

Closes #616
Supersedes #628

### DIFF
--- a/src/core/payments/funding.ts
+++ b/src/core/payments/funding.ts
@@ -8,6 +8,7 @@ import {
   computeAdjustmentForExactDeposit,
   depositUSDFC,
   getPaymentStatus,
+  validateGasRequirement,
   validatePaymentRequirements,
   withdrawUSDFC,
 } from './index.js'
@@ -427,15 +428,17 @@ export async function planFilecoinPayFunding(options: PlanFilecoinPayFundingOpti
     allowWithdraw,
   })
 
-  // Planning, no-op, and withdrawal paths do not spend wallet USDFC. Only
-  // validate wallet funding when this plan will actually make a deposit.
-  if (plan.delta > 0n) {
-    const isCalibnet = status.chainId === calibration.id
-    const validation = validatePaymentRequirements(status.filBalance, status.walletUsdfcBalance, isCalibnet)
-    if (!validation.isValid) {
-      const help = validation.helpMessage ? ` ${validation.helpMessage}` : ''
-      throw new Error(`${validation.errorMessage}${help}`)
-    }
+  // Every plan may be followed by a transaction (allowance, deposit, or
+  // withdrawal), so always validate FIL for gas. Only a deposit spends wallet
+  // USDFC, so retain that check exclusively for positive deltas.
+  const isCalibnet = status.chainId === calibration.id
+  const validation =
+    plan.delta > 0n
+      ? validatePaymentRequirements(status.filBalance, status.walletUsdfcBalance, isCalibnet)
+      : validateGasRequirement(status.filBalance, isCalibnet)
+  if (!validation.isValid) {
+    const help = validation.helpMessage ? ` ${validation.helpMessage}` : ''
+    throw new Error(`${validation.errorMessage}${help}`)
   }
 
   const allowances = ensureAllowances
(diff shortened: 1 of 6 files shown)


src/core/payments/funding.ts
```diff
@@ -367,10 +368,10 @@ export interface PlanFilecoinPayFundingOptions {
  *
  * This async function handles the full workflow:
  * - Fetches current payment status and SDK account summary in parallel
- * - Optionally ensures allowances are configured
- * - Validates payment requirements (FIL for gas, USDFC availability)
  * - Fetches pricing if needed
  * - Calculates funding plan via `calculateFilecoinPayFundingPlan`
+ * - Validates wallet funding only when the plan requires a deposit
```
beck-8:
```suggestion
 * - Validates wallet funding whenever the plan will send a transaction
```

I might want to change the comments here. Otherwise, it may be misinterpreted by AI tools later.

beck-8 (approved):

PR #630, opened 2026-07-27
feat(payments): support selected EVM funding sources

snissn:
## Summary

This draft adds optional source funding to `payments fund` and `payments setup --auto` on Filecoin mainnet.

Users choose one source network, one token, and a maximum source amount. New acquisitions support Filecoin, Arbitrum, Ethereum, Polygon, Avalanche, and BNB Chain. The token must resolve unambiguously in the current Squid catalog and return a valid route to the FIL or USDFC required by the command.

The existing Filecoin Pin funding calculations determine the wallet shortfall. The source-funding layer requests only that shortfall, validates the route and spending limits, waits for the assets to reach the Filecoin wallet, and then continues through the existing Filecoin Pay deposit and approval flow.

If the wallet already has enough FIL and USDFC, the direct path remains unchanged and does not request a quote or sign a source transaction. External source funding remains disabled on Calibration and local devnet.

Base and Optimism are not available for new acquisitions. Their total transaction fees include OP Stack fee components that are not bounded by the transaction-level gas fields used by the current hard native-spend ceiling. Earlier Base or Optimism checkpoints with already-broadcast transactions remain recoverable without quoting or signing another source transaction.

## Current status

The implementation work tracked by [#15](https://github.com/snissn/filecoin-pin/issues/15), [#16](https://github.com/snissn/filecoin-pin/issues/16), [#17](https://github.com/snissn/filecoin-pin/issues/17), and [#18](https://github.com/snissn/filecoin-pin/issues/18) is complete on the fork.

This includes:

- runtime resolution against Squid chain and token catalogs within the six-network boundary;
- native-token and ERC-20 sources with catalog-derived token decimals;
- chain-aware RPC, route, spender, expiry, quote, and spending-limit checks;
- proportional re-quoting when an initial quote exceeds the amount needed for the shortfall;
- explicit failure for provider minimums while preserving transient provider errors;
- Filecoin same-chain handling that preserves FIL needed for later transactions;
- recovery data tied to the exact wallet, source network, token, and recorded balance target;
- receipt and provider-status recovery without blind resubmission;
- bounded Filecoin balance polling after a recovered route is confirmed;
- cleanup of confirmed approval-only checkpoints after direct Filecoin funding satisfies the recorded target;
- integration with both `payments fund` and `payments setup --auto`;
- deterministic unit and mocked integration coverage for direct funding, acquisition, interruption, recovery, and rerun behavior;
- local-devnet deposit, approval, fail-closed, and rerun checks;
[...]

PR #637, opened 2026-07-29
feat(payments): add interactive Squid funding to payments fund

snissn:
## What changed

This adds an optional way for interactive `payments fund` to acquire the FIL needed for gas and the USDFC needed for a requested Filecoin Pay deposit.

- Users explicitly choose a source chain, token, RPC URL, and source-token limit.
- Supported source chains are Filecoin, Arbitrum, Ethereum, Base, Optimism, Polygon, Avalanche, and BNB Chain.
- Source tokens may be selected by `native`, address, or an unambiguous symbol in Squid's current catalog.
- Before signing, the CLI shows the quoted source spend, the source-token limit, and the buffered network-fee limit.
- After the tokens arrive on Filecoin, the existing payment planner and deposit flow continue unchanged.

Direct wallet funding remains the default. Squid is not contacted unless all four source options are supplied and the wallet has a positive FIL or USDFC shortfall. This path is limited to Filecoin Mainnet, interactive terminals, and owner private-key authentication. `payments setup --auto`, Calibration, and devnet are unchanged.

Provider route planning and transaction execution use the published [`@filecoin-project/squid-evm-funding` `v0.3.1` package](https://www.npmjs.com/package/@filecoin-project/squid-evm-funding/v/0.3.1). Filecoin Pin ships with the public Filecoin Squid integrator ID and supports `SQUID_INTEGRATOR_ID` as an optional override.

If a process stops after confirmation, a small pending marker blocks another acquisition attempt until the user verifies both chains and removes it.

This supersedes #630, which implemented the provider, automatic-setup integration, persistence framework, fixtures, and smoke-test tooling directly in Filecoin Pin.

## Checks

- Clean frozen installation of `@filecoin-project/squid-evm-funding@0.3.1`
- `pnpm run build`
- `pnpm test` — 755 passing, 11 skipped across 69 files

## Remaining validation

No funded mainnet route has been executed at this exact Filecoin Pin and library revision. A low-value capped run still needs to verify FIL and USDFC arrival, the Filecoin Pay deposit, and a rerun without a second acquisition.

Tracked in [snissn/filecoin-pin#31](https://github.com/snissn/filecoin-pin/issues/31).


snissn:
@codex review

snissn:
@codex review

snissn:
@codex review

snissn:
@codex review

Please review exact head `5b0563d69a6a33681a1fee352d90d1c2306b8d72`. This revision pins the already-reviewed library implementation to its merged `v0.1.0` release; all current-head CI checks are green.

src/payments/squid-funding.ts
```diff
@@ -0,0 +1,268 @@
...
+        slippage: SLIPPAGE_PERCENT,
+      },
+      squid
+    )
+    const sourceBalanceFloor =
+      policy.chain.id === filecoinMainnet.id && plan.source.token.toLowerCase() === FILECOIN_USDFC.toLowerCase()
+        ? input.requiredWalletUsdfc
+        : policy.chain.id === filecoinMainnet.id && plan.source.token === NATIVE_TOKEN_ADDRESS
+          ? MIN_FIL_FOR_GAS
+          : 0n
```
SgtPooki: no nested ternaries please

src/payments/squid-funding.ts
```diff
@@ -0,0 +1,268 @@
...
+    const nativeBalanceFloor = policy.chain.id === filecoinMainnet.id ? MIN_FIL_FOR_GAS : 0n
```
SgtPooki: `nativeBalanceFloor` is set to `MIN_FIL_FOR_GAS` (0.1 FIL) whenever the source chain is Filecoin. But `executeSquidFunding` enforces that floor as a precondition on the current balance before it sends the swap tx (execution.ts:266): nativeBalance >= fee + 0.1 FIL. 

The only way a FIL shortfall exists in the first place is filBalance < 0.1 FIL, so the check is false by construction. Anyone doing the "i have USDFC on Filecoin but no gas" bootstrap swap (--from-chain filecoin --from-token USDFC) gets "Native balance would not cover the fee and floor" 100% of the time.

README.md
```diff
@@ -205,6 +205,21 @@ filecoin-pin data-set <dataset-id>
 filecoin-pin add myfile.txt --network calibration
 ```
 
+### Fund from another EVM chain
+
+On Filecoin Mainnet, interactive `payments fund` can use [Squid](documentation/glossary.md#squid) when the owner wallet is short of FIL for gas or USDFC for the requested deposit. It supports Filecoin, Arbitrum, Ethereum, Base, Optimism, Polygon, Avalanche, and BNB Chain, and accepts `native`, a token address, or an unambiguous token symbol from Squid's current catalog.
+
+```bash
+export SQUID_INTEGRATOR_ID=your-integrator-id
```
SgtPooki: is this something most users will have? It seems unlikely.. are we supposed to set up a squid integrator id for use by filecoin-pin?

README.md
```diff
@@ -275,6 +290,9 @@ RPC_URL=wss://...              # Filecoin RPC endpoint (overrides NETWORK if spe
                                # Mainnet: wss://wss.node.glif.io/apigw/lotus/rpc/v1
                                # Calibration: wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v1
 
+# Optional - interactive Mainnet funding through Squid
+SQUID_INTEGRATOR_ID=...        # Squid-issued integration identifier
```
SgtPooki: ditto to above.. it seems like we need to actually go through the `Apply for an[ integrator ID](https://squidrouter.typeform.com/integrator-id?typeform-source=app.gitbook.com).` step from https://docs.squidrouter.com/getting-started/integrator-quickstart

package.json
```diff
@@ -152,6 +152,7 @@
     "picocolors": "^1.1.1",
     "pino": "^10.3.1",
     "semver": "^7.7.4",
+    "squid-evm-funding": "github:FilOzone/squid-evm-funding#v0.1.0",
```
SgtPooki: we should publish that package and import normally through npm

src/test/unit/squid-funding.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  afterEach(async () => {
+    for (const [name, value] of Object.entries(originalEnvironment)) {
+      const key =
+        name === 'databasePath'
+          ? 'DATABASE_PATH'
+          : name === 'integratorId'
+            ? 'SQUID_INTEGRATOR_ID'
+            : name === 'network'
+              ? 'NETWORK'
+              : 'RPC_URL'
```
SgtPooki: no nested ternaries please. I thought I had a biome rule enabled for this..

src/payments/squid-funding.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    await input.confirm({
+      source: plan.source,
+      sourceChainName: policy.chain.name,
+      quotes: plan.quotes,
+      maxSourceAmount: plan.maxSourceAmount,
+      maxNativeFee: policy.maxNativeFee,
+      nativeCurrency: policy.chain.nativeCurrency,
+    })
```
SgtPooki, line 251: I think we should also allow an optional amount of headroom for the price to drift.. otherwise the users could really fight with trying to get things to pass on a slight price drift.. though we may want that.

src/test/unit/squid-funding.test.ts
```diff
@@ -0,0 +1,261 @@
...
+      maxSourceAmount: '10000000',
+    })
+  })
+
+  it('uses atomic marker creation to block an ambiguous rerun', async () => {
+    await writeFile(marker, '{}', { mode: 0o600 })
+
+    await expect(acquirePaymentShortfalls(input())).rejects.toThrow(marker)
+    expect(mockExecute).not.toHaveBeenCalled()
+  })
```
SgtPooki, line 217: this pre-plants a marker and asserts the run rejects and `mockExecute` isn't called, but it never asserts `confirm` wasn't called, but it looks like it is called: confirm fires before the marker write, so this test passes while the user still sits through a spend prompt for a run that's doomed.

src/test/unit/run-fund-squid.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    const values = [options.fromChain, options.fromToken, options.maxSourceAmount, options.sourceRpcUrl]
+    const requested = values.some((value) => value != null && value.trim() !== '')
+    if (requested && values.some((value) => !value?.trim())) {
+      throw new Error(
+        'Source acquisition requires --from-chain, --from-token, --max-source-amount, and --source-rpc-url together'
+      )
+    }
+    return requested
+  },
+}))
```
SgtPooki: we should importActual where possible so we can test the actual implementations.


```suggestion
vi.mock('../../payments/squid-funding.js', async (importOriginal) => {
  const actual = await importOriginal<typeof import('../../payments/squid-funding.js')>()
  return { ...actual, acquirePaymentShortfalls: mockAcquire }
})
```

src/test/unit/squid-funding.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    for (const secret of [PRIVATE_KEY, privateKey, rpcUrl, integratorId]) {
+      expect(serialized).not.toContain(secret)
+    }
+    if (process.platform !== 'win32') expect((await stat(marker)).mode & 0o777).toBe(0o600)
+    expect(JSON.parse(await readFile(marker, 'utf8'))).toMatchObject({
+      owner: OWNER,
+      sourceChain: 42161,
+      maxSourceAmount: '10000000',
+    })
+  })
```
SgtPooki, line 192: this test asserts the marker survives, but since the library is fully mocked, it can't distinguish "failed before any broadcast" from "failed mid-flight".. I think we need a test that would catch a problem I call out in squid-funding.ts:

```
it('checks the pending marker before asking the user to confirm', async () => {
  await writeFile(marker, '{}', { mode: 0o600 })
  const request = input() // hold the reference so we can inspect the spy

  await expect(acquirePaymentShortfalls(request)).rejects.toThrow(marker)
  expect(request.confirm).not.toHaveBeenCalled() // ← FAILS today: confirm was called once
})
```

src/payments/squid-funding.ts
```diff
@@ -0,0 +1,268 @@
...
+            token: requirement.token,
+            amount: requirement.amount.toString(),
+          })),
+        }),
+        { encoding: 'utf8', flag: 'wx', mode: 0o600 }
+      )
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code === 'EEXIST') throw pendingError(path)
+      throw error
+    }
```
SgtPooki: This marker check runs too late. The order today is planSquidFunding->input.confirm->writeFile/EEXIST, which means the pending-marker guard fires after the user already approved the spend.

Three problems with that:

1. The confirm becomes a lie. The user approves "spend X USDC to receive Y FIL" and then the run refuses. A yes on a money prompt should mean "this may now happen".
2. The amounts the user approved were computed from state the marker says cant be trusted. The marker exists because a previous attempt may still be in flight.. funds from that bridge could land on Filecoin at any moment, but the shortfall preview and quotes were computed as if nothing is inbound. The check exists to prevent exactly this, and it runs after the damage.
3. Wasted live calls. planSquidFunding hits the Squid API (token catalog + route quotes with the convergence loop) for a run that a single stat() on a local file would have killed instantly.

src/payments/squid-funding.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    await executeSquidFunding(
```
SgtPooki, line 253: i'm not sure what timeframes are, but we checked the balance way above and may benefit from checking again? leaving to your expertise after diving in here.

src/payments/fund.ts
```diff
@@ -265,14 +270,50 @@ export async function runFund(options: FundOptions): Promise<void> {
...
+    spinner.start('Calculating funding plan...')
+    if (sourceRequested) {
+      const [status, accountSummary] = await Promise.all([
+        getPaymentStatus(synapse),
+        synapse.payments.accountSummary({}),
+      ])
+      const preview = calculateFilecoinPayFundingPlan({ status, accountSummary, ...planOptions })
+      const filShortfall =
+        preview.delta !== 0n && status.filBalance < MIN_FIL_FOR_GAS ? MIN_FIL_FOR_GAS - status.filBalance : 0n
+      const usdfcShortfall = preview.delta > 0n ? (preview.walletShortfall ?? 0n) : 0n
```
SgtPooki: should we add a small buffer to usdfcShortfall? 

Thinking about timeframes here where epochs passing is actually reasonable:


- Human part: reading the confirm and deciding — call it ≤5 minutes for anyone actually interacting.
- Machine part: quote refresh + approve tx + swap tx + cross-chain settle polling, capped at 120 × 5s = 10 min per quote, two quotes (FIL + USDFC) worst case ≈ 20 min.

That's ~25 minutes end-to-end at the documented poll limits. Round up to 1 hour for 2× margin and you get a clean constant: 120 epochs at 30s each.

Magnitude check: one hour of spend = perDay / 24 ≈ 0.14% of a 30-day target — for a 10 USDFC top-up that's about 0.014 USDFC of headroom. Effectively invisible to the user, provably sufficient for the whole window the feature is designed to operate in, and the overage stays in the wallet as spendable USDFC since the deposit still uses the replanned delta.

```suggestion
      const ACQUISITION_BUFFER_EPOCHS = 120n // ~1h: worst-case dual-quote settle (~25m) with margin
      const usdfcShortfall = preview.delta > 0n ? (preview.walletShortfall ?? 0n) + rateUsed * ACQUISITION_BUFFER_EPOCHS : 0n
```

src/payments/fund.ts
```diff
@@ -265,14 +270,50 @@ export async function runFund(options: FundOptions): Promise<void> {
...
+
+    spinner.start('Calculating funding plan...')
+    if (sourceRequested) {
+      const [status, accountSummary] = await Promise.all([
+        getPaymentStatus(synapse),
+        synapse.payments.accountSummary({}),
+      ])
+      const preview = calculateFilecoinPayFundingPlan({ status, accountSummary, ...planOptions })
+      const filShortfall =
+        preview.delta !== 0n && status.filBalance < MIN_FIL_FOR_GAS ? MIN_FIL_FOR_GAS - status.filBalance : 0n
```
SgtPooki: if someone is withdrawing from their FilecoinPay balance, they may need FIL for gas, so its reasonable to fire for withdraws.. but it may be unexpected based on current readme and documentation?

maybe we should just restrict to `preview.delta > 0n` (deposits) for now?

SgtPooki (changes requested):

src/payments/squid-funding.ts
snissn: Addressed in [1463199](https://github.com/filecoin-project/filecoin-pin/commit/14631999543327f11ca84693cdcc9e3250794adf): replaced the production nested ternary with explicit Filecoin USDFC and native FIL branches.

src/payments/squid-funding.ts
snissn: Addressed in [207a291](https://github.com/filecoin-project/filecoin-pin/commit/207a29107f1914d07e3a86713bd991217ded89bf): Filecoin USDFC to FIL acquisition no longer requires the 0.1 FIL floor before execution. The dependency still checks the prepared transaction fee, the acquired FIL includes the Filecoin fee cap, and reserved USDFC remains protected. Regression coverage includes this bootstrap path and the separate native FIL reserve path.

src/test/unit/squid-funding.test.ts
snissn: Addressed in [5daad40](https://github.com/filecoin-project/filecoin-pin/commit/5daad40b4720a244019ce1f60b5d752b6345ca4f): replaced the test nested ternary with a direct environment snapshot and restore loop.

src/test/unit/squid-funding.test.ts
snissn: Addressed in [a17ebe5](https://github.com/filecoin-project/filecoin-pin/commit/a17ebe528d29b48eadbfe7ec2ab088ef6e9d9b65): the existing-marker test now retains the request and explicitly asserts that confirmation is skipped.

src/test/unit/run-fund-squid.test.ts
snissn: Addressed in [c588bda](https://github.com/filecoin-project/filecoin-pin/commit/c588bda52bdff63d79d9602bea8203a6b5a1b767): the unit test now uses `importOriginal` and overrides only `acquirePaymentShortfalls`, so the real module implementation remains under test.

src/test/unit/squid-funding.test.ts
snissn: Addressed in [047df3c](https://github.com/filecoin-project/filecoin-pin/commit/047df3c14800cb5411d185cf569524da6b08b4e9): failures before entering execution remove the marker, while failures after entering `executeSquidFunding` preserve it because broadcast status may be ambiguous. The tests cover both confirmation failure and execution failure.

src/payments/squid-funding.ts
snissn: Addressed in [8586219](https://github.com/filecoin-project/filecoin-pin/commit/8586219069d4d48539516eb645a38c956238f1e0): atomic marker creation now happens before client creation, Squid planning, and confirmation. An existing marker prevents both API calls and the spend prompt.

src/payments/squid-funding.ts
snissn: Confirmed in the pinned dependency commit [b89293a](https://github.com/FilOzone/squid-evm-funding/commit/b89293a964cc233909cac60899c9c1a0bc7a6df9), specifically the [per-transaction execution guard](https://github.com/FilOzone/squid-evm-funding/blob/b89293a964cc233909cac60899c9c1a0bc7a6df9/src/execution.ts#L245-L275): native and source balances are reread after transaction preparation and immediately before the pending-nonce and broadcast checks. I kept that authoritative late check instead of duplicating an earlier, weaker snapshot in this adapter.

src/payments/fund.ts
snissn: Addressed in [aca9bab](https://github.com/filecoin-project/filecoin-pin/commit/aca9bab3b9f9718cc5ea07fc32deacaaf879f8e2): required wallet USDFC now includes one hour of spend using the SDK `TIME_CONSTANTS.EPOCHS_PER_HOUR`, and the shortfall is calculated from that buffered requirement. The deposit is still replanned after acquisition.

src/payments/fund.ts
snissn: Addressed in [b210af7](https://github.com/filecoin-project/filecoin-pin/commit/b210af725d65a1190e6fb25b4620d96d972c6784): Squid acquisition now runs only when `preview.delta > 0n`, with a regression test showing withdrawals do not enter the acquisition adapter.

src/payments/squid-funding.ts
snissn: Fixed in [8d4a2c8](https://github.com/filecoin-project/filecoin-pin/pull/637/commits/8d4a2c800b3fa3ea20f0419faa42e1958a408335). Filecoin Pin now adds up to 1% source-amount headroom, clipped by the user-approved --max-source-amount cap. Execution refreshes each Squid route with that buffered source amount while retaining the original FIL or USDFC destination requirement. A focused tight-cap test covers both the buffer and the bound. Validation: pnpm test (69 files, 754 passed, 11 skipped).

snissn:
I think the unresolved question here is how to handle the squid integrator id. This context is for a CLI, so each user would be running it separately. It definitely doesn't make sense for each user to have their own squid id. I think we need to decide one of these two options, with a) possibly being invalid and b) being my preference.

a) we are OK with our squid application id being in the repo/publicaly accessible and users simply use it (Dependent on seeing the TOS and expectations of the squid id)
b) we have the CLI connect to the filecoin pay explorer service which handles the 'squid id required' api calls for them  - relevant PR https://github.com/FilOzone/filecoin-pay-explorer/pull/320

My immediate follow up is to look into the squid api key / id details and see if it's meant to be a public facing application ID or a private API "key"

snissn:
OK so it looks like it is not a secret and is an identifier more like a username

  > Follow-up: Squid’s integrator ID is intended as a public project/integration identifier, not a secret credential. Their official browser widget embeds it directly in client-side configuration, and their API permits direct cross-origin requests carrying x-integrator-
  > id. The registration form also scopes it to the integrating project rather than individual users.
  >
  > I think the clean solution is therefore to register or confirm a production ID for Filecoin Pin, ship it as the CLI default, and retain SQUID_INTEGRATOR_ID only as an optional override. We should confirm with Squid that distributed CLI use is covered by the
  > registration and terms before embedding it. A proxy through Explorer should not be necessary.

I personally went through the steps here https://docs.squidrouter.com/getting-started/integrator-quickstart for an identifier labeled `filecoin-testing-{snip}` and got a reply with an id very shortly after the request

README.md
snissn: it looks like we want a new ID associated with this app and it's a public identifier

snissn:
OK Great so since it's a public identifier, i'll update the PR to use the one I generated, and if we want to we can update it to different key than `filecoin-testing-94a4a25a-d40b-41cb-b148-e96098862`

snissn:
Added the public Filecoin Squid integrator ID as the default in [`aaeec19`](https://github.com/filecoin-project/filecoin-pin/pull/637/commits/aaeec1990a85761ad599b5b4e0bd3fe5bf361352); `SQUID_INTEGRATOR_ID` remains an optional override. Build and all tests pass (755 passed, 11 skipped).

README.md
snissn: Resolved in [`aaeec19`](https://github.com/filecoin-project/filecoin-pin/pull/637/commits/aaeec1990a85761ad599b5b4e0bd3fe5bf361352): Filecoin Pin now supplies the registered public ID by default; users only need the optional override.

README.md
snissn: Resolved in [`aaeec19`](https://github.com/filecoin-project/filecoin-pin/pull/637/commits/aaeec1990a85761ad599b5b4e0bd3fe5bf361352): the registered public ID is now the documented default.

snissn:
Updated this PR to the published `@filecoin-project/squid-evm-funding@0.2.0` release in [4dc8229](https://github.com/snissn/filecoin-pin/commit/4dc8229a12ca7b1c939b7584e5e6e99d58431b95). The CLI now consumes the price-only plan type, and its fixture enforces that contract. Validation passes: full test suite (69 files, 755 passed, 11 skipped), build, lint, type-check, focused Squid tests (13/13), and diff checks. Package release: https://github.com/FilOzone/squid-evm-funding/releases/tag/v0.2.0

snissn:
Updated the CLI to [`@filecoin-project/squid-evm-funding@0.2.1`](https://github.com/FilOzone/squid-evm-funding/releases/tag/v0.2.1) in [`75b906a`](https://github.com/snissn/filecoin-pin/commit/75b906a1cd871e553a0503e40b934d215b9c304a). This release retains the best feasible quote when a later refinement quote slips just below the target, avoiding the spurious “could not converge” result caused by quote-price jitter.

Local validation passed: frozen install, build, 69 test files (755 passed, 11 skipped), type-check, lint, and diff checks. Independent review found no issues.

snissn:
Updated to [`@filecoin-project/squid-evm-funding@0.2.2`](https://github.com/FilOzone/squid-evm-funding/releases/tag/v0.2.2) in [`87d73f8`](https://github.com/snissn/filecoin-pin/commit/87d73f89cd7764847b45c629d864afad51fb7403). The release adds 1% planning headroom after an insufficient price quote, accepts the first sufficient follow-up quote, and retains an eight-call fallback. Local build, tests, lint, and type-check pass.

snissn:
Updated the CLI to [`@filecoin-project/squid-evm-funding@0.3.1`](https://www.npmjs.com/package/@filecoin-project/squid-evm-funding/v/0.3.1) in [`44ba287`](https://github.com/snissn/filecoin-pin/commit/44ba287). I also made Squid source-chain native route fees explicit in the confirmation and clarified the separate transaction-gas cap in [`b8659f3`](https://github.com/snissn/filecoin-pin/commit/b8659f3). Local validation passed: build, all 69 test files (755 passed, 11 skipped), and two independent local reviews.

package.json
```diff
@@ -152,6 +152,7 @@
     "picocolors": "^1.1.1",
     "pino": "^10.3.1",
     "semver": "^7.7.4",
+    "@filecoin-project/squid-evm-funding": "^0.3.1",
```
SgtPooki: Pin 0.3.3 exactly. 0.3.2 fixed real-route trust checks for ERC-20 sources and 0.3.3 made `maxTotalNativeRouteFee` required, so a caret here breaks typecheck on the next lockfile regen (I hit this rebasing onto master). Pass `maximumNativeRouteFee(reviewedFee)` and show that number in the confirm instead of the estimate. Once squid-evm-funding#27 lands, `instanceof SquidExecutionError` can replace `executionStarted` for the marker decision.

src/payments/squid-funding.ts
```diff
@@ -0,0 +1,287 @@
...
+import { publicActionsL2 } from 'viem/op-stack'
+import { CliIncomplete } from '../common/cli-errors.js'
+import { createConfig } from '../config.js'
+import { MIN_FIL_FOR_GAS } from '../core/payments/index.js'
+import { mainnet as filecoinMainnet } from '../core/synapse/index.js'
+import type { CLIAuthOptions } from '../utils/cli-auth.js'
+import type { FundingSourceOptions } from './types.js'
+
+const SQUID_ROUTER = getAddress('0xce16F69375520ab01377ce7B88f5BA8C48F8D666')
+const DEFAULT_SQUID_INTEGRATOR_ID = 'filecoin-testing-94a4a25a-d40b-41cb-b148-e96098862'
```
SgtPooki, line 26: Is this `filecoin-testing-*` the final id? Can we get a "production" one?

src/test/unit/run-fund-squid.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    mockPlan: vi.fn(),
+  }))
+
+vi.mock('@clack/prompts', () => ({ confirm: mockConfirm, isCancel: vi.fn(() => false) }))
+vi.mock('../../core/synapse/index.js', () => ({
+  getClientAddress: vi.fn(() => '0x1111111111111111111111111111111111111111'),
+  initializeSynapse: mockInitialize,
+  mainnet: { id: 314, contracts: { usdfc: { address: '0x80B98d3aa09ffff255c3ba4A241111Ff1262F045' } } },
+}))
+vi.mock('../../utils/cli-auth.js', () => ({
```
SgtPooki, line 21: default branch `runFund` now calls `assertOwnerAuth`. we should add it to this mock or all five tests here fail after rebase. The adapter's own walletAddress/sessionKey/viewAddress guard is redundant with it now.. keep or drop, but trim the "owns Mainnet and private-key authentication checks" test if it goes.

SgtPooki (changes requested):
Needs a rebase (lockfile conflicts, and master's `assertOwnerAuth` breaks the run-fund-squid mocks), and a version bump. filozone/squid-evm-funding#38 says 0.3.1 fails the trust checks on real ERC-20 routes, which is the README example. Details inline.

PR #652, opened 2026-08-12
feat: add migrate command

SgtPooki:
> [!IMPORTANT]
> **Superseded by a stacked split for review.** This PR stays open as the umbrella: full pipeline description below, plus the pending calibnet validation. Review and merge the stack in order instead of this branch:
>
> 1. #653 state store and CLI parsers
> 2. #654 gateway fetch and CAR verify (also fixes `@ipld/dag-cbor` shipped as a devDependency)
> 3. #655 piece packing and GC windows
> 4. #656 batched upload with resume
> 5. #657 command wiring, pipeline, docs
>
> Each PR's base is the previous branch; the command activates in #657. Close this PR once #657 merges.

### What changed

Adds `filecoin-pin migrate <cid-list-file>`: bulk-migrate IPFS CIDs onto FOC in batched pieces. Refs #651. Ported from FilOzone/foc-migrate@e52218a, then reworked into the pipeline documented in [documentation/migrate.md](https://github.com/filecoin-project/filecoin-pin/blob/feat/migrate/documentation/migrate.md) after three rounds of multi-agent review.

One pipeline, clocked by a disk budget: each CID downloads once into a verified member CAR on disk (every block hash-checked, DAG walked for completeness from the requested root, CommP computed in the same pass), members bin-pack into ~1000MiB multi-root pieces assembled disk-to-disk in constant memory, and uploads run concurrently with GC-aware batched `addPieces` commits (cap 40). Committed pieces evict and free budget, so a disk smaller than the migration cycles instead of failing; `--max-staged-bytes` caps it explicitly.

Resume is receipt-first from a per-network, per-owner `migrate.db`: unknown-outcome commits resolve via the `PiecesAdded` event (or `findPieceIdsByCid` when no tx hash was captured), staged files re-verify against recorded hashes on startup, and a blind on-chain re-add is never issued. Egress defaults to `none`.

### How to verify

```
pnpm install && pnpm run build && pnpm test
node dist/cli.js migrate --help
```

A calibnet validation run (kill mid-run, resume, confirm no duplicate adds) is still pending before undraft.

### Notes / risks

- `node:sqlite` (`DatabaseSync`, WAL) is unflagged since Node 23.4; the `lts/*` CI lane covers it. No `engines` change.
- Concurrent runs against one `migrate.db` are unsupported.
- New dependency: `@ipld/dag-pb` (already in the tree transitively).
[...]

PR #653, opened 2026-08-14
feat: add migrate state store and CLI parsers

SgtPooki:
### What changed

Part 1 of 5 splitting #652 into reviewable slices. Adds the migrate command's SQLite resume state store (`migrate.db` schema and lifecycle transitions: source CIDs go pending, done, failed or oversized; packed pieces are built or failed; each per-provider upload goes parked, add_unconfirmed, committed, collected or failed), trustless-gateway CAR URL construction, the size/int flag parsers (shared in `src/utils/cli-helpers.ts` next to `formatFileSize`), and the CID-list parser. Uses `node:sqlite` (`DatabaseSync`, WAL), unflagged since Node 23.4.

Also carries [documentation/migrate.md](https://github.com/filecoin-project/filecoin-pin/blob/feat/migrate-1-state-store/documentation/migrate.md) and the migrate glossary entries: the design doc rides in the first slice so the pipeline, resume guarantees, and exit contract get agreement before the code that implements them. Start the review there.

No CLI wiring yet.. the command activates in part 5, so this is dead code until the stack lands. Code review focus: the schema and the state transitions, they are the resume model everything else leans on.

### How to verify

```
pnpm install && pnpm run build && pnpm test
```

### Notes / risks

Stack: merge order is 1 through 5; each PR's base is the previous branch. The original umbrella #652 is closed, this slice carries the design doc and the full pipeline description lives in [documentation/migrate.md](https://github.com/filecoin-project/filecoin-pin/blob/feat/migrate-1-state-store/documentation/migrate.md). A calibnet kill-and-resume validation (no duplicate adds) is still pending before #657 undrafts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ test-car-output/
 test-e2e-cars/
 test-simple-car-output/
 coverage
+
+# Local scratch artifacts
+.lavish/
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,3 @@ test-car-output/
 test-e2e-cars/
 test-simple-car-output/
 coverage
-
-# Local scratch artifacts
-.lavish/
--- a/documentation/glossary.md
+++ b/documentation/glossary.md
@@ -110,6 +110,10 @@ See [Management Console affordance in the README](../README.md#-management-conso
 
 This is the primary smart contract used when interacting with the warm storage functionality offered in [Filecoin Onchain Cloud](#filecoin-onchain-cloud).  It acts as both a "service" contract and a "validator" contract for payment management and settlements, ensuring the warm storage service is actually delivered before payments are released to the [Service Provider](#service-provider).
 
+## GC Window
+
+The time a [Service Provider](#service-provider)'s [Curio](#curio) instance keeps a [Parked Piece](#parked-piece) before garbage-collecting it (roughly 2 hours by default, provider-configurable, not discoverable on chain). The `migrate` command schedules its batched `addPieces` commits to land before its per-provider estimate of this window expires, and lowers the estimate when it observes a GC rejection.
+
 ## IPFS Root CID
 
 The CID for the root of a merkle DAG that is usually encoding a file or directory as UnixFS.  Since each `filecoin-pin add` creates a [CAR](#car), regardless if passed a file or directory, there is a single root corresponding to root of the Merkle DAG made out of encoding the file or directory as UnixFS. Typically this will be presented in base32, beginning with `bafy` and be 59 characters long. See [Relationship between Piece CID and IPFS Root CID](#relationship-between-piece-cid-and-ipfs-root-cid) for how this relates to the [Piece CID](#piece-cid).
@@ -136,6 +140,14 @@ Key | Purpose | Scope
 `withIPFSIndexing` | Set to empty string to signal the [SP](#service-provider) to index and advertise the data to [IPNI](#ipni) | Data Set
 `ipfsRootCid` | Stored on each Piece to link the [Piece CID](#piece-cid) back to the [IPFS Root CID](#ipfs-root-cid).  While this is a convention that Filecoin Pin follows, there is nothing onchain enforcing a correct link between `ipfsRootCid` and `pieceCid`. | Piece
 `name` | Original basename of the source path (file or directory). Auto-derived during `add` so the human-readable label survives even though the [UnixFS profile](#unixfs-v1-2025-profile) does not wrap single files in a parent directory. User-supplied piece metadata wins over the auto-derived value; an explicit empty string is treated as an opt-out. Consumers that need to know whether the source was a file or a directory inspect the [IPFS Root CID](#ipfs-root-cid) (codec + UnixFS `Data.Type`), matching the IPFS Pinning Service `name` convention. | Piece
+
+## Member CAR
+
+A single source [CID](#cid)'s verified [CAR](#car), downloaded from a trustless gateway by the `migrate` command and staged on disk. Members are bin-packed into one multi-root CAR that becomes a [Piece](#piece); the member file is deleted once the packed piece records it. See [How migrate works](migrate.md).
+
+## Parked Piece
+
+A [Piece](#piece) stored on a [Service Provider](#service-provider) but not yet committed on chain via `addPieces`. Parking starts the provider's [GC Window](#gc-window) clock: a piece parked past the window is garbage-collected and must be stored again.
 
 ## unixfs-v1-2025 profile
 
@@ -238,6 +250,10 @@ Session keys require specific permissions (such as CREATE_DATA_SET and ADD_PIECE
 Note that the filecoin-pin CLI's `--session-key` flag (and `SESSION_KEY` environment variable) expect the session key's **private key** — the `SESSION_KEY` value printed by `filecoin-pin session create` or `filecoin-pin session generate` — not the session address. The (public) session address is only used when authorizing or revoking: `filecoin-pin session authorize <session-address>` and `filecoin-pin session revoke <session-address>`.
 
 
+
+## Staging Budget
+
+The byte budget the `migrate` command grants its staging directory (free space at startup, capped by `--max-staged-bytes`). Everything staged counts against it: [Member CARs](#member-car), in-flight downloads, assembling pieces, and packed pieces not yet committed. A full budget blocks downloads until commits evict pieces, so a disk smaller than the migration cycles instead of failing. See [How migrate works](migrate.md).
 
 ## Standard IPFS Tooling
 
--- a/documentation/migrate.md
+++ b/documentation/migrate.md
@@ -52,4 +52,4 @@ Egress defaults to `none` for migrate: bulk archives will not pay [FilBeam](glos
 
 ## What the exit code means
 
-`0` means every CID in the list is committed on chain at the requested number of copies. Anything short of that (failed downloads, over-cap CIDs, unresolved commits, missing copies) exits `1`, and the summary JSON on stdout says exactly which CIDs need attention. Re-running is always safe.
+`0` means every CID in the list is committed on chain at the requested number of copies. Anything short of that (failed downloads, over-cap CIDs, unresolved commits, missing copies) exits `1`, and the summary printed at the end says exactly which CIDs need attention (the full record lives in `migrate.db`). Re-running is always safe.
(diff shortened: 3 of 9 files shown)


.gitignore
```diff
@@ -12,3 +12,6 @@ test-car-output/
 test-e2e-cars/
 test-simple-car-output/
 coverage
+
+# Local scratch artifacts
+.lavish/
```
SgtPooki:
```suggestion
```

documentation/glossary.md
```diff
@@ -239,6 +251,10 @@ Note that the filecoin-pin CLI's `--session-key` flag (and `SESSION_KEY` environ
 
 
 
+## Staging Budget
+
+The byte budget the `migrate` command grants its staging directory (free space at startup, capped by `--max-staged-bytes`). Everything staged counts against it: [Member CARs](#member-car), in-flight downloads, assembling pieces, and packed pieces not yet committed. A full budget blocks downloads until commits evict pieces, so a disk smaller than the migration cycles instead of failing. See [How migrate works](migrate.md).
```
SgtPooki, line 256: idk if we need this in the glossary.

documentation/migrate.md
```diff
@@ -0,0 +1,55 @@
+# How migrate works
+
+`filecoin-pin migrate <cid-list-file>` moves a list of IPFS CIDs onto [Filecoin Onchain Cloud](glossary.md#filecoin-onchain-cloud). You give it a text file with one [CID](glossary.md#cid) per line; it downloads each one from a trustless gateway, packs them into large [pieces](glossary.md#piece), uploads those pieces to [storage providers](glossary.md#storage-provider), and commits them on chain in batches. Every step is recorded in a local sqlite database, so a killed or crashed run resumes where it left off.
+
+This page explains the pipeline and the guarantees behind it. For flags and defaults, run `filecoin-pin migrate --help`.
+
+## Why not just run `add` in a loop
+
+Two reasons: cost and throughput. Each on-chain `addPieces` transaction can carry up to 40 pieces, and each piece can hold up to 1016 MiB. Uploading a 2 MiB CID as its own piece wastes both dimensions. Migrate packs many source CIDs into one ~1000 MiB multi-root [CAR](glossary.md#car) and commits up to 40 of those pieces per transaction, so a million small files does not mean a million transactions.
```
SgtPooki, line 9: 40 pieces cap is changing, this week.

SgtPooki (commented):
self review

documentation/migrate.md
hugomrdias: it may not we are keeping 40 piece limit until curio supports it

documentation/glossary.md
hugomrdias: i dont think we need this

documentation/migrate.md
```diff
@@ -0,0 +1,55 @@
+# How migrate works
+
+`filecoin-pin migrate <cid-list-file>` moves a list of IPFS CIDs onto [Filecoin Onchain Cloud](glossary.md#filecoin-onchain-cloud). You give it a text file with one [CID](glossary.md#cid) per line; it downloads each one from a trustless gateway, packs them into large [pieces](glossary.md#piece), uploads those pieces to [storage providers](glossary.md#storage-provider), and commits them on chain in batches. Every step is recorded in a local sqlite database, so a killed or crashed run resumes where it left off.
```
hugomrdias, line 3: i think we need to explain its an aggregated piece, a car file with pieces inside.

hugomrdias (changes requested):

src/migrate/db.ts
```diff
@@ -0,0 +1,785 @@
...
+    this.#db
+      .prepare(
+        `INSERT INTO uploads (scope, sub_piece_cid, provider_id, role, data_set_id, status, parked_at, updated_at)
+         VALUES (?, ?, ?, ?, ?, 'parked', ?, ?)
+         ON CONFLICT (scope, sub_piece_cid, provider_id) DO UPDATE SET
+           role = excluded.role, data_set_id = excluded.data_set_id,
+           status = 'parked', parked_at = excluded.parked_at,
+           tx_hash = NULL, piece_id = NULL, committed_at = NULL, error = NULL,
+           updated_at = excluded.updated_at
+         WHERE uploads.status != 'committed'`
```
hugomrdias, line 584: **[P1] Preserve unresolved transactions when re-parking an upload**

`recordUploadParked()` protects `committed` rows but overwrites `add_unconfirmed`, clearing the transaction hash. This is reachable in the later upload slice when resume restores a collected primary while the secondary still has an unresolved transaction: the subsequent secondary pull re-parks the row, and the next flush submits another `addPieces` despite reconciliation having left the original transaction unresolved. Preserve `add_unconfirmed` here and require explicit reconciliation before returning it to `parked`.

Reproduced with the later upload loop from #657 and this PR's unchanged database implementation: the test issues an unexpected secondary commit while the original transaction remains unresolved.

src/migrate/db.ts
```diff
@@ -0,0 +1,785 @@
...
+   * so the failure record is the first thing written for that (piece, provider).
+   */
+  markUploadFailed(subPieceCid: string, providerId: string, role: UploadRole, error: string): void {
+    const now = new Date().toISOString()
+    this.#db
+      .prepare(
+        `INSERT INTO uploads (scope, sub_piece_cid, provider_id, role, status, parked_at, error, updated_at)
+         VALUES (?, ?, ?, ?, 'failed', ?, ?, ?)
+         ON CONFLICT (scope, sub_piece_cid, provider_id) DO UPDATE SET
+           status = 'failed', error = excluded.error, updated_at = excluded.updated_at`
```
hugomrdias, line 677: **[P1] Preserve committed and unresolved uploads when a pull fails**

`markUploadFailed()` unconditionally replaces an existing status with `failed`. If the primary was collected while the secondary already committed, restoring the primary triggers a redundant secondary pull. A transient failure demotes the committed row; a successful retry then re-parks and commits the same piece to the same data set again. Guard the upsert against overwriting `committed` and `add_unconfirmed` rows so a transfer error cannot discard the on-chain recovery state.

Reproduced with the later upload loop from #657 and this PR's unchanged database implementation: after a transient secondary pull failure, the test issues another secondary commit despite that copy already being committed.

hugomrdias (commented):
Two additional findings

.gitignore
SgtPooki: Dropped in f5f2c5d; master now ignores `.lavish/` under "local tool artifacts" anyway.

documentation/glossary.md
SgtPooki: Removed the Staging Budget entry in 3df3f81. The disk budget section in migrate.md covers it.

documentation/migrate.md
SgtPooki: Left the doc at 40 since Hugo says the cap stays until Curio supports more. The code reads `SIZE_CONSTANTS.MAX_ADD_PIECES_BATCH_SIZE` from the SDK, so a bump there needs no change here.

src/migrate/db.ts
```diff
@@ -0,0 +1,785 @@
...
+    this.path = path
+    this.scope = scope
+    this.#db = new DatabaseSync(path)
+    // busy_timeout must come before any other exec: even setting WAL needs
+    // the write lock, and a sibling process holding it will error out the
+    // very first pragma without this. Without busy_timeout, the sqlite layer
+    // surfaces SQLITE_BUSY as `disk I/O error`.
+    this.#db.exec('PRAGMA busy_timeout = 5000')
+    this.#db.exec('PRAGMA journal_mode = WAL')
+    this.#migrate()
```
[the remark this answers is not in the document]
SgtPooki: node:sqlite opens with `enableForeignKeyConstraints: true`, so the FKs are enforced without the pragma. Added a test that a member row for an unregistered CID throws `FOREIGN KEY` (bdda19f), and a comment in the constructor saying so.

src/migrate/db.ts
```diff
@@ -0,0 +1,785 @@
...
+        `SELECT sp.sub_piece_cid, sp.assembled_car_length, sp.assembled_sha256,
+                sp.target_size_bytes, sp.car_path, sp.status
+         FROM sub_pieces sp
+         WHERE sp.scope = ? AND sp.status = 'built'
+           AND NOT EXISTS (
+             SELECT 1 FROM uploads u
+             WHERE u.scope = sp.scope
+               AND u.sub_piece_cid = sp.sub_piece_cid
+               AND u.provider_id = ?
+               AND u.status IN ('parked', 'add_unconfirmed', 'committed', 'failed')
```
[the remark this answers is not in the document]
SgtPooki: Intended, the docstring was wrong. A failed primary is a commP mismatch: storing the same bytes again fails the same way, so it stays out. Failed secondaries are retried by the retry pass in #656, not by this queue. Collected rows do come back. Docstring now says that (f00a53a).

src/utils/cli-helpers.ts
```diff
@@ -141,3 +141,52 @@ function formatFileSizeBigInt(bytes: bigint): string {
...
+  }
+  return value * multiplier
+}
+
+/** Parse a positive integer flag value; throws a clear error for missing or non-positive input. */
+export function parsePositiveInt(raw: string | undefined, flag: string): number {
+  if (raw == null) {
+    throw new Error(`${flag} requires a value`)
+  }
+  const n = Number.parseInt(raw, 10)
```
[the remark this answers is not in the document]
SgtPooki: Added in bdda19f: `it.each` over the accepted forms and over `undefined`, empty, `0`, `-3`, `01`, `2.5`, `abc`, `1e3`, plus boundary cases for `parseSize`.

documentation/migrate.md
SgtPooki: Added a paragraph after the intro: a migrate piece is one multi-root CAR whose blocks are the member CARs of many source CIDs, only the aggregate gets a PieceCID and is uploaded, and which table holds which (3df3f81).

src/migrate/db.ts
SgtPooki: Fixed in bdda19f: `recordUploadParked` now refuses `committed` and `add_unconfirmed` rows, and #656 skips the secondary pull when the row is live. Regression test for your scenario in migrate-db.test.ts and the flow test in #656.

src/migrate/db.ts
SgtPooki: Fixed in bdda19f: same guard on `markUploadFailed`, a transient pull error no longer demotes a committed or unconfirmed row. Regression tests in migrate-db.test.ts and migrate-direct-upload-flow.test.ts (#656).

PR #658, opened 2026-08-18
feat!(cli): default egress to none and reuse existing data sets

SgtPooki:
## What changed

Two isolated commits:

1. `feat!(cli): default --egress-provider to none`. FilBeam CDN egress is now opt-in for `add` and `import` (`--egress-provider beam` or `EGRESS_PROVIDER=beam`). The default no longer creates CDN-enabled data sets or locks the extra 1 USDFC, and the CLI now matches the upload-action, which already defaulted to `none`. Reverts the default introduced in #464.

2. `feat(cli): reuse existing filecoin-pin data sets by default`. Without `--data-set-id`, `--provider-id`, or `--data-set-metadata`, `add`/`import` now match any live, active data set with `source=filecoin-pin` and upload into it. The SDK's smart-select requires exact metadata equality, so after the default flip every upload from an existing user would have skipped their `withCDN`-tagged data sets and created new ones per copy. The subset match reuses them instead. When more data sets match than copies requested, it picks the oldest ones already holding pieces, at most one per provider (as documented in the glossary's Data Set entry). With `--egress-provider beam`, only CDN-enabled data sets qualify.

## Calibration test runs

Wallet `0x44f0...3759` (972 data sets: 970 non-CDN, 2 CDN from the old `beam` default).

<details>
<summary>Old flow: v1.3.0, no flags (beam default)</summary>

```
Filecoin Pin Add

✓ File validated (69.0 B)
✓ Connected to Filecoin - Calibration testnet

Egress: FilBeam
  • Egress consumes the data set owner's locked-up funds.
  • FilBeam routes piece/CAR retrieval only, not IPFS blocks.
  • Each new data set locks an extra 1 USDFC.
  • Disable: --egress-provider none
✓ File packed with root CID: bafkreigajxvcttwifh5vw2vggn3bne6vterq5khcimz6u6uql525j3x7iy
✓ Cost estimate ready

Copies
  Requested: 2 (2 existing data sets)
```

Every upload defaults into CDN-tagged data sets (egress billed to the owner). For a wallet without CDN data sets, this creates 2 new ones and locks 2 USDFC, even when hundreds of non-CDN filecoin-pin data sets exist: exact metadata matching cannot see them.

</details>

<details>
<summary>New flow: this branch, no flags (none default)</summary>

```
Filecoin Pin Add

✓ File validated (69.0 B)
✓ Connected to Filecoin - Calibration testnet

✓ Reusing existing data sets 231, 54 (966 matched, picked the 2 storing the most data)

✓ File packed with root CID: bafkreigajxvcttwifh5vw2vggn3bne6vterq5khcimz6u6uql525j3x7iy
✓ Cost estimate ready

Copies
  Requested: 2 (2 existing data sets)
```

No FilBeam notice, no CDN lockup, and the upload lands in the wallet's own long-lived data sets (231: 595 pieces on provider 7, 54: 215 pieces on provider 4). Data set 84 (541 pieces) did not qualify: its PDP payments have ended (`pdpEndEpoch` set).

</details>

<details>
<summary>New flow with --egress-provider beam: reuses only CDN data sets, confirmed on-chain</summary>

```
✓ Reusing existing data sets 14030, 14031

✓ [Primary] Stored on provider 4
✓ [Secondary] Stored on provider 2
✓ IPNI provider records found. IPFS retrieval possible.
✓ [Primary] Piece added to Data Set (confirmed on-chain)
✓ [Secondary] Piece added to Data Set (confirmed on-chain)

Copies
  [Primary] Provider 4
    Data Set ID: 14030
    Piece ID: 6
  [Secondary] Provider 2
    Data Set ID: 14031
    Piece ID: 5
```

With `beam` requested, only the 2 CDN-enabled data sets qualify (not the 966 non-CDN ones), so the egress request is never silently dropped.

</details>

## How to verify

`pnpm run build && npx vitest run`. New coverage: `resolveDefaultDataSetReuse`/`pickDataSetsForReuse` in `upload-flow.test.ts`, glue tests in `add.test.ts`/`import.test.ts`, and the flipped egress-default tests.

## Notes / risks

- With fewer matching data sets than requested copies (including none), behavior is unchanged: the SDK resolves or creates data sets. A partial match (1 existing set, 2 copies requested) still creates 2 new sets rather than mixing reuse with creation.
- Reuse pins the upload to the picked data sets' providers with no health check. Observed on calibration: provider 7 (backing data set 231) was returning 502s, which fails the default upload until the SP recovers; `--provider-id` overrides. Candidate follow-up: skip unreachable providers during the pick.
- With `--egress-provider none`, CDN-tagged data sets still qualify for reuse: `none` means "do not request or create CDN", not "never add to a CDN data set". Excluding them would recreate the new-set churn for wallets whose only data sets came from the old `beam` default. Documented in the `resolveDefaultDataSetReuse` docblock.
- Reuse requires one data set per distinct provider. When the matches do not cover enough providers, including an exact-count match whose data sets share one, reuse is abandoned and new data sets are created.
- The pinning server's upload path still uses SDK smart-select; it never set `withCDN`, so its data sets already match exactly.
- The subset matcher now also skips data sets scheduled for termination (`pdpEndEpoch` set); this applies to `--data-set-metadata` resolution too.

generated with claude, peer-reviewed with three other models (codex, cursor-agent, gemini)


src/core/data-set/resolve-by-metadata.ts
```diff
@@ -32,11 +39,17 @@ export async function resolveDataSetIdsByMetadata(
...
+
   const matched = await listDataSets(synapse, {
     filter: (dataSet) => {
       if (!dataSet.isLive) {
         return false
       }
+      // Exclude datasets scheduled for termination; uploads must target active ones.
+      if ((dataSet.pdpEndEpoch ?? 0n) !== 0n) {
+        return false
+      }
```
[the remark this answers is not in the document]
SgtPooki: `pdpEndEpoch` is always a `bigint` at runtime; synapse-sdk types it as `bigint` on `DataSetInfo`, and the `?? 0n` covers test fixtures that omit the field. The number literals in `data-set.test.ts` were loose fixtures behind a `Record<string, unknown>` fixture type, not representative of SDK data; #666 fixes them and tightens the fixture type so the compiler rejects number literals.

src/common/upload-flow.ts
```diff
@@ -120,6 +121,89 @@ export async function promptDataSetSelection(
...
+ * lowest ID for determinism), spreading picks across distinct providers before
+ * doubling up on one — copies on the same provider add no redundancy.
+ */
+export function pickDataSetsForReuse(dataSets: DataSetSummary[], count: number): bigint[] {
+  const sorted = [...dataSets].sort((a, b) => {
+    if (a.activePieceCount !== b.activePieceCount) {
+      return a.activePieceCount > b.activePieceCount ? -1 : 1
+    }
+    return a.dataSetId < b.dataSetId ? -1 : 1
+  })
```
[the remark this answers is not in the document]
SgtPooki: `activePieceCount` is a required `bigint` on `EnhancedDataSetInfo` (synapse-sdk `types.d.ts`), and BigInt relational operators do not throw on `undefined` (`1n > undefined` evaluates to `false`; only arithmetic operators throw on mixed operands). The `?? 0n` elsewhere is display code rendering user-selected rows. No change.

SgtPooki (commented):
self review. Note that this removes FilBeam as default for add/import operations, and fixes data-set matching so filecoin-pin focuses only on finding data-sets where `source: filecoin-pin`, so we don't force existing users who are adding data to create new data-sets when this goes through.

I think defaulting to just source matching for data-sets is a net improvement for filecoin-pin users.

beck-8 (approved):
Although I don't know what the background of the change is. But there seems to be no problem with logic.

src/common/upload-flow.ts
```diff
@@ -120,6 +121,90 @@ export async function promptDataSetSelection(
...
+  }
+
+  if (resolution.kind === 'too-few-matches') {
+    spinner.stop(
+      `${pc.gray('•')} Found ${resolution.matchedIds.length} existing data set(s) but need ${expectedCopies}; the SDK will resolve or create data sets`
+    )
+    return undefined
+  }
+
+  spinner.stop(`${pc.gray('•')} No existing filecoin-pin data sets found; the SDK will create new ones`)
```
rvagg: "the SDK" - this is leaky detail the user doesn't need to see

src/common/upload-flow.ts
```diff
@@ -120,6 +121,90 @@ export async function promptDataSetSelection(
...
+
+  if (resolution.kind === 'matched') {
+    spinner.stop(`${pc.green('✓')} Reusing existing data sets ${resolution.dataSetIds.join(', ')}`)
+    return resolution.dataSetIds
+  }
+
+  if (resolution.kind === 'too-many-matches') {
+    const chosen = pickDataSetsForReuse(resolution.matchedDataSets, expectedCopies)
+    spinner.stop(
+      `${pc.green('✓')} Reusing existing data sets ${chosen.join(', ')} (${resolution.matchedIds.length} matched, picked the ${expectedCopies} storing the most data)`
```
rvagg: "most data" maybe not, but "most pieces"
but, I don't know you really need to be too wordy in these messages, terseness is nice so you don't overload the user with detail they may not want to try to understand

src/common/upload-flow.ts
```diff
@@ -120,6 +121,90 @@ export async function promptDataSetSelection(
...
+  for (const ds of sorted) {
+    if (picked.length >= count) break
+    if (seenProviders.has(ds.providerId)) continue
+    seenProviders.add(ds.providerId)
+    picked.push(ds)
+  }
+  for (const ds of sorted) {
+    if (picked.length >= count) break
+    if (!picked.includes(ds)) picked.push(ds)
+  }
```
rvagg: I think you need to remove _"before doubling up on one"_ variation in this method, as you say: "copies on the same provider add no redundancy" so why do it? if you can't find unique providers, then it's time to set up a new data set, but becasuse this function is assumed to pass you're going to need to handle a failure case in `resolveDefaultDataSetReuse` to fall-through to the "let the SDK handle it"

SgtPooki: yea, good catch. fixed.

SgtPooki:
@rvagg @hugomrdias i'm hoping to get this out before [releasing 2.0.0](https://github.com/filecoin-project/filecoin-pin/pull/669) tomorrow. please flag anything you want addressed before that!

hugomrdias (approved):
LGTM

PR #666, opened 2026-08-19
test: use bigint literals for SDK bigint fields

SgtPooki:
## What changed

`data-set.test.ts` fixtures set `pdpEndEpoch` and `commissionBps` as number literals, but synapse-sdk types both as `bigint` on `DataSetInfo`. The loose fixture type (`Record<string, unknown> & {...}`) let the mismatch through, and the display helpers only use relational comparisons, which accept mixed bigint/number operands. This switches the literals to bigint and declares both fields on `EnhancedDataSetFixture` so the compiler rejects number literals.

Runtime values were never affected; SDK data is always bigint. The number-typed fixtures were cited in #658 review as evidence that runtime `pdpEndEpoch` can be a number, which requested an unneeded `BigInt()` normalization in production code. Fixtures should not misrepresent SDK data shapes.

## How to verify

`npx vitest run src/test/unit/data-set.test.ts` and `npx tsc --noEmit`.

### DIFF
--- a/src/test/unit/data-set.test.ts
+++ b/src/test/unit/data-set.test.ts
@@ -217,6 +217,10 @@ type EnhancedDataSetFixture = Record<string, unknown> & {
   isLive?: boolean
   isManaged?: boolean
   withCDN?: boolean
+  /** bigint to mirror synapse-sdk's DataSetInfo; number literals here misrepresent SDK data. */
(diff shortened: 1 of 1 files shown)


SgtPooki (commented):
self review

beck-8 (approved):
LGTM

PR #670, opened 2026-08-25
feat(session): --scopes flag on create/authorize/revoke

jennijuju:
allow users to better control  session key scopes and manage their operations respectively

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,8 +193,8 @@ npm install -g filecoin-pin
 #    Revoke later:  filecoin-pin session revoke <session-address>
 #    Scoped keys:   session create/authorize/revoke take --scopes <ids> to grant or
 #                   revoke a subset (default: all). Ids: createDataSet, addPieces,
-#                   schedulePieceRemovals, terminateService (short aliases: create, add,
-#                   remove, delete, terminate). E.g. --scopes add,create or --scopes remove.
+#                   schedulePieceRemovals, terminateService.
+#                   E.g. --scopes createDataSet,addPieces or --scopes schedulePieceRemovals.
 
 # 1. Configure payment permissions (one-time setup)
 filecoin-pin payments setup --auto
--- a/README.md
+++ b/README.md
@@ -191,6 +191,10 @@ npm install -g filecoin-pin
 #                   `filecoin-pin session create` or `filecoin-pin session generate`
 #                   (or pass --wallet-address <addr> --session-key <private-key> to each command)
 #    Revoke later:  filecoin-pin session revoke <session-address>
+#    Scoped keys:   session create/authorize/revoke take --scopes <ids> to grant or
+#                   revoke a subset (default: all). Ids: createDataSet, addPieces,
+#                   schedulePieceRemovals, terminateService (short aliases: create, add,
+#                   remove, delete, terminate). E.g. --scopes add,create or --scopes remove.
 
 # 1. Configure payment permissions (one-time setup)
 filecoin-pin payments setup --auto
--- a/src/commands/session.ts
+++ b/src/commands/session.ts
@@ -7,7 +7,7 @@
 
 import { Command } from 'commander'
 import { runSessionAuthorize, runSessionCreate, runSessionGenerate, runSessionRevoke } from '../session/index.js'
-import { addOwnerAuthOptions, sessionKeyOption } from '../utils/cli-options.js'
+import { addOwnerAuthOptions, scopesOption, sessionKeyOption } from '../utils/cli-options.js'
 
 export const sessionCommand = new Command('session').description('Manage session keys for delegated upload access')
 
@@ -16,10 +16,7 @@ export const sessionCommand = new Command('session').description('Manage session
 const createCommand = new Command('create')
   .description('Generate (or reuse) a session key and authorize it on-chain')
   .option('--validity-days <days>', 'Number of days the session key should be valid (max 365)', '10')
-  .option(
-    '--scopes <ids>',
-    'Comma-separated permissions to grant (default: all). Any of: createDataSet, addPieces, schedulePieceRemovals, terminateService'
-  )
+  .addOption(scopesOption('Comma-separated permissions to grant (default: all)'))
   .addOption(sessionKeyOption('Reuse an existing session private key'))
   .action(async (options) => {
     try {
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -191,6 +191,10 @@ npm install -g filecoin-pin
 #                   `filecoin-pin session create` or `filecoin-pin session generate`
 #                   (or pass --wallet-address <addr> --session-key <private-key> to each command)
 #    Revoke later:  filecoin-pin session revoke <session-address>
+#    Scoped keys:   session create/authorize/revoke take --scopes <ids> to grant or
+#                   revoke a subset (default: all). Ids: createDataSet, addPieces,
+#                   schedulePieceRemovals, terminateService (short aliases: create, add,
+#                   remove, delete, terminate). E.g. --scopes add,create or --scopes remove.
```
BigLep: I don't love the alias concept given this is a low-usage API (I assume). Your call, but I would simplify and not include it.

Also, it's not obvious to me how we have 4 verbose scopes:

createDataSet, addPieces, schedulePieceRemovals, terminateService

but 5 aliases:

create, add,  remove, delete, terminate

BigLep (commented):
Conceptually makes sense to me. I'll leave to agent and Russell to actually certify the code.  Happy to spend more time on this though if useful.

README.md
SgtPooki: Dropped the aliases in 360be6f..HEAD. `--scopes` now accepts only the four ids (createDataSet, addPieces, schedulePieceRemovals, terminateService), case-insensitive. The fifth alias was `remove` and `delete` both mapping to schedulePieceRemovals. README, help text, and tests updated.

src/commands/session.ts
```diff
@@ -16,6 +16,10 @@ export const sessionCommand = new Command('session').description('Manage session
 const createCommand = new Command('create')
   .description('Generate (or reuse) a session key and authorize it on-chain')
   .option('--validity-days <days>', 'Number of days the session key should be valid (max 365)', '10')
+  .option(
+    '--scopes <ids>',
+    'Comma-separated permissions to grant (default: all). Any of: createDataSet, addPieces, schedulePieceRemovals, terminateService'
+  )
```
SgtPooki: we should extract this to a shared helper.

src/session/run-create.ts
```diff
@@ -42,6 +43,17 @@ export async function runSessionCreate(options: SessionCreateOptions): Promise<C
...
+  let scopes: ParsedScopes | undefined
+  if (options.scopes !== undefined) {
+    try {
+      scopes = parseScopes(options.scopes)
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error)
+      cancel(message)
+      throw error
+    }
+  }
```
SgtPooki: this is also something that's repeated a few times. I think we need a helper for this, and this should probably be piped up to /core, but we could open a tracking issue for bumping this up later if desired via programmatic access..

src/session/run-revoke.ts
```diff
@@ -43,6 +44,17 @@ export async function runSessionRevoke(options: SessionRevokeOptions): Promise<R
...
+  let scopes: ParsedScopes | undefined
+  if (options.scopes !== undefined) {
+    try {
+      scopes = parseScopes(options.scopes)
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error)
+      cancel(message)
+      throw error
+    }
+  }
```
SgtPooki: same code repeated three times..

src/test/unit/session-scopes.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  it('rejects input with no valid scope', () => {
+    expect(() => parseScopes('')).toThrow(/at least one/)
+    expect(() => parseScopes(' , , ')).toThrow(/at least one/)
+  })
+})
+
+describe('describeScopes', () => {
+  it('renders undefined as the full set', () => {
+    expect(describeScopes()).toBe(`all (${SCOPE_IDS.join(', ')})`)
+  })
```
SgtPooki: this test isn't needed.. its just verify something that is easily read in the code.

src/test/unit/session-scopes.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  it('joins provided ids', () => {
+    expect(describeScopes(['addPieces', 'createDataSet'])).toBe('addPieces, createDataSet')
+  })
```
SgtPooki: this test also feels off.

src/test/unit/session-scopes.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import { describe, expect, it } from 'vitest'
+import { TerminateServicePermission } from '../../core/session/index.js'
+import { describeScopes, parseScopes, SCOPE_IDS } from '../../session/scopes.js'
+
+describe('parseScopes', () => {
+  it('parses a single scope to its permission typehash', () => {
+    const { ids, permissions } = parseScopes('addPieces')
+    expect(ids).toEqual(['addPieces'])
+    expect(permissions).toEqual([AddPiecesPermission])
+  })
```
SgtPooki: feels unnecessary given the `maps all four ids to their permission typehashes in canonical order` test below.

src/session/scopes.ts
```diff
@@ -0,0 +1,74 @@
...
+export function parseScopes(value: string): ParsedScopes {
+  const tokens = value
+    .split(',')
+    .map((token) => token.trim())
+    .filter((token) => token.length > 0)
+  if (tokens.length === 0) {
+    throw new Error(`--scopes needs at least one of: ${SCOPE_IDS.join(', ')}`)
+  }
+
+  const canonicalByLower: Record<string, string> = Object.fromEntries(SCOPE_IDS.map((id) => [id.toLowerCase(), id]))
```
SgtPooki: we can bump this up so that it doesn't need to run on each pass through this function.

SgtPooki (approved):
some minor changes needed.. I'll get those in

src/commands/session.ts
SgtPooki: Extracted `scopesOption(description)` into `src/utils/cli-options.ts` in 94efba0; the three subcommands (and `login`, further up the stack) use it.

src/session/run-create.ts
SgtPooki: Extracted `parseScopesOption` in `src/session/scopes.ts` (94efba0); the three runners call it. Parsing stays in the CLI layer since core takes `Permission[]` and never sees the `--scopes` string. I have not opened a tracking issue for programmatic access to the id list; say so if you want one.

src/session/run-revoke.ts
SgtPooki: Same helper as above; the three blocks are one call now (94efba0).

src/test/unit/session-scopes.test.ts
SgtPooki: Removed in 94efba0.

src/test/unit/session-scopes.test.ts
SgtPooki: Removed in 94efba0.

src/test/unit/session-scopes.test.ts
SgtPooki: Removed in 94efba0.

src/session/scopes.ts
SgtPooki: Hoisted to a module-level `SCOPE_ID_BY_LOWERCASE` in 94efba0.

PR #671, opened 2026-08-25
feat(cli): add --credentials-file to load dotenv-style credentials

jennijuju:
Adds `--credentials-file <path>` so a downloaded credentials file (for example the Filecoin Pay console's session-key `.env`) can be loaded without sourcing it into the shell. Values from the file never override variables already set in the environment, and flags still win over both.

The flag was `--env-file` in the first revision. Node reserves that name and validates it before our code runs, so a bad path produced a Node error instead of ours. `--credentials-file` is not a Node flag, and it names what the file holds.

The option is registered on every command tree that takes auth options, including `server` and `session create/authorize/revoke`, so Commander accepts it wherever the pre-parse loader reads it.

Parsing uses `util.parseEnv` from Node.

Part of the login epic, tracked in #697.


src/utils/env-file.ts
```diff
@@ -0,0 +1,123 @@
...
+      if ((first === '"' && last === '"') || (first === "'" && last === "'")) {
+        value = value.slice(1, -1)
+      }
+    }
+
+    result[key] = value
+  }
+
+  return result
+}
```
SgtPooki: you should use the [built-ins]( https://nodejs.org/api/environment_variables.html#programmatic-apis), specifically [`Util.parseEnv`](https://nodejs.org/api/util.html#utilparseenvcontent)

src/test/unit/env-file.test.ts
```diff
@@ -0,0 +1,138 @@
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { applyEnvFileArg, findEnvFileArg, loadEnvFile, parseEnvFile } from '../../utils/env-file.js'
+
+describe('parseEnvFile', () => {
```
SgtPooki: we should be able to remove the `parseEnvFile` tests once we use the builtin

SgtPooki (changes requested):
in general lgtm, one main change requested

SgtPooki (changes requested):
ok some findings when testing this out:

**Big issue in non-happy path**: since `--env-file` is also used by nodejs, it triggers nodejs validation (but not application) prior to hitting our code, which ends up returning a nodejs error that isn't super informative. We should rename the option so it doesn't conflict

**Smaller issue we should handle for completeness**: `server` and `session` sub-commands do read and process the `--env-file` if passed, but then emit a "error: unknown option '--env-file'" error. we likely need to add `envFileOption` into `addSigningAuthOptions`(server) and `addOwnerAuthOptions`(session). A later PR should probably normalize these better..

SgtPooki:
Renamed to `--credentials-file` in 0565a54. Both findings addressed: our error now fires on a bad path, and the option is registered through `addSigningAuthOptions` (covers `server` and every `addAuthOptions` command) and `addOwnerAuthOptions` (session create/authorize/revoke). A test asserts the flag is present on each of those trees.

SgtPooki (approved):
Re-reviewed at `18c3e73`. Both previously requested changes are in: the rename off Node's reserved `--env-file`, and the switch to stdlib `util.parseEnv` (redundant parser tests removed).

**One minor finding (non-blocking):** the flag is rejected when placed *before* the subcommand, even though the pre-parse loader still loads the file. `findCredentialsFileArg` scans argv position-independently, but the top-level `program` in `src/cli.ts` never registers the option:

```
$ filecoin-pin add --credentials-file X path   ✓ works
$ filecoin-pin add path --credentials-file X   ✓ works
$ filecoin-pin --credentials-file X add path   ✗ error: unknown option '--credentials-file'
```

The asymmetry: with a *bad* path before the subcommand, our loader error fires first; with a *valid* path, Commander rejects it. Suggest registering `credentialsFileOption()` on the top-level `program` (one line in `src/cli.ts`) so the flag works in every position the loader honors — or, if after-subcommand placement is deliberate, soften the PR description's "Commander accepts it wherever the pre-parse loader reads it".

Verified live:
- Missing file (space + `=` forms) → `could not read "<path>": file not found`; comment-only file → `no usable entries` with the dotenv format hint
- Precedence flag > env > file: file set `RPC_URL=:49999`, env `:49998` → CLI probed `:49998`
- `server` and `session create` accept the flag (pass parse, fail at business logic with dummy creds)
- `add --help` lists the option exactly once (no double registration via `addAuthOptions` + `addContextSelectionOptions`)
- `vitest run` on both touched test files: 32/32 pass

Note: local `pnpm run build` fails on pre-existing type errors in files this PR doesn't touch (`@filoz/synapse-sdk` missing `FilecoinChain` export, `hasActivePieces` in `data-set-display.test.ts`) — environment drift, unrelated to this change.

SgtPooki (approved):

PR #683, opened 2026-08-26
feat!: stop exposing piece metadata

### LINKED ISSUE #668: Stop exposing piece metadata
Storing piece metadata is being removed from FWSS; we should stop querying from the chain.
The pathways can remain in filecoin-pin if we want to reintroduce it in the future through alternative means, but for now, it's OK to drop features that rely on reading piece metadata.

silent-cipher:
## Summary

- Stop querying piece metadata from fwss
- Remove piece metadata and IPFS Root CID enrichment from piece-status output
- Preserves piece-metadata writes during `add` and `import`
- Preserve Data Set metadata reads and display
- Remove the metadata prefetch machinery and direct `p-queue` dependency
- Update documentation to distinguish stored dataset metadata from event-only piece metadata

Closes #668 

## Breaking Changes

- Remove `PieceInfo.metadata`
- Remove `PieceInfo.rootIpfsCid`
- Remove `GetDataSetPiecesOptions.includeMetadata`
- Remove the exported `enrichPieceMetadata()` function
- Remove IPFS Root CID filtering from `data-set piece-status`

**Note:** This PR does not use the preview packages from FilOzone/synapse-sdk#936. It removes filecoin-pin's calls to the piece-metadata getter APIs removed by that PR while retaining the currently release dependencies. Once synapse release prs FilOzone/synapse-sdk#941 and FilOzone/synapse-sdk#942 are merged and `@filoz/synapse-core@v0.9.0` and `@filoz/synapse-sdk@v2.0.0` are published, the dependency upgrade can be handled as a follow-up or included here.

### DIFF
--- a/src/core/data-set/types.ts
+++ b/src/core/data-set/types.ts
@@ -39,12 +39,8 @@ export interface PieceInfo {
   /** Piece Commitment (CommP) as string */
   pieceCid: string
   status: PieceStatus
-  /** Root IPFS CID (from metadata, if available) */
-  rootIpfsCid?: string
   /** Piece size in bytes (if available) */
   size?: number
-  /** Additional piece metadata (key-value pairs) */
-  metadata?: Record<string, string>
 }
 
 /**
@@ -127,7 +123,7 @@ export interface ListDataSetsOptions {
 /**
  * Options for getting pieces from a dataset
  */
-export interface IterateDataSetPiecesOptions {
+export interface GetDataSetPiecesOptions {
   /** Abort signal for cancellation */
   signal?: AbortSignal
   /** Logger instance for debugging (optional) */
@@ -135,9 +131,6 @@ export interface IterateDataSetPiecesOptions {
 }
 
 /**
- * Options for getting pieces from a dataset
+ * Options for iterating over pieces from a dataset
  */
-export interface GetDataSetPiecesOptions extends IterateDataSetPiecesOptions {
-  /** Whether to fetch and include piece metadata from WarmStorage */
-  includeMetadata?: boolean
-}
+export interface IterateDataSetPiecesOptions extends GetDataSetPiecesOptions {}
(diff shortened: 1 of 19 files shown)


src/core/data-set/types.ts
```diff
@@ -127,17 +123,14 @@ export interface ListDataSetsOptions {
...
 
 /**
- * Options for getting pieces from a dataset
+ * Options for iterating over pieces from a dataset
  */
-export interface GetDataSetPiecesOptions extends IterateDataSetPiecesOptions {
-  /** Whether to fetch and include piece metadata from WarmStorage */
-  includeMetadata?: boolean
-}
+export interface IterateDataSetPiecesOptions extends GetDataSetPiecesOptions {}
```
beck-8: Do we still need to keep these two same structures here?

beck-8 (commented):
I'm not professional here, so I won't give a clear approve. I can only express the doubts I see.

src/core/data-set/types.ts
SgtPooki: we can likely remove one of these or merge them.

src/data-set/display.ts
SgtPooki, line 322: this is the only place `renderMetadata` is called now.. could remove some unused parameters

SgtPooki (approved):
in general lgtm.. a few comments we could address as followups

hugomrdias (approved):
LGTM

src/data-set/display.ts
silent-cipher: Fixed in [5d758e6](https://github.com/filecoin-project/filecoin-pin/pull/683/commits/5d758e6cda8c015ca6f4e6fdc917b9039fe7d789)

src/core/data-set/types.ts
silent-cipher: Fixed in [5d758e6](https://github.com/filecoin-project/filecoin-pin/pull/683/commits/5d758e6cda8c015ca6f4e6fdc917b9039fe7d789)

PR #686, opened 2026-08-27
fix(deps)!: update synapse-sdk to 1.2.1 and core to 0.8.1

### LINKED ISSUE #684: Update to latest Synapse release (synapse-sdk@1.2.1)
filecoin-pin is pinned to `@filoz/synapse-sdk@1.1.1` / `@filoz/synapse-core@0.7.1`, well behind the latest published `1.2.1`. Bump both deps and fix whatever breaks.

**Why a manual issue instead of a Dependabot PR:** Dependabot is currently blocked from opening new PRs on this repo. It's hit the open-PR limit (5) on the [dependency graph's Dependabot tab](https://github.com/filecoin-project/filecoin-pin/network/updates/1540941093). Until those existing PRs are cleared, it won't propose this bump on its own, so filing it here to make sure it doesn't get missed.

**Why now:** per the 2026-08-25 FOC WG engineering sync, this is a step in [[Tracking Issue] 202608 FOC Release (FilOzone/filecoin-services#597)](https://github.com/FilOzone/filecoin-services/issues/597). It picks up the recent pagination/out-of-gas fixes and the new piece-metadata deprecation warnings (synapse-sdk#938) ahead of their removal in Synapse 2.0.

**Scope:**
- Bump `@filoz/synapse-sdk` to `^1.2.1` and `@filoz/synapse-core` to the matching latest.
- Fix any breakage from the pagination / active-piece-count changes, particularly in `src/data-set/piece-status-pager.ts` and `src/core/data-set` (`iterateDataSetPieces`, `enrichPieceMetadata`).
- Run full test suite against the bumped deps.

**Not necessarily in scope:** removing piece-metadata reliance entirely. That's tracked in #668, which is already out for PR and fine to land independently of this bump.

silent-cipher:
## Summary

- bump `@filoz/synapse-sdk` from `^1.1.0` to `^1.2.1`
- bump `@filoz/synapse-core` from `^0.7.0` to `^0.8.1`
- migrate active-piece enumeration from offset pagination to `getActivePieceByCursor`
- pass opaque continuation cursors through unchanged and reject non-advancing cursors
- paginate approved-provider queries so all pages are loaded
- replace the removed `activePieceCount` field with `hasActivePieces` in data-set displays and selection prompts
- migrate sdk `chain` usages to `FilecoinChain`, while retaining filecoin-pin's exported `Chain` alias for compatibility
- retain the existing piece-metadata enrichment behavior pending its separately tracked removal

Closes #684

## Breaking Change

This pr includes an intentional public API change: `DataSetSummary.activePieceCount` is replaced by `DataSetSummary.hasActivePieces`, following the synapse sdk 1.2.1 api.

Preserving an exact `activePieceCount` would require fetching and traversing every page of active pieces. The cost would grow with the size of the data set and would recreate the potentially unsafe, unbounded query behavior that the upstream sdk change was designed to eliminate. Using `hasActivePieces` provider the information needed by `filecoin-pin` without requiring a complete enumeration.


src/core/synapse/index.ts
```diff
@@ -8,9 +8,12 @@
...
  */
 
-import { type Chain, calibration, mainnet, Synapse, type SynapseOptions } from '@filoz/synapse-sdk'
+import { calibration, type FilecoinChain, mainnet, Synapse, type SynapseOptions } from '@filoz/synapse-sdk'
 
-export { calibration, mainnet, type Chain }
+export { calibration, mainnet, type FilecoinChain }
+
+/** Backwards-compatible alias for the Filecoin chain configuration type. */
+export type Chain = FilecoinChain
```
SgtPooki: we are already doing breaking changes, lets just remove this

SgtPooki (approved):
approving, but I would like to remove the "backwards compat" `Chain` export

src/core/synapse/index.ts
silent-cipher: I've completely removed the use of the Chain alias (in exports as well as for internal use).

SgtPooki (approved):

PR #687, opened 2026-08-28
feat(session): per-command scope gating with console-first remediation

jennijuju:
e2e preview:https://inbrowser.link/ipfs/bafybeicfitpkj5i5nu37o2dlntbibzbzsl6dul3mvoh2fy3dn34qguxv64
depend on https://github.com/FilOzone/filecoin-pay-explorer/pull/355

Part of the login epic, tracked in #697.

currently session key mode preflights all four FWSS permissions on every command — even pure reads like `payments status` — so a least-privilege key (upload-only) can't use the CLI at all. Since SessionKeyRegistry supports per-permission grants, the chain already allows subsets; only the CLI refused them. (inherited from synapse)

Commands now declare what they need (checked in the preflight AND passed to `Synapse.create`, whose own gate otherwise defaults to all four):

| command | permissions required |
|---|---|
| reads (`payments status`, `data-set ls`, `provider ls`, …) | none |
| `add` / `import` | createDataSet, addPieces |
| `rm` | schedulePieceRemovals |
| `data-set terminate` | terminateService |
| pinning server | createDataSet, addPieces, schedulePieceRemovals |

A missing scope fails up front with a console-first message: the problem, with per-scope detail (`expired at <timestamp>` vs `never granted`, read from the on-chain expirations) and the network it happened on;

 a Filecoin Pay console deep link (`…/console/session-keys?authorize=<addr>&scopes=<missing>&network=<net>`) to approve with the owner wallet; then the owner-only CLI alternatives (`session authorize/create --scopes`, from the base PR). A delegate is never told to run a root-key command as their own action.

### DIFF
--- a/src/core/session/console-url.ts
+++ b/src/core/session/console-url.ts
@@ -28,11 +28,28 @@ export function resolveConsoleUrl(chainId: number, override?: string): string | 
   return override ?? process.env.CONSOLE_URL ?? DEFAULT_CONSOLE_URLS[chainId]
 }
 
+/** Console network slug by chain id; the console validates and guards on it. */
+const CONSOLE_NETWORK_SLUG: Record<number, string> = {
+  314: 'mainnet',
+  314159: 'calibration',
+}
+
 /**
  * Build the console deep link that pre-fills the session address and the
- * scopes it needs on the session-keys authorization page.
+ * scopes it needs on the session-keys authorization page. Carries the
+ * network the failure happened on so the console can refuse to prefill
+ * when the connected wallet is on a different chain — without it, a
+ * calibration remediation link approved by a mainnet-connected wallet
+ * silently grants the scopes on mainnet.
  */
-export function buildAuthorizeUrl(consoleUrl: string, sessionAddress: string, scopeIds: string[]): string {
+export function buildAuthorizeUrl(
+  consoleUrl: string,
+  sessionAddress: string,
+  scopeIds: string[],
+  chainId?: number
+): string {
   const base = consoleUrl.endsWith('/') ? consoleUrl.slice(0, -1) : consoleUrl
-  return `${base}/console/session-keys?authorize=${sessionAddress}&scopes=${scopeIds.join(',')}`
+  const network = chainId != null ? CONSOLE_NETWORK_SLUG[chainId] : undefined
+  const networkParam = network ? `&network=${network}` : ''
+  return `${base}/console/session-keys?authorize=${sessionAddress}&scopes=${scopeIds.join(',')}${networkParam}`
 }
--- a/src/core/session/console-url.ts
+++ b/src/core/session/console-url.ts
@@ -20,12 +20,12 @@ export const DEFAULT_CONSOLE_URLS: Record<number, string> = {
 }
 
 /**
- * Pick the console base URL: an explicit override wins, then `CONSOLE_URL`,
- * then the known deployment for the chain. Returns `undefined` when none of
- * those resolve — the caller falls back to a console-without-a-link message.
+ * Pick the console base URL: `CONSOLE_URL` wins, then the known deployment
+ * for the chain. Returns `undefined` when neither resolves — the caller
+ * falls back to a console-without-a-link message.
  */
-export function resolveConsoleUrl(chainId: number, override?: string): string | undefined {
-  return override ?? process.env.CONSOLE_URL ?? DEFAULT_CONSOLE_URLS[chainId]
+export function resolveConsoleUrl(chainId: number): string | undefined {
+  return process.env.CONSOLE_URL ?? DEFAULT_CONSOLE_URLS[chainId]
 }
 
 /** Console network slug by chain id; the console validates and guards on it. */
--- a/src/core/session/console-url.ts
+++ b/src/core/session/console-url.ts
@@ -41,6 +41,10 @@ const CONSOLE_NETWORK_SLUG: Record<number, string> = {
  * when the connected wallet is on a different chain — without it, a
  * calibration remediation link approved by a mainnet-connected wallet
  * silently grants the scopes on mainnet.
+ *
+ * The address is lowercased: the console validates it with viem's strict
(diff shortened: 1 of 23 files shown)


src/core/synapse/index.ts
```diff
@@ -280,6 +322,10 @@ export async function initializeSynapse(config: SynapseSetupConfig, logger?: Log
   }
   if (sessionKey) {
     synapseOptions.sessionKey = sessionKey
+    // Match the SDK's own permission gate to this command's needs; without it
+    // Synapse.create defaults to requiring all FWSS permissions and re-rejects a
+    // subset key that our preflight already accepted.
+    synapseOptions.requiredPermissions = config.requiredPermissions ?? []
```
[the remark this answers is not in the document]
SgtPooki: Fixed in ef7a254. `payments deposit`, `withdraw`, `fund`, and `setup --auto` now call `assertOwnerAuth` before `initializeSynapse` and refuse session-key auth with a message naming the command. Payment operations are owner-only per the PRD, so refusing up front matches "session keys cannot deposit" better than declaring a permission the registry does not have.

src/core/synapse/index.ts
```diff
@@ -153,28 +148,75 @@ export function assertSessionKeyPrivateKey(value: string): asserts value is Hex
...
+    : `Session key ${key.address} lacks ${scopeLabels} for this operation on ${networkName}.`
 
-  throw new Error(
-    `Session key ${key.address} is missing ${missing.length} required permission(s):\n` +
-      lines.join('\n') +
-      `\nAuthorize this session key from owner wallet ${ownerAddress}.\nPermission hashes:\n` +
-      footnotes.join('\n')
-  )
+  const consoleUrl = resolveConsoleUrl(chainId)
+  const lines = neverAuthorized ? [problem, ''] : [problem, ...scopeDetails, '']
```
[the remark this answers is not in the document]
SgtPooki: Fixed in ef7a254. The scope detail lines are now included in both branches; the test for the all-zero case asserts `never granted` shows up.

src/test/unit/synapse-service.test.ts
```diff
@@ -98,6 +106,142 @@ describe('synapse-service', () => {
...
+      })
+
+      const config: SynapseSetupConfig = {
+        walletAddress: '0x0000000000000000000000000000000000000002',
+        sessionKey: '0x0000000000000000000000000000000000000000000000000000000000000001',
+        rpcUrl: 'wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v1',
+        requiredPermissions: [AddPiecesPermission, CreateDataSetPermission],
+      }
+
+      await expect(initializeSynapse(config, logger)).resolves.toBeDefined()
```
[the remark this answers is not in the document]
SgtPooki: Fixed in ef7a254. The test now asserts `Synapse.create` received exactly `[AddPieces, CreateDataSet]`, and a second test asserts `[]` is forwarded when nothing is required.

src/core/session/console-url.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Console deployments by chain id. One deployment serves both networks and
+ * the console switches network in-app, so mainnet and calibration share a
+ * base URL; the network travels in the `network` query parameter.
+ * `CONSOLE_URL` still overrides for local or preview consoles.
+ */
+export const DEFAULT_CONSOLE_URLS: Record<number, string> = {
+  314: 'https://pay.filecoin.cloud',
+  314159: 'https://pay.filecoin.cloud',
+}
```
SgtPooki: we probably don't need this split for now if the urls are the same. the comment can say network switch happens via query param

src/core/session/console-url.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Pick the console base URL: `CONSOLE_URL` wins, then the known deployment
+ * for the chain. Returns `undefined` when neither resolves — the caller
+ * falls back to a console-without-a-link message.
+ */
+export function resolveConsoleUrl(chainId: number): string | undefined {
+  return process.env.CONSOLE_URL ?? DEFAULT_CONSOLE_URLS[chainId]
+}
```
SgtPooki, line 16: can remove this when we just use `DEFAULT_CONSOLE_URL`

src/core/synapse/index.ts
```diff
@@ -153,28 +150,71 @@ export function assertSessionKeyPrivateKey(value: string): asserts value is Hex
...
       return `  • ${name}: expired at ${new Date(Number(expiry) * 1000).toISOString()}`
     }
-    return `  • ${name}: never authorized`
+    return `  • ${name}: never granted`
   })
 
-  const footnotes = missing.map((p) => `  ${PERMISSION_NAMES[p] ?? p}: ${p}`)
+  const problem = neverAuthorized
+    ? `Session key ${key.address} isn't authorized for account ${ownerAddress} on ${networkName} — never authorized, expired/revoked, or the key is for a different network (check --network).`
+    : `Session key ${key.address} lacks ${scopeLabels} for this operation on ${networkName}.`
```
SgtPooki: i'm wondering if we should look at this error the same way people look at "forgot password" form submissions for websites to determine if someone has registered for a site with a particular email... 

should we just say "no permissions" ?

src/test/mocks/synapse-core-session-key.ts
```diff
@@ -19,6 +19,13 @@ export const DefaultFwssPermissions = [
   SchedulePieceRemovalsPermission,
 ]
 
+export const PermissionNames: Record<string, string> = {
+  [CreateDataSetPermission]: 'CreateDataSet',
+  [TerminateServicePermission]: 'TerminateService',
+  [AddPiecesPermission]: 'AddPieces',
+  [SchedulePieceRemovalsPermission]: 'SchedulePieceRemovals',
+}
```
SgtPooki: why are we redefining this here?

SgtPooki (changes requested):
a few changes needed after unsupervised agent work

SgtPooki (approved):

Issue #665, opened 2026-08-19
Stop relying on filecoinpin.contact for indexing verification, use Curio piece-status endpoint instead


BigLep:
Depends on: #664, [curio#1405](https://github.com/filecoin-project/curio/issues/1405)

## Summary

filecoin-pin currently polls filecoinpin.contact directly (`waitForIpniProviderResults` in `src/core/utils`, exported via `filecoin-pin/core/utils`) to verify that uploaded content has been indexed and is discoverable. dealbot also consumes this same function today. We're migrating off filecoinpin.contact entirely (see #664), and the replacement logic belongs in filecoin-pin's library so both the CLI and dealbot get it from one place rather than each reimplementing it.

The new polling function does two things, in order:

1. **Poll Curio, not an indexer.** Poll Curio's `GET /pdp/piece/{pieceCid}/status` until it reports `synced: true` (curio#1405). This replaces today's direct polling of filecoinpin.contact's IPNI-compatible API.
2. **Confirm against cid.contact.** Once Curio reports `synced`, do one confirming query against cid.contact for the root CID (`GET https://cid.contact/cid/{rootCid}`). By this point it should already be indexed, so this shouldn't trigger cid.contact's negative cache — see #664's "Intended solution" section for why this ordering matters.

We don't expect step 2 to ever fail once step 1 says `synced`, but if it does — Curio thinks cid.contact has indexed the content and cid.contact has no record — that's a real anomaly, not routine "still indexing" noise. The library needs to expose that specific case (Curio/cid.contact disagreement) as a distinct, identifiable error or event, separate from "not yet indexed" or "timed out," so callers can react to it deliberately rather than it getting lost inside a generic verification failure.

## Steps

- [ ] Identify where filecoin-pin currently polls/queries filecoinpin.contact for indexing status (`waitForIpniProviderResults` and callers in `src/core/`).
- [ ] Replace that polling with calls to Curio's `GET /pdp/piece/{pieceCid}/status` endpoint, waiting for `synced`/`syncedAt`, once available (curio#1405).
- [ ] After `synced`, add the confirming cid.contact query for the root CID.
- [ ] Expose a distinct error/event when Curio reports `synced` but the cid.contact confirmation query doesn't find the CID (e.g. a dedicated error type or an emitted event distinguishable from "not yet indexed" / "timeout"), so callers can handle it explicitly.
[...]

Issue #664, opened 2026-08-19
[Tracking] Migrate FOC off filecoinpin.contact, use cid.contact only


BigLep:
_This is about migrating FOC as a whole from using filecoinpin.contact (not just filecoin-pin affordances). The tracking issue needs to live somewhere, and this seemed like the best repo to house it._ 

## Intended solution

<details>
<summary>Details</summary>

- Curio SPs advertise only to cid.contact (no longer to filecoinpin.contact).
- Curio SPs confirm indexing themselves: since Curio already knows the advertisement CID it just announced, it can poll cid.contact's `/sync/status/ad/{adCid}` endpoint directly to learn whether that specific ad has been fully processed. This endpoint doesn't have the negative-caching problem because it's keyed on the ad CID, not the content CID.
- Curio surfaces that confirmation on its own piece-status endpoint (`GET /pdp/piece/{pieceCid}/status`, per [curio#1405](https://github.com/filecoin-project/curio/issues/1405): adding `synced`/`syncedAt` fields). Clients like filecoin-pin and dealbot poll **that** endpoint, not cid.contact, to learn indexing status.
  - The key reason clients can't just poll cid.contact directly: cid.contact's ad-status check is keyed on the advertisement CID, and Curio is the only party that knows the advertisement CID for a given upload. An end client like filecoin-pin only knows the IPFS root CID, and has no structural way to derive the ad CID from it. If such a client polled `/cid/{rootCid}` on cid.contact before the content was actually indexed, it would trigger the negative-cache issue this whole migration is trying to avoid, hence the need to go through Curio's own status endpoint first, and only fall back to cid.contact once Curio has confirmed the content is ready.
- Only once Curio's endpoint reports the piece as synced do clients query cid.contact for the root CID (`GET https://cid.contact/cid/{rootCid}`). By then it should already be indexed, so the query shouldn't produce a negative-cache miss.
  - This final cid.contact query is a genuine confirmation step, not a formality: we don't expect Curio's `synced` signal and cid.contact's own record to ever disagree, but if they do (Curio says synced, cid.contact has no record for the root CID), that's a real anomaly worth surfacing loudly rather than silently retrying or swallowing. filecoin-pin's library should expose this as a distinct, identifiable error/event (separate from "not yet indexed"), and callers (the filecoin-pin CLI, dealbot) are each responsible for logging/metricing it. See the filecoin-pin and dealbot issues below for specifics.

</details>

## Steps

[...]

PR #689, opened 2026-08-28
feat!: confirm indexing via Curio piece-status

beck-8:
## Summary

Post-upload indexing verification no longer depends on filecoinpin.contact — its
hardware is being retired (#664). Instead:

1. Poll Curio's `GET /pdp/piece/{pieceCid}/status` per provider until `synced: true`.
2. Once every provider confirms, do a single bounded confirming query against
   `cid.contact` (never poll it repeatedly — it negative-caches misses for minutes).
3. If Curio says synced but cid.contact still disagrees, surface a distinct
   `IndexerMismatchError` / `indexingConfirmation:mismatch` event instead of a plain timeout.

`waitForIndexingConfirmation` replaces the old default check; `checkIpniIndexer`
(renamed from `waitForIpniProviderResults`, kept as a deprecated alias) remains as
the generic single-indexer primitive for standalone use.

## Breaking change

`CheckIpniIndexerOptions.ipniIndexerUrl` is now required — the previous implicit
default (filecoinpin.contact) is being retired, so every caller must choose an
indexer explicitly.

## Do not merge yet

Per #664: hold until Curio ships this piece-status behavior in a release and PDP
SPs upgrade. Opening as a draft.

## Test plan
- [x] `pnpm test` (lint, typecheck, unit, integration)

Relates to #665, #664


beck-8:
## CLI Output Examples

### Success — single provider (live-tested on calibration, provider 6)

```
✓ [Primary] Stored on provider 6
✓ [Primary] Piece added to Data Set (unconfirmed on-chain)
  Tx: https://calibration.filfox.info/en/message/0x72eb05a7fafe2d363a0efc6739d54ee9e71568775c6aee64d0665b82fbb33d7b
✓ Advertisement confirmed indexed
✓ IPNI provider records found. IPFS retrieval possible.

IPFS Retrieval URLs
  View in a browser: https://inbrowser.link/ipfs/bafkreicrz2vbbfmu3ulds4bv2arvqjp2qoke77vo37at2g2i4xuypokcfe
✓ [Primary] Piece added to Data Set (confirmed on-chain)
```

### Success — multiple providers *(expected output — traced from code, not yet observed live)*

```
✓ [Primary] Stored on provider 123
✓ [Secondary] Stored on provider 456
✓ [Primary] Piece added to Data Set (unconfirmed on-chain)
✓ [Secondary] Piece added to Data Set (unconfirmed on-chain)
✓ [Primary] Piece added to Data Set (confirmed on-chain)
✓ [Secondary] Piece added to Data Set (confirmed on-chain)
✓ [1/2] Advertisement confirmed indexed on https://sp-a.example.com
✓ [2/2] Advertisement confirmed indexed on https://sp-b.example.com
✓ Advertisement confirmed indexed on all 2 providers
✓ IPNI provider records found. IPFS retrieval possible.
```

### Piece-sync timeout — Curio never reports synced *(expected output — traced from code, not yet observed live)*

```
✓ [Primary] Stored on provider 123
✓ [Primary] Piece added to Data Set (unconfirmed on-chain)
✓ [Primary] Piece added to Data Set (confirmed on-chain)
⚠ Advertisement not confirmed indexed in time.
  Reason
    Piece "bafkzcib..." not synced on "https://sp.example.com/pdp" after 20 attempts. Last observation: Piece not yet synced
```

### Mismatch — Curio confirms synced, but cid.contact still 404s *(expected output — traced from code, not yet observed live)*

```
✓ [Primary] Stored on provider 123
✓ [Primary] Piece added to Data Set (unconfirmed on-chain)
✓ [Primary] Piece added to Data Set (confirmed on-chain)
✓ Advertisement confirmed indexed
⚠ Storage provider reported sync, but a direct indexer lookup still disagrees
  Curio reported piece "bafkzcib..." as synced, but the confirming https://cid.contact check still failed: IPFS CID "bafkrei..." does not have expected IPNI ProviderResults after 1 attempt. Last observation: IPNI indexer request failed with status 404. Expected serviceURLs: [https://sp.example.com/pdp]. Actual multiaddrs in response: []
```

### Partial failure — multiple providers, one syncs, one times out *(expected output — traced from code, not yet observed live)*

```
✓ [Primary] Stored on provider 123
✓ [Secondary] Stored on provider 456
✓ [Primary] Piece added to Data Set (unconfirmed on-chain)
✓ [Secondary] Piece added to Data Set (unconfirmed on-chain)
✓ [Primary] Piece added to Data Set (confirmed on-chain)
✓ [Secondary] Piece added to Data Set (confirmed on-chain)
✓ [1/2] Advertisement confirmed indexed on https://sp-a.example.com
⚠ Advertisement not confirmed indexed in time.
  Piece "bafkzcib..." not synced on "https://sp-b.example.com" after 20 attempts. Last observation: Piece not yet synced
```

src/core/utils/validate-ipni-advertisement.ts
```diff
@@ -38,6 +38,11 @@ interface ProviderResult {
   }
 }
 
+/** Response shape from Curio's `GET /pdp/piece/{pieceCid}/status`. Only `synced` is used. */
+interface PdpPieceStatusResponse {
+  synced?: boolean
+}
```
SgtPooki, line 44: doesn't synapse-sdk already do this? Or did they remove that functionality?

src/core/utils/validate-ipni-advertisement.ts
```diff
@@ -205,7 +210,27 @@ export async function waitForIpniProviderResults(
   }
 }
 
-async function waitForIpniProviderResultsForCid(
+/** @deprecated Use {@link CheckIpniIndexerOptions}. */
+export type WaitForIpniProviderResultsOptions = CheckIpniIndexerOptions
```
SgtPooki: we are already breaking things... so feel free to tell your AIs that breaking changes can happen when we are doing breaking changes (but they should be necessary and important)

one thought however, we still have to wait for IPNI provider results.. we're just waiting after we wait for providers themselves first, right.. so, do we need this rename?

SgtPooki (commented):
just looked at some of the code since it was relevant to #439.. going to open that PR back up..

src/core/utils/validate-ipni-advertisement.ts
beck-8: Ah? There shouldn't be, this syncd feature has just been implemented in Curio.

src/core/utils/validate-ipni-advertisement.ts
beck-8: Yes, but there is no way for this background. All that's left is cid.contact, which has a negative cache. My original idea here is that this method must be entered by the user, instead of having a default value as before. If there is a default value and the user has not confirmed too much before, it will definitely encounter a negative cache.

So... I had him change the name, and in the previous code writing sessions, he directly deleted this part. He didn't add a deprecated statement, so I still chose to add it temporarily later on.

src/core/utils/validate-ipni-advertisement.ts
SgtPooki: you might be right about the specific field in piece/status

src/core/upload/index.ts
```diff
@@ -349,27 +350,37 @@ export async function executeUpload(
...
+      case 'stored': {
+        pieceCidForValidation ??= event.data.pieceCid
+        break
+      }
       case 'piecesAdded': {
-        // Begin IPNI validation on the first piecesAdded event
+        // Begin indexing confirmation on the first piecesAdded event
         if (options.ipniValidation?.enabled !== false && ipniValidationPromise == null) {
+          if (pieceCidForValidation == null) {
+            logger.warn('piecesAdded fired before stored; skipping indexing confirmation (no pieceCid to poll)')
```
SgtPooki: question: is the `stored` before `piecesAdded` ordering guaranteed by the SDK?

If it is not, this path returns without setting `ipniValidationPromise`, so `ipniValidated` stays false at line 457 and no progress event is emitted at all. The CLI renders no indexing section, and the user cannot tell "not indexed" from "we never checked".

If the order is guaranteed, a comment saying so would help. If it is not, prefer deriving the pieceCid from the upload result rather than from event order, or emit a distinct skipped event so the outcome is visible.

src/core/utils/validate-ipni-advertisement.ts
```diff
@@ -205,7 +210,27 @@ export async function waitForIpniProviderResults(
...
+
+/**
+ * @deprecated Use {@link checkIpniIndexer} — same behavior, renamed because it's now one
+ * of two exports here rather than the sole post-upload check. `ipniIndexerUrl` is now
+ * required (previously defaulted to filecoinpin.contact, now retired): pass it explicitly.
+ */
+export const waitForIpniProviderResults = checkIpniIndexer
```
SgtPooki: blocking: this alias does not preserve the old call signature. `ipniIndexerUrl` is required at line 115, so every existing caller that relied on the previous default now fails to compile.

#665 assumes otherwise:

> filecoin-pin-website also depends on this function (`src/hooks/use-ipni-check.ts` calls `waitForIpniProviderResults` directly, with no explicit `ipniIndexerUrl` override), so bumping its filecoin-pin dependency should pick up the new behavior automatically.

It will not pick anything up. It gets a type error, and the obvious fix is to pass `https://cid.contact`, which gives 20 attempts at 5s intervals against cid.contact for a root CID. That is the exact negative-cache path #664 exists to remove, so the alias currently points callers at the one function that is no longer safe to call naively.

The website only has a root CID, so it cannot call `waitForIndexingConfirmation` either, since that needs a pieceCid. Options in order of preference:

1. Give the alias a real body that throws with a migration message naming `waitForIndexingConfirmation`.
2. Add a root-CID-only entry point that runs the Curio-first flow.
3. Amend #665 to say the named consumers (website, dealbot) must migrate by hand, and track that migration.

src/core/utils/validate-ipni-advertisement.ts
```diff
@@ -476,3 +508,359 @@ function deriveExpectedUris(
...
+  if (!response.ok) {
+    throw new Error(`${requestLabel} to "${url}" failed with status ${response.status}`)
+  }
+  return (await response.json()) as T
+}
+
+async function fetchPieceSynced(serviceURL: string, pieceCid: PieceCID | CID, signal: AbortSignal | undefined) {
+  const url = `${serviceURL.replace(/\/+$/, '')}/pdp/piece/${encodeURIComponent(pieceCid.toString())}/status`
+  const body = await fetchJson<PdpPieceStatusResponse>(url, signal, 'Piece status request')
+  return body.synced === true
```
SgtPooki: blocking: against a Curio that predates curio#1405, `body.synced` is undefined, so this returns false. `waitForPieceSynced` then treats it as "Piece not yet synced" and retries 20 times over 100 seconds before throwing, and the catch in `waitForIndexingConfirmation` rethrows without falling back.

Net effect on an SP that has not upgraded: 100 seconds burned, the indexer is never queried at all, and `ipniValidated` is false for every upload even when the content is indexed. That is the state of the entire fleet until SPs ship curio#1405.

Suggest distinguishing "endpoint not supported" (404, or 200 with `synced` absent) from "not synced yet", and on the former falling back to the bounded legacy indexer query for the transition, with a warn saying the check is degraded.

src/core/utils/validate-ipni-advertisement.ts
```diff
@@ -476,3 +508,359 @@ function deriveExpectedUris(
...
+      lastFailureReason = getErrorMessage(error)
+      options?.logger?.warn({ error, serviceURL }, `${lastFailureReason}. Retrying...`)
+    }
+
+    if (++retryCount >= maxAttempts) {
+      throw new Error(
+        `Piece "${pieceCid.toString()}" not synced on "${serviceURL}" after ${maxAttempts} attempt${maxAttempts === 1 ? '' : 's'}. Last observation: ${lastFailureReason}`
+      )
+    }
+    await new Promise((resolve) => setTimeout(resolve, delayMs))
```
SgtPooki: non-blocking: `signal` is in scope here, so this can reject on abort instead of riding out `delayMs`.

src/test/unit/validate-ipni-advertisement.test.ts
```diff
@@ -620,3 +654,330 @@ describe('waitForIpniProviderResults', () => {
     })
   })
 })
+
+describe('waitForIpniProviderResults (deprecated alias)', () => {
+  it('is the same function as checkIpniIndexer', () => {
```
SgtPooki: let's delete this test..

src/core/utils/validate-ipni-advertisement.ts
SgtPooki: I think you may be misunderstanding me.

1. We should not change the function and type names at all. its fine that a name is the same as a previous one, though now with a required argument.
2. this PR is already marked as a breaking change, so anyone getting this will need to make changes anyway; doing `@deprecated` is adding twice the work for maintainers in this, and all consuming repos.

it just seems like the AI got out of hand trying to get things too perfect here.. when the actual fixes here should be pretty straightforward and small.

src/core/utils/validate-ipni-advertisement.ts
beck-8: @SgtPooki  Normally, all our curio will be upgraded before this version is released 

It does not include people who are not actively upgraded. But those SPs should have almost excluded them.

src/index-types.ts
```diff
@@ -50,7 +50,11 @@ export type {
...
 } from './core/upload/index.js'
 export type { AnyProgressEvent, ProgressEvent, ProgressEventHandler, Warning } from './core/utils/types.js'
+// IndexerMismatchError is a runtime class, not a type; import it from 'filecoin-pin/core/utils'.
 export type {
+  CheckIpniIndexerOptions,
+  IndexingConfirmationProgressEvents,
   ValidateIPNIProgressEvents,
-  WaitForIpniProviderResultsOptions,
+  WaitForIndexingConfirmationOptions,
+  /** @deprecated Use {@link CheckIpniIndexerOptions}. */
```
BigLep: Dangling JSDoc comment — this `@deprecated` block is not attached to any export (the export list ends right after it with just `} from ...`). Looks like a leftover from `6c19583 refactor(ipni)!: drop waitForIpniProviderResults alias`, which removed the deprecated `WaitForIpniProviderResultsOptions` alias this comment used to annotate. Harmless to TS/JS, but confusing to a reader — worth deleting.

documentation/behind-the-scenes-of-adding-a-file.md
```diff
@@ -113,9 +113,9 @@ This is a function of the CAR size and the throughput between the client and the
...
 
-Filecoin Pin validates the IPNI advertisement process by polling `https://filecoinpin.contact/cid/$cid` (NOT cid.contact due to [negative caching issues discussed below](#how-long-does-an-ipni-indexer-cache-results)). 
+Filecoin Pin validates the IPNI advertisement process in two steps: first it polls the SP's `GET /pdp/piece/{pieceCid}/status` endpoint until it reports `synced: true` (the SP checks the indexer's own sync-status endpoint on Filecoin Pin's behalf here — see curio#1450); then, once synced, it makes a single confirming query to `https://cid.contact/cid/$cid` to verify the expected provider actually shows up. Querying cid.contact directly before that confirmation is avoided due to [negative caching issues](content-routing-faq.md#how-long-does-an-ipni-indexer-cache-results).
```
BigLep:
```suggestion
Filecoin Pin validates the IPNI advertisement process in two steps: first it polls the SP's `GET /pdp/piece/{pieceCid}/status` endpoint until it reports `synced: true`. (Behind the scenes, the SP checks the IPNI instance's own sync-status endpoint on Filecoin Pin's behalf.  This was added in [curio#1450](https://github.com/filecoin-project/curio/pull/1450)). Then, once synced, Filecoin Pin makes a single confirming query to `https://cid.contact/cid/$cid` to verify the expected provider actually shows up. Querying cid.contact directly before that confirmation is avoided due to [negative caching issues](content-routing-faq.md#how-long-does-an-ipni-indexer-cache-results).
```

documentation/content-routing-faq.md
```diff
@@ -23,16 +23,7 @@ In the event that an SP wipes their existing index state, the previously announc
 
 ## How long does an IPNI indexer cache results?
 
-This depends on both the [IPNI](glossary.md#ipni) indexer instance (e.g., cid.contact, filecoinpin.contact) and whether there is a cache hit or cache miss.
+[cid.contact](http://cid.contact) tends to cache hits for multiple hours and cache misses (negative cache) for minutes.  As a result of this, there are "gotchas" we have to be careful to avoid or can unavoidably fall into.
```
BigLep:
```suggestion
[cid.contact](http://cid.contact) tends to cache hits for multiple hours and cache misses (negative cache) for minutes. (cid.contact caching teraform is [here](https://github.com/ipni/storetheindex/blob/main/deploy/infrastructure/prod/us-east-2/cloudfront.tf).)  As a result of this, there are "gotchas" we have to be careful to avoid or can unavoidably fall into.
```

documentation/content-routing-faq.md
```diff
@@ -23,16 +23,7 @@ In the event that an SP wipes their existing index state, the previously announc
...
+- cid.contact cache miss "gotcha" - Because cid.contact caches misses (i.e., negative cache), it's important for Filecoin Pin to not poll `GET /cid/{cid}` aggressively before the advertisement has actually been processed.  The act of polling could cause the empty result set to get cached for minutes.  Instead, Filecoin Pin first polls the storage provider's `GET /pdp/piece/{pieceCid}/status` until it reports `synced: true` — the SP checks the indexer's own sync-status endpoint on Filecoin Pin's behalf here, which isn't a CID lookup and isn't subject to the same negative caching, so it's safe for the SP to do that patiently.  Only once the SP confirms sync does Filecoin Pin make a single confirming `GET /cid/{cid}` query, to confirm the expected provider's record actually shows up.  If a provider never reports `synced: true`, Filecoin Pin reports failure without ever falling back to `GET /cid/{cid}` — the SP already did the patient, safe check on its end, so a further `/cid/{cid}` fallback would be redundant and risk the exact negative-cache problem this whole flow exists to avoid.
```
BigLep:
```suggestion
- cid.contact cache miss "gotcha" - Because cid.contact caches misses (i.e., negative cache), it's important for Filecoin Pin to not poll `GET /cid/{cid}` before the advertisement has actually been processed.  The act of polling could cause the empty result set to get cached for minutes.  Instead, Filecoin Pin first polls the storage provider's `GET /pdp/piece/{pieceCid}/status` until it reports `synced: true`. The SP checks the IPNI instance's own sync-status endpoint on Filecoin Pin's behalf, which isn't a CID lookup and isn't subject to the same negative caching, so it's safe for the SP to do that patiently.  Only once the SP confirms sync does Filecoin Pin make a single confirming `GET /cid/{cid}` query, to confirm the expected provider's record actually shows up.  If a provider never reports `synced: true`, Filecoin Pin reports failure without ever falling back to `GET /cid/{cid}`. The SP already did the patient, safe check on its end, so a further `/cid/{cid}` fallback would be redundant and risk the exact negative-cache problem this whole flow exists to avoid.
```

BigLep (commented):
I'll leave to @SgtPooki to approve and merge. These are my few comment changes.

src/index-types.ts
SgtPooki: thanks.. deleted

SgtPooki (approved):

beck-8 (commented):
These robustness fixes look good

Issue #697, opened 2026-09-02
feat(cli): login command with console session pairing


SgtPooki:
### Description

Using filecoin-pin today means pasting a wallet private key into the shell. That key controls every fund in the wallet and never expires, and agents driving the CLI have no other way to get credentials.

`filecoin-pin login` replaces that with a scoped session key: the CLI generates the key, opens the Filecoin Cloud console for one wallet approval, polls `SessionKeyRegistry` for the grant, and prints a readiness scorecard. No callback server. Auth and funding stay separate steps; funding never blocks login.

PRD: https://app.notion.com/p/filecoindev/Login-Auth-3cadc41950c180a6b663f8024e7ca423 (v2, 2026-08-27). Sections 6, 8, 9 are the UX contract, scope, and acceptance criteria. This issue tracks the engineering work; it does not restate the PRD.

Ground rules from the PRD: the code is a self-contained module inside filecoin-pin, and no existing data command changes syntax or behavior.

One change to the PRD: decision #14 named `~/.filecoin-session-key.env`. The session env file goes in the existing filecoin-pin data dir instead, next to the other per-user state.

### Impact

Blocks agent onboarding and any team with a key-handling policy.

### MVP scope (PRD section 8)

- `login`, `logout`, `balance`, `dashboard`
- credential auto-load: flags, then env vars, then the session env file in the filecoin-pin data dir (`getDataDirectory()` in `src/config.ts`, platform-specific), then the no-credentials error
- funds preflight in `add`
- per-command scope gating with add-scopes links

### Children

- [x] #670 `--scopes` on create/authorize/revoke
- [x] #671 `--env-file`
- [x] #687 per-command scope gating (depends on FilOzone/filecoin-pay-explorer#355)
- [x] `login` and `logout`
- [x] credential auto-load and the no-credentials error
- [x] `add` funds preflight
- [x] `balance` and `dashboard`

Console work is tracked in FilOzone/filecoin-pay-explorer#368.

Related, not blocking: #457 (credential storage). `--json` output (#4) is its own epic and is not a pre-requisite for login.

SgtPooki:
Closing this one as done.

PR #699, opened 2026-09-02
feat(session): add on-chain authorization watcher

SgtPooki:
Stacked on #687 (`feat/per-command-scope-gating`).

Adds `watchAuthorization` in `src/core/session/watch-authorization.ts`: the polling half of `filecoin-pin login`. It implements the two detection mechanisms in PRD section 4 ("How the CLI detects approval"): an `eth_getLogs` scan for `AuthorizationsUpdated` events matching our signer when the owner is unknown, and direct per-scope expiry reads (`getExpirations` from synapse-core) once the owner is known. Events are decoded with `extractLoginEvent`; nothing is re-implemented from the registry ABI.

The wait is bounded by a deadline, five minutes by default (PRD appendix A, decision 11). The result carries owner, expiry, and the granted and missing scopes, so the caller can print a requested-versus-granted diff (decision 13); `none` means the owner acted but granted no requested scope, and `timeout` is reserved for the deadline. Progress goes through the same `onProgress` event pattern as the other session helpers (`watch:tick`, `watch:ownerFound`, `watch:error`). A failed poll is reported and the wait continues; the error is thrown only when no poll ever succeeded.

Unit tests cover the event match, the timeout, a partial grant, a grant of none, the known-owner path (never replaced by an event), a transient RPC error, an RPC endpoint that never answers, and a lapsed expiry.

Part of #697. No CLI surface changes in this PR; `login` itself comes in the next one.

### DIFF
--- a/src/core/session/watch-authorization.ts
+++ b/src/core/session/watch-authorization.ts
@@ -1,0 +1,219 @@
+/**
+ * Wait for a session key to be authorized on-chain.
+ *
+ * `filecoin-pin login` prints a console link, and the wallet owner approves
+ * the key in the browser. Nothing calls back to the CLI: the registry
+ * contract is the shared state, so the CLI polls it.
+ *
+ * Two ways to detect the grant:
+ *
+ *  1. Owner unknown (first pairing): poll `eth_getLogs` for the registry's
+ *     `AuthorizationsUpdated` events from the block login started at, and
+ *     match the `signer` field to our key. One small, bounded query per
+ *     tick. The matching event names the owner.
+ *  2. Owner known (renewal, or after the event was found): read the
+ *     per-scope expiries directly. No events involved.
+ *
+ * Both paths end in the per-scope read, so partial grants (the owner
+ * declined a scope in the console) are reported scope by scope.
+ */
+
+import { type Expirations, extractLoginEvent, getExpirations, type Permission } from '@filoz/synapse-core/session-key'
+import { type Address, type Chain, type Client, isAddressEqual, type Log, type Transport, toEventSelector } from 'viem'
+
+/** Default poll deadline: about five minutes (PRD decision #11). */
+export const DEFAULT_WATCH_DEADLINE_MS = 5 * 60 * 1000
+/** Default interval between registry polls. */
+export const DEFAULT_WATCH_INTERVAL_MS = 5000
+
+const AUTHORIZATIONS_UPDATED_TOPIC = toEventSelector(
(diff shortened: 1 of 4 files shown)


src/core/session/watch-authorization.ts
```diff
@@ -0,0 +1,294 @@
...
+    last: { status: 'timeout', granted: [], missing: [...options.permissions] },
+    anySuccess: false,
+    lastError: undefined,
+  }
+
+  for (;;) {
+    const remainingMs = deadline - Date.now()
+    if (remainingMs <= 0) break
+    options.onProgress?.({ type: 'watch:tick', data: { remainingMs } })
+    if (await pollTolerantly(options, state)) return state.last
```
BravoNatalie, line 318: small edge case: `eth_getLogs` and `getExpirations` inside `pollTolerantly` don't have a per-poll timeout. If the endpoint never resolves, the loop won't reach the deadline check.

src/core/session/watch-authorization.ts
```diff
@@ -0,0 +1,294 @@
...
+  lastBlock?: bigint
+}
+
+async function findOwnerInLogs(
+  client: Client<Transport, Chain>,
+  registryAddress: Address,
+  sessionAddress: Address,
+  fromBlock: bigint
+): Promise<LogScan> {
+  const logs = (await client.request({
```
BravoNatalie, line 176: edge case: when the scan returns `[]`, `findOwnerInLogs()` returns `{}`, so `scanFrom` does not advance. That means each poll repeats the same `fromBlock..latest` scan

BravoNatalie (approved):
Just two small edge cases to handle, but other than that, it looks good!

src/core/session/watch-authorization.ts
SgtPooki: Agreed on tolerating one failure. viem's transport already times each request out at 10s, so a hung endpoint fails a poll rather than skipping the deadline check. What was missing was giving up on a dead endpoint: https://github.com/filecoin-project/filecoin-pin/commit/44a4615 ends the wait after 3 failed polls in a row with a resume hint. A success resets the count, so a single blip still gets through.

src/core/session/watch-authorization.ts
SgtPooki: Done in https://github.com/filecoin-project/filecoin-pin/commit/44a4615. An empty scan now moves scanFrom to eth_blockNumber, so the next poll starts at the head instead of the login block.

PR #700, opened 2026-09-02
feat(cli): add login and logout commands

SgtPooki:
Stacked on #699 (`feat/login-watcher`).

Adds `filecoin-pin login` and `filecoin-pin logout`, the `login` and `logout` rows of PRD section 5 and the login mocks in section 6.

`login` generates a session key or resumes the saved one, writes `session.env` (key, address, network) under the data directory with mode 0600 before the browser opens, prints the console pairing link on its own line (`/console/session-keys?authorize=<lowercase addr>&scopes=a,b&network=<mainnet|calibration>`), opens it on a TTY unless `--no-browser` or `BROWSER=none`, waits for the grant with a countdown, then prints the granted scopes and the readiness scorecard with a funding link when something is missing. `--scopes` (default `createDataSet,addPieces`), `--fresh`, `--no-wait` (print the link and exit 2), `--timeout <seconds>`. Exit 0 when the key can upload (every requested scope, or at least createDataSet and addPieces), 2 when incomplete, 1 on error.

Refused up front: a chain the console has no page for (devnet, custom RPC), and resuming a key made for another network. `login` warns when `PRIVATE_KEY`, `SESSION_KEY`, or `VIEW_ADDRESS` is set in the shell, since they shadow the saved key, and when `--fresh` replaces a key that stays live on chain. The browser opener spawns without credential variables in its environment.

`logout` deletes the file, prints the address it removed, and how to revoke it on chain. The `session` group's help now says it is the owner-signed, advanced path. `login --help` has examples, files, environment, and exit codes.

Tests: session file round trip, permissions, and network; the login runner (file before the wait, link contents, partial grant, resume, `--fresh`, `--no-wait`, `--no-browser`, `--timeout`, refused chain, network mismatch, shadow warnings); command wiring.

Part of #697.

PR #701, opened 2026-09-02
feat(cli): auto-load the login session and add the no-credentials error

SgtPooki:
Stacked on #700 (`feat/login-command`).

Implements credential resolution from PRD section 5 and the wall and expiry messages from section 6.

For every command the order is: explicit flags, then env vars (`PRIVATE_KEY`, or `SESSION_KEY` + `WALLET_ADDRESS`), then `--credentials-file`, then the session file `login` wrote, then the `No credentials found` wall pointing at `login`. `applySessionFileCredentials` runs in `cli.ts` before Commander parses and loads the file only when no auth flag is on argv and no auth env var is set. A file without an owner address is ignored. `login`, `logout`, `dashboard`, and `server` never auto-load.

The saved login's network applies when neither `--network` nor `NETWORK` chose one; using it under another network prints a warning naming the fix. The `Using session …` line names the key, its source, the owner, and the network. When `VIEW_ADDRESS` alone keeps a usable login out, the command says so. When `CI` is set and the saved login supplied the credentials, a stderr warning tells the job to set credentials explicitly.

A credentials file may set only `PRIVATE_KEY`, `SESSION_KEY`, `WALLET_ADDRESS`, `VIEW_ADDRESS`, and `NETWORK`; anything else is ignored and named, so a file from a console or a teammate cannot point `CONSOLE_URL` or `RPC_URL` elsewhere.

Expired grants fail with `Session expired` and point at `login`. Top-level help has a CREDENTIALS block with the full order.

Tests: env beats file, flags beat both, unfinished login ignored, network export and precedence, VIEW_ADDRESS skip, CI warning, allowlist, expired wording, wall text.

Part of #697.

### DIFF
--- a/src/core/synapse/index.ts
+++ b/src/core/synapse/index.ts
@@ -151,29 +151,21 @@ export function assertSessionKeyPrivateKey(value: string): asserts value is Hex 
 }
 
 /**
- * First line of the preflight error. Three shapes: nothing ever granted,
- * every needed grant lapsed (the session expired: renew it with `login`),
- * or a live key that lacks a scope.
+ * What the on-chain expirations say about a key that lacks a needed scope:
+ * nothing was ever granted, every grant lapsed (the session expired), or
+ * some grant is live but not the one this operation needs.
  */
-function describeProblem(
-  key: SessionKey<'Secp256k1'>,
-  ownerAddress: string,
-  scopeLabels: string,
-  networkName: string,
-  now: bigint
-): string {
+function classifyAuthorization(key: SessionKey<'Secp256k1'>, now: bigint): 'never' | 'expired' | 'missing' {
   const allExpirations = Object.values(key.expirations)
-  const neverAuthorized = allExpirations.length > 0 && allExpirations.every((expiry) => expiry === 0n)
-  if (neverAuthorized) {
-    return `Session key ${key.address} isn't authorized for account ${ownerAddress} on ${networkName} — never authorized, expired/revoked, or the key is for a different network (check --network).`
-  }
-  // No live grant at all, but at least one past grant: the session ran out.
-  if (allExpirations.every((expiry) => expiry <= now)) {
-    const latest = allExpirations.reduce((a, b) => (a > b ? a : b), 0n)
-    const date = new Date(Number(latest) * 1000).toISOString().slice(0, 10)
-    return `Session expired (key ${key.address}, grants lapsed ${date})\n  Renew it:  filecoin-pin login`
-  }
-  return `Session key ${key.address} lacks ${scopeLabels} for this operation on ${networkName}.`
+  if (allExpirations.length > 0 && allExpirations.every((expiry) => expiry === 0n)) return 'never'
+  if (allExpirations.every((expiry) => expiry <= now)) return 'expired'
+  return 'missing'
+}
+
+/** Latest expiry among the key's grants, as a date. */
+function latestExpiryDate(key: SessionKey<'Secp256k1'>): string {
+  const latest = Object.values(key.expirations).reduce((a, b) => (a > b ? a : b), 0n)
+  return new Date(Number(latest) * 1000).toISOString().slice(0, 10)
 }
 
 /**
@@ -198,6 +190,17 @@ function checkSessionKeyPermissions(
   const missing = required.filter((p) => !key.hasPermission(p))
   if (missing.length === 0) return
 
+  const now = BigInt(Math.floor(Date.now() / 1000))
+  const state = classifyAuthorization(key, now)
+  // The wall from PRD section 6: two lines, one remedy. Scope details and
+  // console links would only compete with "run login".
+  if (state === 'expired') {
+    throw new Error(
+      `Session expired (key ${key.address}, grants lapsed ${latestExpiryDate(key)})\n  Renew it:  filecoin-pin login`
+    )
+  }
+  const neverAuthorized = state === 'never'
+
   const scopeIds = missing.map(scopeIdOf)
   const scopeLabels = missing.map((p) => PermissionNames[p] ?? p).join(', ')
   const scopesArg = scopeIds.join(',')
@@ -205,7 +208,6 @@ function checkSessionKeyPermissions(
   // Per-scope detail preserves the expired-at vs never-granted distinction
   // the on-chain expirations carry — an expired grant points at renewal,
   // a never-granted scope points at a fresh authorization.
-  const now = BigInt(Math.floor(Date.now() / 1000))
   const scopeDetails = missing.map((p) => {
     const name = PermissionNames[p] ?? p
     const expiry = key.expirations[p] ?? 0n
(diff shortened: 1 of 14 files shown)


SgtPooki (commented):
self review

src/core/synapse/index.ts
```diff
@@ -150,6 +150,24 @@ export function assertSessionKeyPrivateKey(value: string): asserts value is Hex
...
+ * What the on-chain expirations say about a key that lacks a needed scope:
+ * nothing was ever granted, every grant lapsed (the session expired), or
+ * some grant is live but not the one this operation needs.
+ */
+function classifyAuthorization(key: SessionKey<'Secp256k1'>, now: bigint): 'never' | 'expired' | 'missing' {
+  const allExpirations = Object.values(key.expirations)
+  if (allExpirations.length > 0 && allExpirations.every((expiry) => expiry === 0n)) return 'never'
+  if (allExpirations.every((expiry) => expiry <= now)) return 'expired'
+  return 'missing'
+}
```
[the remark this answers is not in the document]
SgtPooki: Done in https://github.com/filecoin-project/filecoin-pin/commit/ef55acc. An empty map now reads as never granted. It should not reach this function, since the SDK fills one entry per requested scope, but the guard costs nothing.

src/utils/credential-source.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @returns the file path when it was used, otherwise undefined
+ */
+export function applySessionFileCredentials(
+  argv: readonly string[] = process.argv,
+  env: NodeJS.ProcessEnv = process.env,
+  path: string = getSessionFilePath()
+): string | undefined {
+  if (isSkippedCommand(argv) || hasFlag(argv, AUTH_FLAGS)) return undefined
+  if (hasAuthEnv(env)) {
```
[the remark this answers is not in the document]
SgtPooki: Done in https://github.com/filecoin-project/filecoin-pin/commit/ef55acc. The three module variables are reset at the top of every call.

src/test/smoke/utils.ts
```diff
@@ -1,4 +1,6 @@
 import { spawn } from 'node:child_process'
+import { mkdtempSync } from 'node:fs'
```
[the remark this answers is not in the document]
SgtPooki: Done in https://github.com/filecoin-project/filecoin-pin/commit/ef55acc. The smoke helper removes its data directory on exit.

src/utils/credential-source.ts
```diff
@@ -0,0 +1,117 @@
...
+    // command can say why the saved login was not used.
+    const onlyViewAddress = isSet(env, 'VIEW_ADDRESS') && !AUTH_ENV_VARS.slice(0, 3).some((n) => isSet(env, n))
+    skippedForViewAddress = onlyViewAddress && readSessionFile(path)?.walletAddress !== undefined
+    return undefined
+  }
+  const session = readSessionFile(path)
+  if (session?.walletAddress === undefined) return undefined
+  env.SESSION_KEY = session.sessionKey
+  env.WALLET_ADDRESS = session.walletAddress
+  if (session.network !== undefined && !isSet(env, 'NETWORK') && !hasFlag(argv, NETWORK_FLAGS)) {
```
BravoNatalie, line 89: `hasFlag` only check the `argv` and  `NETWORK_FLAGS` contains `['--network', '--rpc-url']`, so it does not check `RPC_URL`. Could we check `RPC_URL` here too? 

With `RPC_URL` set in the shell and a saved login, this adds the saved `NETWORK`. Commander then sees both environment variables and exits with a conflict. We should give priority to the user's provided env/parameter.

src/utils/credentials-file.ts
```diff
@@ -44,18 +55,30 @@ export function loadCredentialsFile(path: string, env: NodeJS.ProcessEnv = proce
...
-  if (Object.keys(parsed).length === 0) {
+  const ignored: string[] = []
+  let usable = 0
+  for (const [key, value] of Object.entries(parsed)) {
+    if (!(CREDENTIALS_FILE_VARS as readonly string[]).includes(key)) {
+      ignored.push(key)
+      continue
+    }
+    usable += 1
+    if (env[key] === undefined) {
```
BravoNatalie, line 66: I don't think this preserve precedence across authentication modes.

For example:

`PRIVATE_KEY=<wallet-a-key> filecoin-pin add file.txt --credentials-file team-credentials`

And suppose team-credentials contains:
```
  SESSION_KEY=<session-key-for-wallet-b>
  WALLET_ADDRESS=<wallet-b>
```

these values are all presented to Commander as environment values. `parseCLIAuth()` then selects _session-key mode_ before _private-key mode_, so the lower-priority file wins. 

I think we need to retain the file provenance, or avoid loading file auth when the shell already provides a complete auth mode.

suggestion: add an end-to-end precedence test to catch this.

BravoNatalie (approved):
I’m approving it to avoid blocking you, but I left two comments.

src/utils/credential-source.ts
SgtPooki: Fixed in 2b32ab5. The session loader now runs from the `addAuthOptions` preAction hook instead of before parse, so "did the user choose a network" is `opts.network`/`opts.rpcUrl` (RPC_URL env is already folded into `opts.rpcUrl` by Commander) rather than a hand-maintained list of flags. Same check covers `NETWORK` in a credentials file with `RPC_URL` in the shell, which hit the same conflict.

`credential-precedence.test.ts` has the RPC_URL env, `--rpc-url`, and NETWORK env cases against a saved login.

src/utils/credentials-file.ts
SgtPooki: Went with retaining provenance, in 2b32ab5. Both file loaders moved into preAction hooks (root program for `--credentials-file`, `addAuthOptions` for the saved login) and each value they supply is set with `setOptionValueWithSource(..., 'file' | 'session')`. `resolveAuthMode` ranks `cli > env > file > session` instead of treating everything below a flag as env, so your example now picks `PRIVATE_KEY` and warns that the file's session pair was ignored.

The argv sniffing (`findCredentialsFileArg`, `isSkippedCommand`) is gone with it; `login`/`logout`/`server` never auto-load because they don't call `addAuthOptions`, and there's a test over the real command objects for that. 373d01c closes the one gap the hook opens: Commander checks `.conflicts()` before hooks run, so a file-sourced value is re-checked against them (`--auto-fund` with VIEW_ADDRESS in a file).

`credential-precedence.test.ts` drives a real Commander program with files on disk for every layer, including your exact `PRIVATE_KEY` + file case. I confirmed the old build fails that case.

SgtPooki:
Merged to master on 2026-09-11 as [`7e431c2`](https://github.com/filecoin-project/filecoin-pin/commit/7e431c2be77584fb8e74c73b476300b7db13d30e), but GitHub dropped the merge event and left this PR open. Proof, run against a fresh `git fetch`:

```
$ git branch -r --contains 7e431c2
  origin/master
$ git diff --quiet 373d01c 7e431c2 && echo identical
identical
```

The full tree of this branch's final head [`373d01c`](https://github.com/filecoin-project/filecoin-pin/commit/373d01c2bdba6938a589d3dc90166343968564ef) is byte-identical to the squash commit on master (no path filter), and the squash's diff against its parent [`819dc6a`](https://github.com/filecoin-project/filecoin-pin/commit/819dc6a5c59b0805260d2ce2a312dda0c78ed5b6) is the whole PR. It is in the 2.1.0 release notes on #715, and master CI is green on it as of [`f225a37`](https://github.com/filecoin-project/filecoin-pin/commit/f225a37da85c031a5dfe5d9d37bf31d060b12317). Closing.

PR #702, opened 2026-09-02
feat(add): funds preflight for session keys

SgtPooki:
Stacked on #701 (`feat/login-autoload`).

Implements the funds preflight from PRD section 6 for the session-key path, on `add` and `import`.

Before `add` packs anything (and `import` before uploading) in session-key mode, it estimates the input size, asks the SDK for the upload cost, and reads service approval and available funds. If the service is not approved or the estimate is not covered, it prints the readiness lines with available versus needed, a funding link on its own line pre-filled with the shortfall rounded up to whole USDFC (`/console?deposit=<n>&operator=fwss&network=<mainnet|calibration>`), and the command to rerun, then exits 1 with nothing uploaded. `add` checks again on the packed size before upload.

The rerun hint is rebuilt from argv with `--session-key` and `--private-key` values redacted, so a preflight failure never prints a key into a log.

Session keys cannot deposit, so there is no auto-fund on this path. Private-key auth is untouched. `add --help` and `import --help` say the preflight exists and what exit 1 means.

Tests: pass-through, shortfall block, approval-missing case, directory size estimate, rerun hint redaction.

Part of #697.


SgtPooki (commented):
self review.

src/add/add.ts
```diff
@@ -290,6 +303,11 @@ export async function runAdd(options: AddOptions): Promise<AddResult | AddDryRun
       }
 
       await performAutoFunding(synapse, carSize, spinner, autoFundOptions)
+    } else if (isSessionKeyMode(synapse)) {
```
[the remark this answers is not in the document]
SgtPooki: Good catch. Done in https://github.com/filecoin-project/filecoin-pin/commit/2305c87: session mode is checked before --auto-fund in the second check too, matching the first check and import, with a test that auto-funding is never attempted with a session key.

src/common/funds-preflight.ts
```diff
@@ -0,0 +1,122 @@
...
+ * is for the funds check, not the bill.
+ */
+export async function estimateInputBytes(path: string, isDirectory: boolean, includeHidden = false): Promise<number> {
+  if (!isDirectory) return (await stat(path)).size
+  const entries = await readdir(path, { recursive: true, withFileTypes: true })
+  let total = 0
+  for (const entry of entries) {
+    if (!entry.isFile()) continue
+    // The packer skips dotfiles unless --include-hidden; count what it will pack.
+    if (!includeHidden && entry.name.startsWith('.')) continue
```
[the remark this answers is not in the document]
SgtPooki: Leaving this one. The estimate is for the funds check, not the bill, and an overestimate only refuses an upload that was near the line; the second check on the packed CAR size is exact. If a real tree with a big hidden subtree gets refused, that is the moment to match the packer's rule.

src/common/funds-preflight.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * The command to run again after funding, rebuilt from argv with any secret
+ * flag value redacted: the hint lands in CI logs and agent transcripts.
+ */
+export function rerunHint(argv: readonly string[] = process.argv): string {
+  const words: string[] = []
+  for (let i = 2; i < argv.length; i++) {
+    const arg = argv[i] as string
+    const flag = SECRET_FLAGS.find((f) => arg === f || arg.startsWith(`${f}=`))
```
BravoNatalie, line 51: Shouldn’t we redact the `rpc-url` values too? They can sometimes contain API keys.

src/common/funds-preflight.ts
```diff
@@ -0,0 +1,122 @@
...
+const USDFC_WEI = 10n ** 18n
+
+/**
+ * Bytes the upload will roughly carry: the file's size, or the sum of the
+ * files under a directory. CAR framing adds a little on top; the estimate
+ * is for the funds check, not the bill.
+ */
+export async function estimateInputBytes(path: string, isDirectory: boolean, includeHidden = false): Promise<number> {
+  if (!isDirectory) return (await stat(path)).size
+  const entries = await readdir(path, { recursive: true, withFileTypes: true })
```
BravoNatalie, line 30: this can take too long and consume a lot of memory if the user has a very large directory. 
Maybe consider an async dirwalker using `opendir()`?

src/common/funds-preflight.ts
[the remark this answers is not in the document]
BravoNatalie: Is this really ok?
What if, for example, the user is submitting a code repository and `node_modules` is much larger than the actual code?

There’s also the reverse case. If the user explicitly runs `filecoin-pin add .hidden-folder`, the packer intentionally includes everything inside that folder.

BravoNatalie (commented):
I’m approving it to avoid blocking you, but I left a few comments.

BravoNatalie (approved):

src/common/funds-preflight.ts
SgtPooki: Added in 203ea69.

src/common/funds-preflight.ts
SgtPooki: Switched to an `opendir()` walk in 203ea69 and e5e7dc6: one level at a time, and each directory's handle closes before its subdirectories are walked, so depth never stacks handles. Also made it skip dot-named directories, since the packer's glob does; the old count included `.git/`.

SgtPooki:
Merged to master on 2026-09-11 as [`40389ae`](https://github.com/filecoin-project/filecoin-pin/commit/40389ae50b6c5d3640cb5fb7ee3d6ff98c9bd8f5), but GitHub dropped the merge event and left this PR open. Proof, run against a fresh `git fetch`:

```
$ git branch -r --contains 40389ae
  origin/master
$ git diff --quiet e5e7dc6 40389ae && echo identical
identical
```

The full tree of this branch's final head [`e5e7dc6`](https://github.com/filecoin-project/filecoin-pin/commit/e5e7dc6209db85bbb03aebcf312070105d5c4c31) is byte-identical to the squash commit on master (no path filter), and the squash's diff against its parent [`7e431c2`](https://github.com/filecoin-project/filecoin-pin/commit/7e431c2be77584fb8e74c73b476300b7db13d30e) is the whole PR. It is in the 2.1.0 release notes on #715, and master CI is green on it as of [`f225a37`](https://github.com/filecoin-project/filecoin-pin/commit/f225a37da85c031a5dfe5d9d37bf31d060b12317). Closing.

PR #669, opened 2026-08-25
chore(master): release 2.0.0

FilOzzy:
📦 Release Preparation
---

### ⚠ BREAKING CHANGES

* confirm indexing via Curio piece-status ([#689](https://github.com/filecoin-project/filecoin-pin/issues/689))
* **deps:** update synapse-sdk to 1.2.1 and core to 0.8.1 ([#686](https://github.com/filecoin-project/filecoin-pin/issues/686))
* stop exposing piece metadata ([#683](https://github.com/filecoin-project/filecoin-pin/issues/683))

### Features

* **cli:** add --credentials-file to load dotenv-style credentials ([#671](https://github.com/filecoin-project/filecoin-pin/issues/671)) ([31a6d82](https://github.com/filecoin-project/filecoin-pin/commit/31a6d821ddc1d004bc8edb7218bf8bd98617f9f8))
* confirm indexing via Curio piece-status ([#689](https://github.com/filecoin-project/filecoin-pin/issues/689)) ([0933c5f](https://github.com/filecoin-project/filecoin-pin/commit/0933c5f3b585095878946ef2dabe261ffcf38670))
* **session:** --scopes flag on create/authorize/revoke ([#670](https://github.com/filecoin-project/filecoin-pin/issues/670)) ([2e9cb54](https://github.com/filecoin-project/filecoin-pin/commit/2e9cb54a333028c87a503b01f5f782149c8a8649))
* **session:** per-command scope gating with console-first remediation ([#687](https://github.com/filecoin-project/filecoin-pin/issues/687)) ([24058ab](https://github.com/filecoin-project/filecoin-pin/commit/24058aba70243ee4cd89ac87b505b2351efd9174))
* stop exposing piece metadata ([#683](https://github.com/filecoin-project/filecoin-pin/issues/683)) ([17127f8](https://github.com/filecoin-project/filecoin-pin/commit/17127f84d75e286646c9c01bedd02ffb4969464c))

### Bug Fixes

* **deps:** update synapse-sdk to 1.2.1 and core to 0.8.1 ([#686](https://github.com/filecoin-project/filecoin-pin/issues/686)) ([3e3e7d7](https://github.com/filecoin-project/filecoin-pin/commit/3e3e7d7ab543d2503751155a60c79d0c854f5702))

### Chores

[...]

PR #707, opened 2026-09-03
ci: validate PR titles and commits for release-please

SgtPooki:
### What changed

Two commits were missing from the 2.0.0 release notes (#669): #658 was titled `feat!(cli):` instead of `feat(cli)!:`, and #705 wrote its footer as `BREAKING CHANGE (Windows):`. release-please skips what it cannot parse and does not report it.

This adds a workflow with two jobs and no custom code:

1. `amannn/action-semantic-pull-request` checks the PR title (what squash merge publishes). Pinned to the v6.1.1 commit. It is a node24 action with committed `dist/`; its only write path (`POST /statuses`) is behind the `wip` input, which we do not set.
2. commitlint checks every commit (what rebase merge publishes) via its documented CI setup: `npx` with `@commitlint/cli@21.2.2` and `@commitlint/config-conventional@21.2.2`. I did not use `wagoid/commitlint-github-action` because it is a Docker action whose `action.yml` pulls a mutable Docker Hub tag, so a commit pin would not pin what runs.

Both reject the `feat!(cli):` case. Neither catches `BREAKING CHANGE (Windows):`; commitlint accepts that as a valid note. I did not find an action that checks footer format the way release-please parses it, so that case stays a review item.

### Notes

The `master` ruleset requires no status checks, so this is advisory until an admin adds `pr-title` and `commits` as required.

Refs #669


SgtPooki (commented):
self review

BigLep (approved):

hugomrdias (approved):

SgtPooki:
Ok, did the admin job: 
image

PR #483, opened 2026-05-14
fix: get detailed data set avoid full scan

beck-8:
The command will then work normally.
```
PS D:\workspace\filecoin-pin> node .\dist\cli.js data-set show 4503 --view-address 0xa108be4331296ec8b8c47c2cd2fbfddf06e27523 --network calibration
┌  Filecoin Onchain Cloud Data Set Details for #4503
│
◇  ━━━ Data Set ━━━
│
│  Network: Filecoin - Calibration testnet
│  Client address: 0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│
│  #4503
│    Status: live
│    CDN add-on: disabled
│
│    Provider
│      ID: 14
│      Address: 0xAfdd652756bAe790559cD07a96Caee310D3C6381
│      Name: calib-test
│      Description: calibnet pdp
│      Service URL: http://116.182.28.4:12310
│      Active: yes
│      Location: C=US;ST=California;L=San Francisco
│
│    Metadata
│      ipfsRootCID: "bafkreidbt7pxa72j6kdnwnjsay34bye2nkw5aovqvasks4p23lzeit5vzm"
│      withIPFSIndexing: ""
│
│    Payment
│      PDP rail ID: 4856
│      Payer: 0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│      Payee: 0xAfdd652756bAe790559cD07a96Caee310D3C6381
│      PDP payments ended @ epoch 3385433
│
│    Pieces
│      Total pieces: 1
│      Total size: 6.2 KiB
│      Unique PieceCIDs: 1
│      Unique IPFS Root CIDs: 0
│
│      #0 (active)
│        PieceCID: bafkzcibdraharbp3xgmlpp3u72lipowjsjsymmeiduvp4oohaf427vlzgmzrk4rl
│        Size: 6.2 KiB
│        Metadata
│          none
│
│
└  Data set inspection complete

```

PR #708, opened 2026-09-04
fix(data-set): paginate listDataSets to avoid RPC timeouts on large accounts

### LINKED ISSUE #362: Too many data sets make many commands unusable
`0xa108be4331296ec8b8c47c2cd2fbfddf06e27523`

This wallet was used by me before for stress testing. It has a large number of data sets. I'm not sure exactly how much right now.
I obtained it through the contract method `clientDataSets` and now there are 4214

Now it seems that this method `getClientDataSets` cannot hold so many entries?

```
➜  /tmp filecoin-pin add out2287372121
┌  Filecoin Pin Add
│
◇  ✓ File validated (102.0 B)
│
◇  ✓ Connected to Filecoin - Calibration testnet
│
◇  ✓ Minimum payment setup verified (~0.066 USDFC required)
│
◇  ✓ File packed with root CID: bafybeihpws5pknsqfpfmqufxchoww4m5ry5jidy6s3zu22ac2v63tkediq
│
◇  ✓ IPFS content loaded (297.0 B)
│
◇  ✓ Payment capacity verified for 297.0 B
│
◇  ✗ Add failed: The contract function "getClientDataSets" reverted with the following reason:
message execution failed (exit=[33], revert reason=[none], vm error=[message failed with backtrace:
00: f0179532 (method 3844450837) -- contract reverted at 943 (33)
01: f0176641 (method 3844450837) -- out of gas (7)
02: f0176641 (method 6) -- out of gas (7)
 (RetCode=33)])

Contract Call:
  address:   0x53d235D474585EC102ccaB7e0cdcE951dD00f716
  function:  getClientDataSets(address client)
  args:                       (0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523)

Docs: https://viem.sh/docs/contract/readContract
Version: viem@2.47.4



➜  /tmp filecoin-pin data-set ls --all --rpc-url http://127.0.0.1:1234/rpc/v1
┌  Filecoin Onchain Cloud Data Sets
│
◇  ✗ Failed to list data sets
│
│
│  Error: Failed to get details for data set 11184: HTTP request failed.
│
│  URL: http://127.0.0.1:1234/rpc/v1
│  Request body: {"method":"eth_call","params":[{"data":"0xbdaac0560000000000000000000000000000000000000000000000000000000000002bb0","from":"0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523","to":"0x53d235D474585EC102ccaB7e0cdcE951dD00f716"},"latest"]}
│
│  Raw Call Arguments:
│    from:  0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│    to:    0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    data:  0xbdaac0560000000000000000000000000000000000000000000000000000000000002bb0
│
│  Contract Call:
│    address:   0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    function:  getDataSet(uint256 dataSetId)
│    args:                (11184)
│
│  Docs: https://viem.sh/docs/contract/readContract
│  Details: fetch failed
│  Version: viem@2.47.4
└  Listing failed


➜  /tmp filecoin-pin data-set show 11174
┌  Filecoin Onchain Cloud Data Set Details for #11174
│
◇  ✗ Failed to inspect data set
│
│
│  Error: Failed to get details for data set 2811: The request took too long to respond.
│
│  URL: wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v1
│  Request body: {"method":"eth_call","params":[{"data":"0xbdaac0560000000000000000000000000000000000000000000000000000000000000afb","from":"0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523","to":"0x53d235D474585EC102ccaB7e0cdcE951dD00f716"},"latest"]}
│
│  Raw Call Arguments:
│    from:  0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│    to:    0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    data:  0xbdaac0560000000000000000000000000000000000000000000000000000000000000afb
│
│  Contract Call:
│    address:   0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    function:  getDataSet(uint256 dataSetId)
│    args:                (2811)
│
│  Docs: https://viem.sh/docs/contract/readContract
│  Details: The request timed out.
│  Version: viem@2.47.4
└  Inspection failed
```

beck-8:
## Summary

- `listDataSets()` was the only caller of `synapse.storage.findDataSets()`
  in this repo. That SDK method enriches every data set for an address in
  one unbounded parallel fan-out, which times out for accounts with
  thousands of data sets (#362).
- Switch `listDataSets()` to synapse-core's `getPdpDataSets()` directly,
  which paginates the ID listing and enriches in batches of 20 via
  multicall — the same pattern #483 already used for the single-dataset
  path (`getDetailedDataSet`).
- `listDataSets()` is shared by `data-set ls`, `payments status`, and
  add/import's default data-set-reuse check (`resolveDefaultDataSetReuse`),
  so all three are fixed by this one change.
- `ListDataSetsOptions.filter` now runs on the mapped `DataSetSummary`
  instead of the raw SDK shape (removes a stale comment/TODO about the SDK
  not supporting metadata filtering — no longer relevant since we don't
  call the SDK method at all).

## Test plan

- [x] `pnpm test` (lint + typecheck + unit, 848 passed)
- [x] Manually verified against the calibration address from #362
      (`0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523`, 2274 data sets):
  - `data-set ls --view-address <addr> --network calibration --all` —
    completes in ~2 min, previously timed out
  - `payments status --view-address <addr> --network calibration` —
    completes in ~2.5 min, previously timed out

Fixes #362


src/core/data-set/list-data-sets.ts
```diff
@@ -32,21 +39,32 @@ export async function listDataSets(synapse: Synapse, options?: ListDataSetsOptio
   const address = options?.address ?? getClientAddress(synapse)
   const filter = options?.filter
 
-  const dataSets = await synapse.storage.findDataSets({ address })
-
-  const filteredDataSets = filter ? dataSets.filter(filter) : dataSets
+  const dataSets: DataSetSummary[] = []
 
-  return filteredDataSets.map((ds) => {
+  const pages = paginate(({ cursor }) => getPdpDataSets(synapse.client, { address, cursor }))
```
SgtPooki: blocker: `listDataSets()` throws for any account that has even one terminated dataset.

`getPdpDataSets()` (called here) enriches a whole page in one multicall with `allowFailure: false`. It invokes `getDataSetListenerCall` and `getDataSetLeafCountCall` raw, not the tolerant wrappers. Both revert with 'Data set not live' on a non-live dataset. One revert fails the whole multicall, so `getPdpDataSets()` throws and this function throws with it.

It is very likely most accounts will have expired/old datasets. So `data-set ls`, `payments status`, and the add/import reuse check all break, not just edge cases.

The old `findDataSets()` path tolerated terminated datasets, which is what the `hasActivePieces: live && ...` convention at line 59 assumes. This swap dropped that tolerance.

Fix: enrich per dataset, not per page. Catch the NOT_LIVE revert per dataset and still emit the dataset with `live=false`, `hasActivePieces=false`. The guard at line 59 already does the right thing once terminated datasets actually reach the mapper.

src/core/data-set/list-data-sets.ts
```diff
@@ -32,21 +39,32 @@ export async function listDataSets(synapse: Synapse, options?: ListDataSetsOptio
...
+  for await (const pdpDataSet of pages) {
```
SgtPooki: `ListDataSetsOptions.logger` is accepted (types.ts line 109) but never used, and there is no try/catch. Sibling `getDetailedDataSet()` logs before rethrowing.

On a large account this loops 40+ pages. A failure late in the loop drops the address and cursor from the error, so the caller loses context. Wrap the loop in try/catch, log `{ address, error }` (page/cursor if cheap), then rethrow. Observability parity, not a behavior change.

src/test/unit/core-data-set.test.ts
```diff
@@ -36,7 +36,15 @@ const {
...
-  const mockFindDataSets = vi.fn(async () => state.datasets)
+  const mockGetPdpDataSets = vi.fn(async () => ({
+    items: state.datasets.map((ds) => ({
+      ...ds,
+      dataSetId: ds.pdpVerifierDataSetId,
+      live: ds.isLive,
+      managed: ds.isManaged,
+      cdn: ds.withCDN,
+    })),
+  }))
```
SgtPooki: nit: This mock returns SDK-shaped objects, not real `PdpDataSet` shapes. Each item spreads `...ds` and adds `dataSetId`, `live`, `managed`, `cdn`. A real `PdpDataSet` carries `dataSetId`, `live`, `managed`, `cdn`, `metadata`, `provider` and has no `pdpVerifierDataSetId`/`isLive`/`isManaged`/`withCDN`.

So the 'does not leak' test only checks the three renamed keys, and the terminated-dataset test passes on a mock that never throws. Build a real-shaped fixture (mirror `toPdpDataSet()` in data-set.test.ts) and assert both `dataSetId` and `pdpVerifierDataSetId` come from the real `dataSetId`.

src/test/unit/core-data-set.test.ts
```diff
@@ -96,6 +103,24 @@ vi.mock('@filoz/synapse-sdk', async () => {
...
+  ) {
+    let cursor = 0n
+    while (true) {
+      const page = await getPage({ cursor })
+      yield* page.items
+      if (page.nextCursor == null) return
+      cursor = page.nextCursor
+    }
+  },
+}))
```
SgtPooki: This `paginate` mock is pasted verbatim and omits the production guard where a non-advancing `nextCursor` throws. Same copy lives in upload-flow.test.ts (right side, line 33).

`paginate` is pure. Use `vi.importActual('@filoz/synapse-core')` and only mock `warm-storage`. That removes the duplicate and the risk that a future test pulling the real `paginate` gets an incomplete module.

src/test/unit/core-data-set.test.ts
```diff
@@ -193,6 +218,51 @@ describe('listDataSets', () => {
...
+        hasActivePieces: true,
+        serviceProvider: '0xservice',
+        payer: '0xpayer',
+        payee: '0xpayee',
+      },
+    ]
+
+    const [result] = await listDataSets(mockSynapse as any)
+
+    expect(result?.hasActivePieces).toBe(false)
```
SgtPooki, line 236: nit: This assertion describes a path unreachable on real contracts. `getPdpDataSets()` throws on a non-live dataset before the mapper runs, so `result?.hasActivePieces` is never evaluated for a terminated dataset outside this mock.

It reads as 'terminated datasets get `hasActivePieces: false`', but it only proves the mock returns them. Rename or annotate it to say it tests the mapper in isolation, not the enrichment path. The blocker on line 44 is the real thing to fix.

SgtPooki (changes requested):

src/core/data-set/list-data-sets.ts
beck-8: I checked real accounts on both networks using foc-observer and compared both listing paths. Deleted datasets (ID `1203` on mainnet, ID `9012` on calibration) were absent
  from both results. Service-terminated but still-live dataset ID `405` on calibration was returned successfully by both, including via the CLI.

  The old SDK path also uses the raw listener call with `allowFailure: false` and rethrows. I couldn't reproduce a loss of tolerance here. Do you have a specific account where
  the old path succeeds and the new one fails?

SgtPooki (approved):

src/core/data-set/list-data-sets.ts
SgtPooki: Agreed. `deleteDataSet` calls `dataSetDeleted` in the same transaction, which drops the ID from the client list, so `getPdpDataSets()` never sees a non-live ID. Resolving.

SgtPooki (approved):
Approving. The other four comments are addressed in c1dee445.

Before merge: this now conflicts with master, likely #725, which touches `src/data-set/run.ts` and `src/test/unit/data-set.test.ts`.

non-blocking: the doc comment in `list-data-sets.ts` says a data set deleted mid-listing "can be omitted". The enrich multicall uses `allowFailure: false`, so the whole listing throws instead. Worth rewording.

PR #709, opened 2026-09-04
fix: upgrade synapse and migrate storage cost calculations

### LINKED ISSUE #685: Migrate to Synapse 2.0 (post 202608 FWSS mainnet upgrade)
Synapse 2.0 ships immediately after the [[Tracking Issue] 202608 FOC Release (FilOzone/filecoin-services#597)](https://github.com/FilOzone/filecoin-services/issues/597) and removes the piece-metadata getters that filecoin-pin currently depends on.

**Scope:**
- Bump `@filoz/synapse-sdk` to `^2.0.0` and `@filoz/synapse-core` accordingly once released in https://github.com/FilOzone/synapse-sdk/pull/942=

**Depends on:** 
- #684 (update to latest Synapse release)
- [Stop exposing piece metadata (filecoin-project/filecoin-pin#668)](https://github.com/filecoin-project/filecoin-pin/issues/668) is already out for PR and can land independently/beforehand. This issue is specifically about the dependency bump to 2.0 and validating against the post-upgrade contracts. It depends on that removal work being completed.
- the FWSS mainnet contract upgrade / Synapse 2.0 release.

hugomrdias:
## Summary

- Upgrade `@filoz/synapse-core` to 0.9.0 and `@filoz/synapse-sdk` to 2.0.0
- Migrate storage and payment calculations to the updated leaf-count and reserve-aware APIs
- Support context-based multi-copy upload funding estimates
- Update unit tests for the new calculation behavior

## Testing

Not run (not requested)


src/core/payments/funding.ts
```diff
@@ -159,7 +161,10 @@ export function calculateFilecoinPayFundingPlan(options: FilecoinPayFundingPlanO
     throw new Error('A funding target is required')
   }
 
-  if (pieceSizeBytes != null && priceList == null) {
+  const usePrecomputedUploadCosts =
+    pieceSizeBytes != null && targetRunwayDays != null && uploadCosts != null && (mode === 'minimum' || !allowWithdraw)
+
+  if (pieceSizeBytes != null && priceList == null && !usePrecomputedUploadCosts) {
     throw new Error('priceList is required when pieceSizeBytes is provided')
   }
```
SgtPooki: **non-blocking (should-fix): no direct unit test for this branch**

`payments-funding.test.ts` never supplies `uploadCosts` or `contexts`, so every test there still drives the price-list fallback. The only place this branch runs under test is `upload-action.test.ts`, where the SDK stub returns `0n` for every field. With all-zero inputs the branch cannot fail: reading `lockups.total` instead of `depositNeeded`, adding the rate delta to the wrong base, or flipping the reason-code ternary would all pass.

Since `autoFund` and the action both pass `allowWithdraw: false`, this is now the path every real upload takes.

Suggested tests:

- `calculateFilecoinPayFundingPlan` with a stub `uploadCosts` carrying distinct nonzero values, `priceList` omitted, `allowWithdraw: false`. Assert `delta === depositNeeded`, `targetDeposit === balance + delta`, `projectedRateUsed` and `projectedLockupUsed` equal the account values plus the stub deltas, and `reasonCode` is `piece-upload` at 0 days and `runway-with-piece` otherwise.
- `planFilecoinPayFunding` with `contexts`. Assert `calculateMultiContextCosts` is called with `pieceSizes: [BigInt(size)]` and `extraRunwayEpochs: BigInt(days) * EPOCHS_PER_DAY`, and `getStorageInfo` is not called.

While there: the comment left at `upload-action.test.ts:615` where the `getStorageInfo` assertion was removed could become the second assertion above.

src/core/payments/funding.ts
```diff
@@ -409,8 +428,21 @@ export async function planFilecoinPayFunding(options: PlanFilecoinPayFundingOpti
 
   const [status, accountSummary] = await Promise.all([getPaymentStatus(synapse), synapse.payments.accountSummary({})])
 
+  let uploadCosts: UploadCosts | undefined
+  if (
+    pieceSizeBytes != null &&
+    targetRunwayDays != null &&
+    contexts != null &&
+    (mode === 'minimum' || !allowWithdraw)
```
SgtPooki: **non-blocking (should-fix): gate predicate duplicated, rationale undocumented**

This condition is the same as `usePrecomputedUploadCosts` on line 164. The wrapper decides whether to fetch SDK costs, the pure planner decides whether to use them, and nothing keeps the two in sync.

If the pure planner is later loosened but this one is not, exact mode stays on the old math with no error. If this one is loosened but the pure planner is not, the `getStorageInfo` fallback below is skipped (because `uploadCosts` is set) and the pure planner throws `priceList is required when pieceSizeBytes is provided`.

The reason for the `minimum || !allowWithdraw` part is also not stated anywhere: the SDK's `depositNeeded` is clamped at `0n`, so it cannot express a withdrawal, and exact mode with withdrawals needs the signed local delta.

Suggestion: one predicate used by both sites, with that reason as its doc comment:

```ts
/**
 * SDK upload costs only express deposits (depositNeeded is clamped at 0n),
 * so exact mode with withdrawals stays on the local signed-delta math.
 */
function canUsePrecomputedUploadCosts(options: {
  pieceSizeBytes?: number | undefined
  targetRunwayDays?: number | undefined
  mode?: FundingMode | undefined
  allowWithdraw?: boolean | undefined
}): boolean {
  return (
    options.pieceSizeBytes != null &&
    options.targetRunwayDays != null &&
    (options.mode === 'minimum' || options.allowWithdraw === false)
  )
}
```

Then here: `if (contexts != null && canUsePrecomputedUploadCosts(options))`, and on line 164: `uploadCosts != null && canUsePrecomputedUploadCosts(options)`.

src/core/payments/funding.ts
SgtPooki: non-blocking: this branch is now what every `add` and action run takes, and no unit test hits it with nonzero values (the action test stubs everything to `0n`).

Please add one `calculateFilecoinPayFundingPlan` test with a nonzero `uploadCosts` stub and no `priceList`, asserting `delta`, `targetDeposit`, `projectedRateUsed`, `projectedLockupUsed`, and `reasonCode` for days 0 and >0.

src/core/payments/funding.ts
SgtPooki: suggestion: same predicate as line 164. Extract one helper and note why exact+withdraw is excluded (SDK `depositNeeded` is clamped at 0n, so it cannot express a withdrawal).

SgtPooki (approved):
confirmed no leftover pre-2.0 calls

PR #724, opened 2026-09-14
fix(login): scan to a named head, never latest

### LINKED ISSUE #720: CLI login wait poll
it seems like the login’s grant-wait cannot complete on the default RPC; seems like cli is using viem’s default around 10s and do 3 polls but a tx at least needs 30s to be confirmed on filecoin so glif wont be able to return a recipt before things being timeout;
(note the key was succesfully created with scope added in console properly, and saved locally as well, everything is functional just the ux shows things are timed out ; we might need a higher budget here for the poll wait rather than the viem default; and maybe a --timeout flag so i.e agents can set a higher timeout budget?

SgtPooki:
Fixes #720.

## What was wrong

Not the poll budget. `--timeout` and the 5 minute deadline were already there, and raising either would not have helped.

The owner scan asked for `toBlock: 'latest'`. Timed against the default mainnet RPC (`https://api.node.glif.io/rpc/v1`), same query, same range:

```
toBlock=latest  span=0:    ok 23338ms      toBlock=<head> span=0:    ok 33ms
toBlock=latest  span=120:  ok 23952ms      toBlock=<head> span=120:  ok 119ms
toBlock=latest  span=2880: ok 34820ms      toBlock=<head> span=2880: ok 298ms
```

viem's default request timeout is 10s, so every scan timed out and three in a row ended the wait. The grant was onchain the whole time: key created, scopes correct, session file saved, and a failure message.

## The fix

`pollOnce` reads the head with `eth_blockNumber` and scans `[fromBlock, head]`. `fromBlock` is fixed for the whole wait, resolved once in `runLogin`, and never moves, so every poll rescans the same window. That costs one extra RPC call per poll against a scan that now returns in milliseconds.

The window is one wait long, 10 epochs at the default, and a day of blocks reads in 298ms, so rescanning beats the bookkeeping it replaces. The old cursor advanced to the block of a returned log, which is what let it pass the head, and a reorg below it was never rescanned. viem's `watchEvent` advances to the head it queried, never to a log's block. Dropping the cursor takes `LogScan.lastBlock`, `seenUpTo`, and `WatchState.scanFrom` with it.

Two ranges the RPC rejects outright:

```
fromBlock = head-10, toBlock = head+100   ERROR
fromBlock = head,    toBlock = head-5     ERROR   (inverted)
```

`if (head < fromBlock) return false` rules out the second. That is safe only because `fromBlock` never advances, and `pollTolerantly` counts the skip as a success, which clears strikes a lagging node caused.

The first is still reachable: the head read and the scan are two calls, and a load-balanced endpoint can answer the second from a node behind the first. glif reports it as `tipset height in future`, so `pollTolerantly` recognises it and spends no strike. Three in a row would otherwise have ended the wait for a grant that is on chain, which is the failure this whole module exists to avoid.

## Resuming a wait

The owner authorizes once. A login that timed out after the grant was onchain could not recover: the rerun it prints scanned from the current block, so it waited for a second event that never comes, and the session file it resumed still had no owner. That is the state #720 describes, key created, scopes granted, session saved, CLI reporting a timeout.

The session file now records the block the wait started at, written before the wait rather than on success, and a rerun scans from there. A file written before this change has no block and still starts from the head; `login` is unreleased, so `--fresh` covers the few that exist.

## Verification

End to end on calibration against the staging console, real wallet, both paths:

- Approving while the wait runs: `✓ Authorized! Granted: createDataSet, addPieces`. The wait held its full budget with no RPC errors, where before it threw at 41s.
- Approving with the CLI stopped, then rerunning: also `✓ Authorized!`, in a few seconds. The grant was at block 4069781 and the recorded block was 4069771, so the rerun found an event 10 blocks behind the head. That session was unrecoverable before this change.

The watcher alone, against the default mainnet RPC, scanning 2880 blocks per poll on a 30s budget:

- before: 1 poll, then a throw at 41s
- after: 6 polls, 0 RPC errors

Unit: the window never moves, the head is re-read every poll, no scan uses `latest`, no logs are requested while the head is below the window, a run of lagging-node answers does not end the wait, and a timed-out login still records its block. Mutation-checked: restoring `toBlock: 'latest'`, dropping the skip, caching the head, counting lagging-node answers as strikes, scanning from the head on a resume, and not recording the block each turn tests red.

Three tests that pinned the old cursor are gone. One of them handed the fake a log above the block its own query was bounded at, which a real node cannot do.


SgtPooki (commented):
self review

src/core/session/watch-authorization.ts
```diff
@@ -158,40 +158,37 @@ function grantStatus(scopes: Pick<ScopeGrants, 'granted' | 'missing'>): ScopeGra
...
 ): Promise<LogScan> {
   const logs = (await client.request({
     method: 'eth_getLogs',
     params: [
       {
         address: registryAddress,
         fromBlock: `0x${fromBlock.toString(16)}`,
-        toBlock: 'latest',
+        // An explicit head, never `latest`. Measured against
+        // api.node.glif.io on 2026-09-14: numbered ranges of 0, 120 and 2880
```
jennijuju: remove these comments?

src/core/session/watch-authorization.ts
```diff
@@ -267,7 +282,7 @@ async function pollTolerantly(options: WatchAuthorizationOptions, state: WatchSt
     return done
   } catch (error) {
     state.lastError = error
-    state.consecutiveFailures += 1
+    if (!isNodeBehind(error)) state.consecutiveFailures += 1
```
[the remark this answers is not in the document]
SgtPooki: Fixed in 51edbd5. A lagging node now counts as a completed poll, clearing the strike count and marking progress, so it behaves like the `head < fromBlock` skip rather than a spared failure. Two tests cover the cases you named: two real failures either side of a lagging one no longer reach three, and a wait made only of lagging answers returns the timeout that names the saved key instead of throwing the RPC error at the deadline.

src/login/run-login.ts
```diff
@@ -210,7 +210,12 @@ export async function runLogin(options: LoginOptions): Promise<number> {
 
   // The link is already on screen: an RPC that is down here costs the wait, not the pairing.
   const client = createPublicClient({ chain, transport })
-  const fromBlock = await getBlockNumber(client)
+  // A resumed wait scans from where the first one started, so a grant that
+  // landed while nothing was watching is still found. The owner authorizes
+  // once; a rerun that scanned from the current block would wait for a second
+  // event that never comes.
+  const fromBlock = session.fromBlock ?? (await getBlockNumber(client))
+  if (session.fromBlock === undefined) writeSessionFile({ ...session, fromBlock }, path)
```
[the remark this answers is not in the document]
SgtPooki: Right, and it is the same stranding this PR fixes for the wait path. `--no-wait` now records the block before it returns, best effort: the pairing is already on screen, so an RPC that is down there must not undo it, and a rerun without a marker just starts at its own head as before. Two tests: the marker is written on the `--no-wait` path, and a failed block read still pairs.

src/core/session/watch-authorization.ts
SgtPooki: Trimmed rather than removed, in 06f7a17. The measurements moved to the PR body and commit message, which is where a sourced number belongs; what is left is two lines naming the hazard and pointing at #720.

I kept a warning there because `latest` reads as the obvious simplification, and taking it is the whole bug: the scan goes from milliseconds to tens of seconds and every poll blows viem's 10s timeout. A test catches it too, but the next person editing that literal sees the comment first.

src/core/session/watch-authorization.ts
```diff
@@ -226,14 +219,17 @@ interface WatchState {
...
+  if (state.owner === undefined && fromBlock !== undefined) {
+    const head = await headBlock(client)
+    // A load-balanced endpoint can answer from a node that has not reached the
+    // block the wait started at; [fromBlock, head] would be an inverted range,
+    // which the RPC rejects. Nothing is there yet either way, so wait.
+    if (head < fromBlock) return false
+    // The whole window every poll, never a moving cursor: it spans one wait and
+    // rereading it is cheap. A cursor would have to advance past a head some
+    // node may not have, and would skip a reorg below itself.
+    const scan = await findOwnerInLogs(client, registryAddress, sessionAddress, fromBlock, head)
```
BravoNatalie, line 232: since `fromBlock` is persisted in session, a later rerun can scan days or weeks old events. That means it could potentially hit the provider’s range limit or timeout again.

Could we also have a second scan checkpoint during the current run? this way each poll does not reread everything

src/login/run-login.ts
```diff
@@ -200,17 +200,34 @@ export async function runLogin(options: LoginOptions): Promise<number> {
...
 
+  // The link is already on screen: an RPC that is down here costs the wait, not the pairing.
+  const client = createPublicClient({ chain, transport })
+
   if (options.wait === false) {
+    // This exit tells the owner to rerun `login`, and the grant is usually
+    // mined before they do, so it needs the same resume marker the wait
+    // records. Best effort: the pairing is done and an RPC that is down here
+    // must not undo it; a rerun without a marker starts at its own head.
+    if (session.fromBlock === undefined) {
```
BravoNatalie, line 211: Could we save this block before printing and opening the approval link? 

Right now the user could approve the key before this block is saved. If that happens, or if this block read fails, the next login may start searching after the approval event and never find it.

BravoNatalie (approved):
Just a couple of edge cases to handle, but I’m approving so I don’t block you.

src/core/session/watch-authorization.ts
SgtPooki: Measured today: a third of the mainnet pool caps `eth_getLogs` at 2880 blocks (7 of 20 scans failed with retries off). viem retries that error 3 times, so a scan fails ~1.5% of the time, and the wait needs three failed polls in a row. Three runs against a 10000-block-old marker logged no error. Calibration has no cap.

Rescanning 10000 blocks costs 400-800ms per 5s poll, and stops once the grant is found. Not worth a cursor.

src/login/run-login.ts
SgtPooki: The gap is one `eth_blockNumber` (33ms); the browser has not opened yet. If the read fails, the link is already printed and a rerun scans from its own head, same as today. Keeping the order.

PR #725, opened 2026-09-14
fix(data-set): terminate with the session key

### LINKED ISSUE #721: dataset terminate with session key
`dataset terminate` while succesfully ask users to grant scope in the console; however it doesnt actually execute with that; it looks like? in cli the terminate service is only callable with a  tx when from is the owner wallet.. in our new flow that is recommended to be a scoped session key as well

SgtPooki:
Fixes #721.

## What was wrong

`dataset terminate` gates on the TerminateService scope, sends the user to the console to grant it, lets a session key past the signing check, and then never uses it.

The call passed `skipProvider: true`, and in the SDK that branch submits from `synapse.client`:

```js
// @filoz/synapse-sdk storage/terminate.js
if (options.skipProvider === true) {
    const { receipt } = await terminateServiceSync(synapse.client, { dataSetId, onHash: onSubmitted })
```

Under session-key auth that account is the owner address as a bare string, with no signer. The other branch is the one built for session keys: it signs an EIP-712 `TerminateService(uint256 dataSetId)` payload, the same typehash the scope grants, and the provider submits it.

## The fix

Pick the path by what the CLI holds. A session key goes through the provider; the owner wallet keeps the direct call and today's output.

The provider returns only once the termination is confirmed, so that path skips the confirmation prompt and the receipt wait, and tolerates a provider that names no transaction hash.

## When the provider refuses

`login` mints the session key without gas, so the provider submits for it and a provider that refuses leaves only the owner wallet. That happens when the provider no longer knows the data set (`Data set not found`) or when the account cannot settle its lockup in full, where the SDK advises `skipProvider: true`, which a session key cannot use. Neither named the way out, so the provider path now appends it:

```
A session key can only terminate through the provider.
Retry with the owner wallet:  filecoin-pin data-set terminate 158 --private-key <key>
```

A private-key failure keeps its original error.

## Verification

Live on calibration with a session key holding `terminateService`: `Data set termination complete` for data set 138, `PDP payments ended @ epoch 4069832` on chain. The key ended with nonce 0 and no balance, so the provider submitted it, which is the path a session key cannot reach alone.

Four unit tests: the provider path takes no `skipProvider` and waits on no receipt, an answer with no hash still reports success, a provider refusal names `--private-key`, a private-key failure does not. Mutation-checked: forcing `skipProvider: true`, dropping the hint, and adding it on both paths each turn tests red.


SgtPooki (commented):
self review

src/data-set/run.ts
```diff
@@ -315,15 +358,10 @@ export async function runTerminateDataSetCommand(dataSetId: number, options: Dat
...
 
-    // Direct (client-submitted) termination preserves the existing confirmation
-    // UX: `skipProvider` keeps the old `terminateDataSet` behaviour where the
-    // owner wallet submits the on-chain transaction and always gets a tx hash.
-    const { txHash } = await synapse.storage.terminateService({ dataSetId: BigInt(dataSetId), skipProvider: true })
-    if (txHash == null) {
-      throw new Error('Termination did not return a transaction hash')
-    }
+    const { txHash, alreadyConfirmed } = await submitTermination(synapse, dataSetId, viaProvider)
+    const confirmed = alreadyConfirmed || shouldWait === true
```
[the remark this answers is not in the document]
SgtPooki: Fixed in 6191b18. The final status label and the outro now read `confirmed`, the same value the spinner line already used. One test: a session-key run with no `--wait` ends with "Data set termination complete" and "Final Data Set Status:". Reverting either label to `shouldWait` turns it red.

src/data-set/run.ts
```diff
@@ -204,6 +205,42 @@ export async function runDataSetListCommand(options: DataSetListCommandOptions):
...
+    })
+  } catch (error) {
+    if (!viaProvider) throw error
+    // `login` mints the session key without gas, so the provider submits for
+    // it and a provider that refuses leaves no way through. The SDK answers a
+    // shortfall with `skipProvider: true` and a provider rejection with
+    // nothing at all; both need the owner wallet, so name it.
+    const reason = error instanceof Error ? error.message : String(error)
+    throw new Error(
+      `${reason}\n  A session key can only terminate through the provider. Retry with the owner wallet:  filecoin-pin data-set terminate ${dataSetId} --private-key <key>`,
```
BravoNatalie, line 234: nit: This also catches timeouts and network errors. So, consider only adding the owner-wallet suggestion for errors we know are provider refusals or funding problems

src/data-set/run.ts
```diff
@@ -315,15 +358,10 @@ export async function runTerminateDataSetCommand(dataSetId: number, options: Dat
...
     spinner.start('Submitting termination transaction...')
 
-    // Direct (client-submitted) termination preserves the existing confirmation
-    // UX: `skipProvider` keeps the old `terminateDataSet` behaviour where the
-    // owner wallet submits the on-chain transaction and always gets a tx hash.
-    const { txHash } = await synapse.storage.terminateService({ dataSetId: BigInt(dataSetId), skipProvider: true })
-    if (txHash == null) {
-      throw new Error('Termination did not return a transaction hash')
-    }
+    const { txHash, alreadyConfirmed } = await submitTermination(synapse, dataSetId, viaProvider)
```
BravoNatalie: why not  keep the `endEpoch` returned by `submitTermination()`? 
The provider has already confirmed the termination and returned this value, but we currently discard it and rely on another chain read with `getDetailedDataSet()` later on.

BravoNatalie (approved):

src/data-set/run.ts
SgtPooki: Fixed in 183d294. Hint only on `TerminateServiceError`, `TerminateServiceNotSupportedError`, or the SDK's lockup shortfall; a network error keeps its own message.

src/data-set/run.ts
SgtPooki: Good call, fixed in 1f3fcde. Final status now comes from `endEpoch`, no second read.

PR #726, opened 2026-09-14
feat(logout): deep link to revoke the key

### LINKED ISSUE #722: Enhancement: revoke deep link
when logging out, should we give a deep link with the key to revoke?
Revoke it on the console’s Session keys page:
│  https://pay.filecoin.cloud/console/session-keys

SgtPooki:
Fixes #722. The console half is FilOzone/filecoin-pay-explorer#451, which reads the param this PR emits.

## What this does

`logout` printed the bare session-keys page, so the owner landed on the list and had to find the key it had just dropped. It now prints a link that opens the console's revoke dialog on that key:

```
https://pay.filecoin.cloud/console/session-keys?revoke=0x<session address>&network=calibration
```

`buildRevokeUrl` returns nothing for a network the console has no page for (devnet, a custom RPC) or a session file that never recorded one. The console refuses a revoke link it cannot place, so those cases keep printing the plain page rather than a link that would fail.

## Verification

Unit. `buildRevokeUrl` (both networks, trailing slash, and the three refusals), and `logout` printing the deep link for an authorized key and falling back for devnet and for a file with no network. The explorer PR has a test that parses these exact strings, so the two halves cannot drift apart silently.

## Order

The console PR should land first. Until it does, the param is inert: the page ignores an unknown query param, so the link still opens the session-keys page, which is today's behavior.


SgtPooki (commented):
self review

src/login/run-logout.ts
BravoNatalie: What if the user uses `--no-wait` and `walletAddress` isn’t saved? Logging out would then delete an authorized key without showing them how to revoke it.

BravoNatalie (approved):

src/login/run-logout.ts
SgtPooki: Right. Fixed in 72d1117: the link prints for any session, and without a saved wallet the line reads "If you approved this key, it stays authorized...".

PR #728, opened 2026-09-16
fix(payments): accept a zero piece in setup check

### LINKED ISSUE #719: filecoin-pin add fails for every input: the minimum setup check passes a zero piece size
## Description

`filecoin-pin add` fails for every input on a funded Calibration account:

```
$ printf 'hi\n' > /tmp/repro/a.txt
$ filecoin-pin add /tmp/repro/a.txt --network calibration --private-key $KEY

Filecoin Pin Add
✓ Directory validated
✓ Connected to Filecoin - Calibration testnet
✓ Reusing existing data sets 54, 85 (991 matched, picked the oldest ones with pieces)
✗ Add failed: pieceSizes must contain only positive byte sizes
Add failed
```

A two-byte file and a 1.3 MB directory fail the same way, with the same message, before anything is packed. The content is not involved.

`--dry-run` on the same file succeeds and prints a cost estimate. That is the discriminating case, and it points at the one call the dry-run path skips.

filecoin-pin 2.0.1, @filoz/synapse-sdk 2.0.0, @filoz/synapse-core 0.9.0, node 26.2.0.

## Cause

`add` runs a minimum setup check with a file size of zero, and a zero is no longer an acceptable piece size.

`dist/add/add.js:165`, guarded by `if (!options.autoFund && !options.dryRun)`, which is why a dry run gets through:

```js
await validatePaymentSetup(synapse, 0, spinner, { suppressSuggestions: true })
```

The zero is deliberate. `dist/common/upload-flow.js:275` documents it:

```js
/** @param fileSize - Size of file to upload in bytes (use 0 for minimum setup check) */
```

and the same function branches on it at line 359, so the zero is used on purpose and not left over. It then travels: `validatePaymentSetup` to `checkUploadReadiness({ fileSize: 0 })` to `validatePaymentCapacity(synapse, 0)` to a cost calculation that wraps it as `pieceSizes: [0n]`, and lands on `@filoz/synapse-core/dist/src/utils/pdp-size.js:15`:

```js
if (size <= 0n) throw new ValidationError('pieceSizes must contain only positive byte sizes')
```

So a documented sentinel meets a validator that rejects it. The upload path itself is fine; `executeUpload` stores records against the same account without complaint.

## Suggested fix

The codebase already answers this in one place. `dist/core/payments/funding.js:271` asks the same question about an account rather than a file and passes the smallest legal value instead of zero:

```js
const reserveFunding = calculateLifecycleReserveFunding({ priceList: params.priceList, isNewDataSet: true, pieceSizes: [1n] })
```

The minimum setup check wants exactly that: can this account upload at all, independent of what. Either have `validatePaymentCapacity` substitute `1n` when asked about zero bytes, or have `add` call the setup check with a size rather than a sentinel.

Worth a test that calls `checkUploadReadiness` with `fileSize: 0`, since that is the contract the docstring promises and nothing currently holds it.

## Note on the range

`filecoin-pin@2.0.1` declares `@filoz/synapse-core: ^0.9.0`, and 0.9.0 is the newest, so this is not a floating dependency drifting under a pin. Earlier runs of `filecoin-pin add` from this machine succeeded today, and the global install's files were rewritten partway through the day. I did not capture what version preceded it, so I cannot say what changed, only that the failure is deterministic now.

SgtPooki:
Fixes #719.

## What was wrong

`add` runs the minimum setup check with a file size of 0 (`add.ts:232`), and that 0 reaches the SDK as a piece:

```js
// calculatePieceUploadRequirements
const lockup = calculateAdditionalLockupRequired({ pieceSizes: [BigInt(pieceSizeBytes)], ... })
```

synapse-core 0.9 rejects `[0n]`, so every `add` without `--auto-fund` failed before packing. `--dry-run` skips the check, which is why it kept working.

## The fix

`calculatePieceUploadRequirements` returns zero allowances for exactly 0 and sends every other size, negative and NaN included, through the SDK as before. Every caller of the setup check goes through it, so `add` needs no change.

## Verification

`validatePaymentSetup(synapse, 0)`, the call `add` makes, now runs in a unit test with the readiness and capacity checks and synapse-core's lockup math all real; only the two chain reads are faked. Red on master with the error from the issue. Unit tests on the function itself: size 0 gives zero allowances, -1 and NaN still throw, 1 MiB against an empty balance still needs a deposit.

Live on calibration, provider 2: `add` on a 2.7 KiB file now prints `Minimum payment setup verified` and stores the piece.


SgtPooki (commented):
self review

src/core/payments/index.ts
```diff
@@ -861,13 +863,18 @@ export function calculatePieceUploadRequirements(
...
-    pieceSizes: [BigInt(pieceSizeBytes)],
-    dataSetLeafCount: 0n,
-    priceList,
-    isNewDataSet: false,
-    withCDN: false,
-  })
+  // The SDK rejects a zero piece size, and rightly so: a zero piece is
+  // nothing and locks up nothing.
+  const lockup =
+    pieceSizeBytes > 0
```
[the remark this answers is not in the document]
SgtPooki: Agreed, changed to `=== 0`. -1 and NaN now throw as before, with a test for each.

src/test/unit/payments.compute.test.ts
```diff
@@ -250,6 +252,46 @@ describe('computeAdjustmentForExactDaysWithPiece', () => {
...
+      rateAllowance: 0n,
+      lockupAllowance: 0n,
+      rateUsage: 0n,
+      lockupUsage: 0n,
+      maxLockupPeriod: 0n,
+      isApproved: false,
+    },
+  }
+
+  it('a zero-byte piece (the minimum setup check) locks up nothing and passes (issue #719)', () => {
```
BravoNatalie, line 273: suggestion:
add a test that check the full path that failed in add
```
add → validatePaymentSetup → checkUploadReadiness → validatePaymentCapacity
```

BravoNatalie (approved):

src/test/unit/payments.compute.test.ts
SgtPooki: Done. `payment-setup-check.test.ts` starts at `validatePaymentSetup(synapse, 0)` and only fakes `isFwssMaxApproved` and `getPriceList`, so the readiness check, `validatePaymentCapacity`, and synapse-core lockup math all run. Red on master with the error from the issue. Good call, the existing readiness test mocks exactly the layer that broke.

